PR #44, opened 2015-03-08
PHP 5.6 support, FIX broken build

Jan-E:
See issue https://github.com/Azure/msphpsql/issues/45


SumoLion:
@Jan-E - Thanks for calling attention to the compilation issue you experienced with the latest PHP RC and the extension.  We will review your pull request when we release an update to the extension, at which time we will ensure it compiles correctly with the then latest shipping PHP 5.5/5.6 versions.

Jan-E:
@SumoLion See the discussion I had with Anatol, a colleague of yours at Microsoft:
https://github.com/php/php-src/commit/9bf95d82911dc7028c624494e9ac914f615f66ed#commitcomment-10136379

PR #145, opened 2016-09-05
Fix empty string bind param causes floating point exception

### LINKED ISSUE #141: PHP FPM exit with signal 8 (SIGFPE - core dumped)
OS: Ubuntu 16.04 with Nginx
Driver: 4.0.3
Tested on PHP-FPM 7.0.8 and 7.0.10 give same result

```
unixODBC 2.3.1
DRIVERS............: /etc/odbcinst.ini
SYSTEM DATA SOURCES: /etc/odbc.ini
FILE DATA SOURCES..: /etc/ODBCDataSources
USER DATA SOURCES..: /home/frost/.odbc.ini
SQLULEN Size.......: 8
SQLLEN Size........: 8
SQLSETPOSIROW Size.: 8
```

/var/log/php7.0-fpm.log
```
[26-Aug-2016 00:55:17] WARNING: [pool www] child 72407 exited on signal 8 (SIGFPE - core dumped) after 1064.849232 seconds from start
[26-Aug-2016 00:55:17] NOTICE: [pool www] child 72657 started
```

/var/log/syslog
`Aug 26 00:55:13 api kernel: [ 5691.115511] traps: php-fpm7.0[72407] trap divide error ip:7f2fb56b6cec sp:7ffeaa469b10 error:0 in php_pdo_sqlsrv_7_nts.so[7f2fb565a000+8d000]`

This error will return 502 Bad Gateway on nginx because PHP process crash. It happen with Update query.

### LINKED ISSUE #144: PHP-7.0-Linux sqlsrv_query produces floating point exception when bind parameter is a empty string
While this query worked fine with version `4.0.2` of `sqlsrv`
```php
sqlsrv_query($conn, 'SELECT ?', ['']);
```
version `4.0.3` will produce a floating point exception for the same query.
```
Core was generated by `/usr/bin/php7.0 -n -c /vagrant/msphpsql/source/sqlsrv/tmp-php.ini -d output_han'.
Program terminated with signal SIGFPE, Arithmetic exception.
#0  0x00007f4d923dd0c6 in (anonymous namespace)::default_sql_size_and_scale (param_z=0x7f4d9265cf68, param_z=0x7f4d9265cf68, 
    decimal_digits=<synthetic pointer>, column_size=<synthetic pointer>, encoding=SQLSRV_ENCODING_CHAR, paramno=3, stmt=0x7f4d9266e000)
    at /vagrant/msphpsql/source/sqlsrv/core_stmt.cpp:1963
1963					column_size = SQL_SERVER_MAX_FIELD_SIZE / char_size;
````

`char_size` is defined in [line 1957](https://github.com/Microsoft/msphpsql/blob/PHP-7.0-Linux/source/sqlsrv/core_stmt.cpp#L1957), an will be `0` if ` Z_STRLEN_P( param_z )` is zero, which happens for an empty string.

My Environment:
`Ubuntu 16.04 LTSUbuntu 16.04 LTS`
`PHP 7.0.10-1+deb.sury.org~xenial+1 (cli) ( NTS )PHP 7.0.10-1+deb.sury.org~xenial+1 (cli) ( NTS )`

This issue is similar to #141 where `pdo_sqlsrv` is affected.

PR is in Progress.

### PR DESCRIPTION
**Pull request by xalopp**
Fixes #141 and Fixes #144

### DIFF
--- a/source/pdo_sqlsrv/core_stmt.cpp
+++ b/source/pdo_sqlsrv/core_stmt.cpp
@@ -1960,7 +1960,7 @@ void default_sql_size_and_scale( sqlsrv_stmt* stmt, unsigned int paramno, zval* 
                 column_size = SQL_SERVER_MAX_TYPE_SIZE;
             }
             else {
-				column_size = SQL_SERVER_MAX_FIELD_SIZE / char_size;
+				column_size = (char_size == 0) ? 1 : (SQL_SERVER_MAX_FIELD_SIZE / char_size);
             }
             break;
         }
--- a/source/pdo_sqlsrv/tests/bug_141.phpt
+++ b/source/pdo_sqlsrv/tests/bug_141.phpt
@@ -1,0 +1,30 @@
+--TEST--
+Bug #141 (Assinging empty string as bind parameter produces floating point exception)
+--SKIPIF--
+<?php if(!extension_loaded("pdo_sqlsrv")) print "skip"; ?>
+--INI--
+--FILE--
+<?php
+require('config.inc');
+
+$dbo = new PDO("sqlsrv:server=$serverName ; Database = $database", $username, $password);
+print 'PDO sqlsrv connection successfull: '.($dbo ? 'yes' : 'no').PHP_EOL;
+
+$sth = $dbo->prepare('select ?');
+print 'PDO sqlsrv prepare successfull: '.($sth !== false ? 'yes' : 'no').PHP_EOL;
+
+$result = $sth->bindValue(1, '');
+print 'PDO sqlsrv value binding successfull: '.($result !== false ? 'yes' : 'no').PHP_EOL;
+
+$result = $sth->execute();
+print 'PDO sqlsrv execute successfull: '.($result !== false ? 'yes' : 'no').PHP_EOL;
+
+$row = $sth->fetch(PDO::FETCH_NUM);
+print 'PDO sqlsrv fetch result 1st column is empty string: ' . ($row[0] === '' ? 'yes' : 'no') . PHP_EOL;
+?>
+--EXPECT--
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: source/pdo_sqlsrv/core_stmt.cpp [hunk 1]
```diff
@@ -1960,7 +1960,7 @@ void default_sql_size_and_scale( sqlsrv_stmt* stmt, unsigned int paramno, zval*
                 column_size = SQL_SERVER_MAX_TYPE_SIZE;
             }
             else {
-				column_size = SQL_SERVER_MAX_FIELD_SIZE / char_size;
+				column_size = (char_size == 0) ? 1 : (SQL_SERVER_MAX_FIELD_SIZE / char_size);
```
xalopp: `char_size` is computed differently on Windows:
* [PHP-7.0/sqlsrv/core_stmt.cpp line 1924](https://github.com/Microsoft/msphpsql/blob/PHP-7.0/sqlsrv/core_stmt.cpp#L1924)
* [PHP-7.0-Linux/source/sqlsrv/core_stmt.cpp line 1957)](https://github.com/Microsoft/msphpsql/blob/PHP-7.0-Linux/source/sqlsrv/core_stmt.cpp#L1957)

Maybe not multiplying with the parameter string size would be a better fix?
Explaining the reason why the code changed in 4.0.3 would also help.

meet-bhagdev:
@xalopp thanks a ton for the PR. Your fix is definitely in the right direction. The problem is with character size when it is 0, so on line 1957 when we are calculating the char size, we don’t need the length of the parameter, and removing Z_STRLEN_P( param_z ) from char_size would solve this issue. We believe that will fix this issue along with some other idiosyncrasies that we observed. Stay tuned for our next release :)

PR #159, opened 2016-09-20
Fix some build warnings

remicollet:
As I think tons of warnings (I'm using -Wall option) create lot of noise and can hide real problems, this is a first set of fix for some build warnings.

P.S. I will try to work on some others later, when I will have some spare time (and after getting some feedback)

### DIFF
--- a/source/sqlsrv/conn.cpp
+++ b/source/sqlsrv/conn.cpp
@@ -33,7 +33,7 @@
 namespace {
 
 // current subsytem.  defined for the CHECK_SQL_{ERROR|WARNING} macros
-unsigned int current_log_subsystem = LOG_CONN;
+//(unused for now) unsigned int current_log_subsystem = LOG_CONN;
 
 struct date_as_string_func {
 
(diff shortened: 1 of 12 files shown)


Hadis-Fard:
@remicollet thanks a ton for your contribution, I'll start reviewing the changes right away.

Hadis-Fard:
@remicollet finished the review, super appreciate the changes you have sent.

source/sqlsrv/conn.cpp
```diff
@@ -1120,7 +1120,7 @@ void sqlsrv_conn_close_stmts( ss_sqlsrv_conn* conn TSRMLS_DC )
 
 		try {
 			// this would call the destructor on the statement.
-			int zr = zend_list_close(Z_RES_P(rsrc_ptr));
```
remicollet: Indeed, sqlsrv_stmt_dtor or ~ss_sqlsrv_stmt doesn't raise any exception, so try/catch seems unneeded.

remicollet: done

Hadis-Fard:
@remicollet Accepted the PR, thanks 👍

PR #183, opened 2016-11-17
Backport FETCHES_NUMERIC_TYPE and add FETCHES_DECIMAL_FLOAT

freman:
Moving from PHP5.x's mssql_* functions to PDO::SQLSRV we've had a couple of compatibility issues.

I solved most of our issues by backporting SQLSRV_ATTR_FETCHES_NUMERIC_TYPE from 4.1.0 (Windows) and adding the ability to return Decimals (Money type) as Float (I know people typically prefer not to do this, for obvious reasons but this lets people be backwards compatible with data returned from mssql_* functions)

### DIFF
--- a/source/pdo_sqlsrv/core_sqlsrv.h
+++ b/source/pdo_sqlsrv/core_sqlsrv.h
@@ -1292,7 +1292,7 @@ struct sqlsrv_stmt : public sqlsrv_context {
     virtual ~sqlsrv_stmt( void );
 
     // driver specific conversion rules from a SQL Server/ODBC type to one of the SQLSRV_PHPTYPE_* constants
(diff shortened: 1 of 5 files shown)


Hadis-Fard:
@freman Accepted the PR, thanks for your contribution 👍

Hadis-Fard:
@freman After merging the changes in header file shared between SQLSRV and PDO_SQLSRV, we've faced a couple of issues FETCHES_DECIMAL_FLOAT, so we removed it from the source code, adding this feature requires deeper investigation in both drivers. Thanks again for contributing in porting FETCH_NUMERIC_TYPE, and we definitely consider the investigation on FETCHES_DECIMAL_FLOAT in our future planning.

PR #201, opened 2016-12-13
Issue#192

ulvii:
Fix for the issue with buffered result sets. 
When connection option is set to utf8, we cannot assume that local code page matches the server collation. Need to call SQLGetData with target c type set to SQL_C_WCHAR.

### DIFF
--- a/pdo_sqlsrv/core_results.cpp
+++ b/pdo_sqlsrv/core_results.cpp
@@ -346,7 +346,7 @@ sqlsrv_buffered_result_set::sqlsrv_buffered_result_set( sqlsrv_stmt* stmt TSRMLS
                 else {
                     // If encoding is set to UTF-8, the following types are not necessarily column size.
                     // We need to call SQLGetData with c_type SQL_C_WCHAR and set the size accordingly. 
-                    if ( encoding == SQLSRV_ENCODING( CP_UTF8 )) {
+                    if ( encoding == SQLSRV_ENCODING_UTF8 ) {
                         meta[i].length *= sizeof( WCHAR );
                         meta[i].length += sizeof( SQLULEN ) + sizeof( WCHAR ); // length plus null terminator space
                         offset += meta[i].length;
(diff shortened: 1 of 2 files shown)


pdo_sqlsrv/core_results.cpp
```diff
@@ -314,6 +314,9 @@ sqlsrv_buffered_result_set::sqlsrv_buffered_result_set( sqlsrv_stmt* stmt TSRMLS
         conv_matrix[ SQL_C_DOUBLE ][ SQL_C_WCHAR ] = &sqlsrv_buffered_result_set::double_to_wide_string;
     }
 
+    SQLSRV_ENCODING encoding = (( stmt->encoding() == SQLSRV_ENCODING_DEFAULT ) ? stmt->conn->encoding() :
```
ulvii: Getting the encoding.

pdo_sqlsrv/core_results.cpp
```diff
@@ -335,13 +338,30 @@ sqlsrv_buffered_result_set::sqlsrv_buffered_result_set( sqlsrv_stmt* stmt TSRMLS
                 meta[i].length += sizeof( char ) + sizeof( SQLULEN ); // null terminator space
                 offset += meta[i].length;
                 break;
+            case SQL_CHAR:
```
ulvii: If encoding is set to utf8, we would be calling SQLGetData with target c type set to SQL_C_WCHAR.

pdo_sqlsrv/core_results.cpp
```diff
@@ -335,13 +338,30 @@ sqlsrv_buffered_result_set::sqlsrv_buffered_result_set( sqlsrv_stmt* stmt TSRMLS
...
+            case SQL_VARCHAR:
+                if ( meta[i].length == sqlsrv_buffered_result_set::meta_data::SIZE_UNKNOWN ) {
+                    offset += sizeof( void* );
+                }
+                else {
+                    // If encoding is set to UTF-8, the following types are not necessarily column size.
+                    // We need to call SQLGetData with c_type SQL_C_WCHAR and set the size accordingly. 
+                    if ( encoding == SQLSRV_ENCODING( CP_UTF8 )) {
```
Hadis-Fard: it is better to use SQLSRV_ENCODING_UTF8 syntax instead of SQLSRV_ENCODING( CP_UTF8 )

PR #202, opened 2016-12-13
Adding fix for the issue with buffered result sets, when the connection option is set to UTF-8

ulvii:
PR #201  for Linux


source/shared/core_results.cpp
```diff
@@ -450,13 +453,30 @@ sqlsrv_buffered_result_set::sqlsrv_buffered_result_set( sqlsrv_stmt* stmt TSRMLS
...
                 break;
+            case SQL_CHAR:
+            case SQL_VARCHAR:
+                if ( meta[i].length == sqlsrv_buffered_result_set::meta_data::SIZE_UNKNOWN ) {
+                    offset += sizeof( void* );
+                }
+                else {
+                    // If encoding is set to UTF-8, the following types are not necessarily column size.
+                    // We need to call SQLGetData with c_type SQL_C_WCHAR and set the size accordingly. 
+                    if ( encoding == SQLSRV_ENCODING( CP_UTF8 )) {
```
Hadis-Fard: Similar to Windows cpde

PR #221, opened 2017-01-11
Update pdo_util.cpp

v-kigos:
ODBC version 13 added;  error messaged formatted similar to SQLSRV extension.
"...This extension requires the Microsoft ODBC Driver 11 or 13..."


Hadis-Fard:
Ignoring TravisCI error since it is a change in windows source.

PR #227, opened 2017-01-12
Updated code structure and config.w32

Hadis-Fard:
Code structure has been updated along with config.w32 to mirror Linux structure.


Hadis-Fard:
This is windows only code, therefore TravisCI is failing

PR #242, opened 2017-01-19
Test a number (MAX_INT) to string with custom formats

v-kigos:


v-dareck:
// $conn->setAttribute(PDO::ATTR_STRINGIFY_FETCHES, false);
is in both tests. Should it be uncommented or removed?

v-kigos:
Remove

PR #245, opened 2017-01-23
Test PDO: fetch string with new line and tab characters

v-kigos:


v-kigos:
Fails to install a docker.
build

PR #248, opened 2017-01-24
Don't forget to mention build-dependancies

pmjdebruijn:
On a clean server, given the old instructions the pecl command will fail due to at least missing php7.0-dev (for phpize) and g++ (for cpp).

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ Following instructions shows how to install PHP 7.x, Microsoft ODBC driver, apac
 
 ### Step 4: Install the Microsoft PHP Drivers for SQL Server
 
+    sudo apt-get install php7.0-dev gcc g++ re2c
     sudo pecl install sqlsrv
     sudo pecl install pdo_sqlsrv
     


david-puglielli:
Thank you @pmjdebruijn for the pull request. Please note that php7.0-dev is already listed among the required packages in the apt-get command in step 1. We will add g++ to the list, though it should install gcc as a dependency.

Could you tell us why re2c is necessary?

pmjdebruijn:
The ./configure script (after phpize) warns about it being missing if it's not installed (IIRC), also the Dockerfile installs it as well.

pmjdebruijn:
both sqlsrv and pdo_sqlsrv check for re2c presence:

checking for re2c... re2c
checking for re2c version... 0.16 (ok)

david-puglielli:
Thank you. We haven't had any problems in our testing even though re2c is not available. Do you experience any specific problems because re2c is missing?

v-dareck:
Hi @pmjdebruijn. I'm going to close this. The readme is updated with g++ and re2c for Ubuntu and Red Hat.

PR #251, opened 2017-01-24
Added binary codepage

david-puglielli:


source/shared/localizationimpl.cpp
```diff
@@ -47,10 +47,14 @@ struct cp_iconv
...
 // Array of CodePage-to-IConvEncoding mappings
 // First few elements are most commonly used
+// CodePage 2 corresponds to binary. If the attribute PDO::SQLSRV_ENCODING_BINARY
+// is set, GetIndex() above hits the assert(false) directive unless we include
+// CodePage 2 below and assign an empty string to it.
 const cp_iconv cp_iconv::g_cp_iconv[] = {
     { 65001, "UTF-8" },
     {  1200, "UTF-16LE" },
     {     3, "UTF-8" },
+    {     2, "" },
```
david-puglielli: The empty string gets passed to the IConvCache constructor, which calls iconv_open(). When "" is passed to it, it is replaced with the locale dependent encoding (https://www.gnu.org/savannah-checkouts/gnu/libiconv/documentation/libiconv-1.13/iconv_open.3.html). So I think it just takes the input string and returns it to what the original input was, since the original input was in the local encoding.

yitam: I see... is there any way that when 'codepage' is 2 we simply skip this 'conversion'?

david-puglielli: I don't think there's a simple way to do it - all conversions pass through one of our convert_string_from_utf16 type functions, which call FromUtf16/ToUtf16/etc from xplat, which all rely on the xplat conversions that use iconv. We could bypass the binary conversion inside those functions, or we could skip over our convert_string functions earlier in the code progression.

PR #256, opened 2017-01-25
Github 37 sqlsrv_has_rows 

v-dareck:
Change sqlsrv_has_rows to not scroll if already true when using forward only cursor.

### DIFF
--- a/test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
+++ b/test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
@@ -1,0 +1,102 @@
+--TEST--
+sqlsrv_has_rows() using a forward and scrollable cursor
+--SKIPIF--
+--FILE--
+<?php
+
+require_once("autonomous_setup.php");
+
+// Connect
+$conn = sqlsrv_connect( $serverName, $connectionInfo);
+if( !$conn ) { die( print_r( sqlsrv_errors(), true)); }
+
+// Create database
+sqlsrv_query($conn,"CREATE DATABASE ". $dbName) ?: die();
+
+// Create table
+$query = "CREATE TABLE $tableName (ID VARCHAR(10))";
+$stmt = sqlsrv_query($conn, $query);
+
+$query = "INSERT INTO $tableName VALUES ('1998.1'),('-2004'),('2016'),('4.2EUR')";
+$stmt = sqlsrv_query($conn, $query) ?: die( print_r( sqlsrv_errors(), true) );
+
+// Fetch data using forward only cursor
+$query = "SELECT ID FROM $tableName";
+$stmt = sqlsrv_query($conn, $query)
+		?: die( print_r(sqlsrv_errors(), true));
+        
+// if we skip the next four calls it's fine
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+
+if (sqlsrv_has_rows($stmt)) {
+	while( $row = sqlsrv_fetch_array( $stmt, SQLSRV_FETCH_NUMERIC))  
+	{  
+		echo $row[0]."\n";
+	}
+}
+
+// Fetch data using a scrollable cursor
+$stmt = sqlsrv_query($conn, $query, [], array("Scrollable"=>"buffered"))
+		?: die( print_r(sqlsrv_errors(), true));
+
+// if we skip the next four calls it's fine
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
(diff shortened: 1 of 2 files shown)


test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
```diff
@@ -0,0 +1,102 @@
...
+
+$query = "INSERT INTO $tableName VALUES ('1998.1'),('-2004'),('2016'),('4.2EUR')";
+$stmt = sqlsrv_query($conn, $query) ?: die( print_r( sqlsrv_errors(), true) );
+
+// Fetch data using forward only cursor
+$query = "SELECT ID FROM $tableName";
+$stmt = sqlsrv_query($conn, $query)
+		?: die( print_r(sqlsrv_errors(), true));
+        
+// if we skip the next four calls it's fine
```
yitam: Please remove the two comments "if we skip the next four..."

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
```diff
@@ -0,0 +1,102 @@
...
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+
+if (sqlsrv_has_rows($stmt)) {
+	while( $row = sqlsrv_fetch_array( $stmt, SQLSRV_FETCH_NUMERIC))  
+	{  
+		echo $row[0]."\n";
+	}
+}
+
```
yitam: What will happen if we call sqlsrv_has_rows() in the while block?

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
```diff
@@ -0,0 +1,102 @@
...
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+
+// Fetch data using a scrollable cursor
+$stmt = sqlsrv_query($conn, $query, [], array("Scrollable"=>"buffered"))
+		?: die( print_r(sqlsrv_errors(), true));
+
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+echo "Has Rows?" . (sqlsrv_has_rows($stmt) ? " Yes!" : " NO!") . "\n";
+
```
yitam: Any reason why you call sqlsrv_has_rows() twice? What if you call sqlsrv_fetch() first?

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
v-dareck: has_rows will return true as the result has rows.

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
v-dareck: has_rows still returns false if sqlsrv_fetch is called first.
The test is so that has_rows still returns false on repeated calls.
Previously has_rows could return true one call and false the next.

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
yitam: I meant calling sqlsrv_has_rows() inside the while block (say after the echo). Would it affect fetching anyhow? It should not, right?

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
yitam: That makes sense. Please put the above in a comment?

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
v-dareck: No it does not affect fetching now.

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
```diff
@@ -0,0 +1,103 @@
...
+// Connect
+$conn = sqlsrv_connect( $serverName, $connectionInfo);
+if( !$conn ) { die( print_r( sqlsrv_errors(), true)); }
+
+// Create database
+sqlsrv_query($conn,"CREATE DATABASE ". $dbName) ?: die();
+
+// Create table
+$query = "CREATE TABLE $tableName (ID VARCHAR(10))";
```
Hadis-Fard: after running the script twice, it doesn't reproduce the issue anymore because the table exist and the number of values gets doubled. better to use $dbName to create the table, or drop table at the end of the test

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
```diff
@@ -0,0 +1,103 @@
+--TEST--
+sqlsrv_has_rows() using a forward and scrollable cursor
+--SKIPIF--
```
Hadis-Fard: please add --Description-- to elaborate the bug.

test/sqlsrv/srv_037_sqlsrv_has_rows.phpt
v-dareck: $tablename is defined in autonomous_setup.php as $dbName.dbo.tablename.

yitam: Would you change the script to use temporary tables?

Hadis-Fard:
Please merge this code into 'merged-codebase' branch

lifzgreat:
Thank you! This was driving me nuts. I had a single row returned and if I ran `sqlsrv_has_rows($stmt)` on it twice, it returned `true` then `false` right afterward. This fixed it.

PR #257, opened 2017-01-27
Merged Linux and windows the 'shared' code

Hadis-Fard:


source/shared/xplat.h
```diff
@@ -0,0 +1,503 @@
...
+ULONGLONG
+VerSetConditionMask(
+        IN  ULONGLONG   ConditionMask,
+        IN  DWORD   TypeMask,
+        IN  BYTE    Condition
+        );
+
+
+
+//#include <basetsd.h>
```
david-puglielli: We don't want to keep these comments do we? And all the ones below

source/shared/xplat.h
```diff
@@ -0,0 +1,503 @@
...
+//inline HRESULT CreateGlobalSOSHostInterface() { return 0; }
+
+//// These are temporary solution versions of the real files that contain the minimal declarations
+//// needed to compile for non-Windows platforms.  See the special include path for the
+//// location of these files.
+//#include <guiddef.h>
+//#include <objbase.h>
+//#include <winbase.h>
+//#include <winnls.h>
+#include <winerror.h>
```
david-puglielli: We've renamed it now to xplat_winerror.h

source/shared/xplat.h
Hadis-Fard: done :)

source/shared/core_results.cpp
```diff
@@ -81,78 +89,232 @@ SQLPOINTER read_lob_field( sqlsrv_stmt* stmt, SQLUSMALLINT field_index, sqlsrv_b
...
+
+        for ( const auto &mb : str_num )
+        {
+            char16_t ch16;
+            std::mbstate_t mbs = std::mbstate_t();
+
+            int len = mbrtoc16( &ch16, &mb, str_num_end - str_num_ptr, &mbs );
+            if ( len > 0 || len == -3 )
+            {
+                //str.append( std::u16string( &ch16, len ) );
```
david-puglielli: Comment can be removed.

source/shared/core_conn.cpp
```diff
@@ -124,85 +127,83 @@ sqlsrv_conn* core_sqlsrv_connect( sqlsrv_context& henv_cp, sqlsrv_context& henv_
...
+            wconn_string = utf16_string_from_mbcs_string( SQLSRV_ENCODING_UTF8, conn_str.c_str(), static_cast<unsigned int>(conn_str.length()), &wconn_len );
+
+            CHECK_CUSTOM_ERROR( wconn_string == 0, conn, SQLSRV_ERROR_CONNECT_STRING_ENCODING_TRANSLATE, get_last_error_message())
+            {
+                throw core::CoreException();
+            }
+
+            SQLSMALLINT output_conn_size;
+            r = SQLDriverConnectW( conn->handle(), NULL, reinterpret_cast<SQLWCHAR*>(wconn_string.get()),
+            static_cast<SQLSMALLINT>(wconn_len), NULL, 0, &output_conn_size, SQL_DRIVER_NOPROMPT);
```
david-puglielli: Indentation is off

source/shared/core_results.cpp
```diff
@@ -81,78 +89,231 @@ SQLPOINTER read_lob_field( sqlsrv_stmt* stmt, SQLUSMALLINT field_index, sqlsrv_b
...
+
+    return 0;
+}
+
+#ifndef _WIN32
+// copy the number into a char string using the num_put facet
+template <typename Number>
+SQLRETURN get_string_from_stream(Number number_data, std::basic_string<char> &str_num, size_t precision, sqlsrv_error_auto_ptr& last_error)
+{
+    //std::locale loc( std::locale(""), new std::num_put<char> );	// By default, SQL Server doesn't take user's locale into consideration
```
david-puglielli: Do we need this comment?

source/shared/core_sqlsrv.h
```diff
@@ -1310,7 +1343,11 @@ const int SQLSRV_DEFAULT_SIZE = -1;     // size given for an output parameter th
 const unsigned int QUERY_TIMEOUT_INVALID = 0xffffffff;
 
 // special buffered query constant
+#ifndef _WIN32
```
david-puglielli: Is it necessary to have this defined two ways? The original caused bugs in Linux, but can this be 42 in Windows?

source/shared/typedefs_for_linux.h
```diff
@@ -0,0 +1,109 @@
...
+#define CP_THREAD_ACP             3           // current thread's ANSI code page
+#define CP_SYMBOL                 42          // SYMBOL translations
+
+#define CP_UTF7                   65000       // UTF-7 translation
+
+#define MAKELANGID(p, s)          ((((WORD  )(s)) << 10) | (WORD  )(p))
+#define LANG_NEUTRAL              0x00
+#define SUBLANG_DEFAULT           0x01        // user default
+
+// #ifndef _GETLASTERROR
```
david-puglielli: Comments can go, we've defined the LastError functions elsewhere

source/shared/core_sqlsrv.h
yitam: I agree with @david-puglielli 
Perhaps we should find a constant that works for both?

source/shared/core_conn.cpp
```diff
@@ -533,13 +533,11 @@ connection_option const* get_connection_option( sqlsrv_conn* conn, SQLULEN key,
 // option in the connection string.
 
 void build_connection_string_and_set_conn_attr( sqlsrv_conn* conn, const char* server, const char* uid, const char* pwd, 
-                                                     HashTable* options, const connection_option valid_conn_opts[], 
-                                                     void* driver,_Inout_ std::string& connection_string TSRMLS_DC )
+                                                HashTable* options, const connection_option valid_conn_opts[], 
+                                                void* driver,_Inout_ std::string& connection_string TSRMLS_DC )
 {
-    bool credentials_mentioned = false;
     bool mars_mentioned = false;
```
yitam: where's "credentials_mentioned"?

source/shared/core_conn.cpp
```diff
@@ -654,6 +650,23 @@ void get_server_version( sqlsrv_conn* conn, char** server_version, SQLSMALLINT&
 // and return the string of the processor name.
 const char* get_processor_arch( void )
 {
+#ifndef _WIN32
```
yitam: Please double check if Apple also has such system info names. If not, it seems more straightforward to retain #ifdef __linux__
Also, all these strcmp() calls... is testing for equality, shouldn't we use (! strcmp(...))???

source/shared/core_results.cpp
```diff
@@ -20,9 +20,15 @@
...
 
 #include <functional>
-#include <iostream>
+
 #include <sstream>
 
+#ifndef _WIN32
+#include <type_traits>
+#include <uchar.h>
+#endif // _WIN32
```
yitam: Please correct the comment

source/shared/core_init.cpp
```diff
@@ -52,6 +53,7 @@ void core_sqlsrv_minit( sqlsrv_context** henv_cp, sqlsrv_context** henv_ncp, err
         LOG( SEV_ERROR, "Failed to retrieve Windows version information." );
         throw core::CoreException();
     }
+#endif
```
yitam: Please add the comment to ```#endif```

source/shared/core_results.cpp
yitam: yes @david-puglielli I put it there to explain our decision back then. In case this default behavior of SQL Server is changed in the future, we could 'revert' our code.

source/shared/core_results.cpp
```diff
@@ -81,78 +89,231 @@ SQLPOINTER read_lob_field( sqlsrv_stmt* stmt, SQLUSMALLINT field_index, sqlsrv_b
...
     }
 
-    *out_buffer_length = str_num.size() * sizeof( Char ); // str_num.size() already include the NULL terminator
+   *out_buffer_length = str_num.size() * sizeof( Char ); // str_num.size() already include the NULL terminator
     memcpy_s( buffer, buffer_length, str_num.c_str(), *out_buffer_length );
 
     return SQL_SUCCESS;
+#else
+    std::basic_string<char> str_num;
+    SQLRETURN r;
```
yitam: Just curious, have you confirmed that the code for string conversions written for Linux doesn't work in Windows?

source/shared/core_results.cpp
```diff
@@ -189,7 +350,7 @@ sqlsrv_error* odbc_get_diag_rec( sqlsrv_stmt* odbc, SQLSMALLINT record_number )
     // convert the error into the encoding of the context
     sqlsrv_malloc_auto_ptr<SQLCHAR> sql_state;
     SQLLEN sql_state_len = 0;
-    if (!convert_string_from_utf16( enc, wsql_state, sizeof(wsql_state), (char**)&sql_state, sql_state_len )) {
+    if ( !convert_string_from_utf16( enc, wsql_state, SQL_SQLSTATE_BUFSIZE, (char**)&sql_state, sql_state_len )) {
```
yitam: What's SQL_SQLSTATE_BUFSIZE? Originally it was the size of wsql_state? Just doesn't sound right...

source/shared/core_results.cpp
```diff
@@ -561,7 +723,6 @@ sqlsrv_buffered_result_set::sqlsrv_buffered_result_set( sqlsrv_stmt* stmt TSRMLS
 
             row_count++;
             if( *out_buffer_length == SQL_NULL_DATA ) {
-                unsigned char* null_bits = reinterpret_cast<unsigned char*>( row );
```
yitam: Why is this line gone?

source/shared/core_results.cpp
```diff
@@ -665,14 +826,12 @@ SQLRETURN sqlsrv_buffered_result_set::get_data( SQLUSMALLINT field_index, SQLSMA
         return SQL_SUCCESS;
     }
 
+    
     // check to make sure the conversion type is valid
-    if( conv_matrix.find( meta[ field_index ].c_type ) == conv_matrix.end() ||
-        conv_matrix.find( meta[ field_index ].c_type )->second.find( target_type ) == 
-            conv_matrix.find( meta[ field_index ].c_type )->second.end() ) {
-
+    conv_matrix_t::const_iterator conv_iter = conv_matrix.find(meta[field_index].c_type);
```
yitam: Please put spaces before and after the square brackets and parenthesis

source/shared/core_results.cpp
```diff
@@ -1164,14 +1352,19 @@ SQLRETURN sqlsrv_buffered_result_set::wide_to_system_string( SQLSMALLINT field_i
...
+#else								 			
         BOOL default_char_used = FALSE;
         char default_char = '?';
 
-        // allocate enough to handle WC -> DBCS conversion if it happens
-        temp_string = reinterpret_cast<SQLCHAR*>( sqlsrv_malloc( field_len, sizeof(char), sizeof(char)));
         temp_length = WideCharToMultiByte( CP_ACP, 0, (LPCWSTR) field_data, static_cast<int>(field_len / sizeof(WCHAR)),
                                            (LPSTR) temp_string.get(), static_cast<int>(field_len), &default_char, &default_char_used );
-        
+#endif	// !_WIN32
```
yitam: Add spaces before and after brackets, parentheses, etc

source/shared/core_sqlsrv.h
```diff
@@ -997,6 +1045,8 @@ struct sqlsrv_conn : public sqlsrv_context {
     sqlsrv_conn( SQLHANDLE h, error_callback e, void* drv, SQLSRV_ENCODING encoding  TSRMLS_DC ) :
         sqlsrv_context( h, SQL_HANDLE_DBC, e, drv, encoding )
     {
+        server_version = SERVER_VERSION_UNKNOWN;
+        driver_version = ODBC_DRIVER_13;
```
yitam: Is it a good idea to set default to ODBC_DRIVER_13? What about ODBC 11?
How about DRIVER_VERSION_UNKNOWN or something like that?

source/shared/core_sqlsrv.h
```diff
@@ -2046,6 +2082,18 @@ namespace core {
...
+        // On Linux platform
+        // DriverName: libmsodbcsql-13.0.so.0.0
+        // DriverODBCVer: 03.52
+        // DriverVer: 13.00.0000
+        // unixODBC: 2.3.1
+        // r = ::SQLRowCount( stmt->handle(), &rows_affected );  
+        // returns r=-1 for an empty result set. 
+#ifdef __linux__
+        if( r == -1 && rows_affected == -1 )
+           return 0;
```
yitam: So this block only applies to Linux platform? Also, the version numbers in the comments are outdated

source/shared/core_sqlsrv.h
```diff
@@ -2263,7 +2311,7 @@ namespace core {
     inline void sqlsrv_zend_hash_add( sqlsrv_context& ctx, HashTable* ht, zend_string* key, unsigned int key_len, zval* data, 
                                       unsigned int data_size, zval* pDest TSRMLS_DC )
     {
-        int zr = (pDest = ::zend_hash_add(ht, key, data)) != NULL ? SUCCESS : FAILURE;
+        int zr = ::zend_hash_add(ht, key, data) != NULL ? SUCCESS : FAILURE;
```
yitam: Why is pDest no longer needed?

source/shared/core_sqlsrv.h
```diff
@@ -2283,4 +2331,20 @@ sqlsrv_conn* allocate_conn( SQLHANDLE h, error_callback e, void* driver TSRMLS_D
 
 } // namespace core
 
+// connection attribute functions
+template <unsigned int Attr>
+struct str_conn_attr_func {
+
```
yitam: Why is this last block outside of "namespace core"?

source/shared/core_stmt.cpp
```diff
@@ -233,6 +233,7 @@ sqlsrv_stmt* core_sqlsrv_create_stmt( sqlsrv_conn* conn, driver_stmt_factory stm
 {
 	sqlsrv_malloc_auto_ptr<sqlsrv_stmt> stmt;
     SQLHANDLE stmt_h = SQL_NULL_HANDLE;
+    sqlsrv_stmt* return_stmt;
```
yitam: Just a suggestion, when declaring a pointer variable, set it to NULL by default

source/shared/core_stmt.cpp
```diff
@@ -922,20 +920,18 @@ void core_sqlsrv_get_field( sqlsrv_stmt* stmt, SQLUSMALLINT field_index, sqlsrv_
...
-					efree( field_value );
-					field_value = NULL;
-					*field_len = 0;
-				}
-			}
+                   sqlsrv_phptype invalid;
+                   invalid.typeinfo.type = SQLSRV_PHPTYPE_INVALID;
+                   for( int i = stmt->last_field_index + 1; i < field_index; ++i ) {
+                       SQLSRV_ASSERT( reinterpret_cast<field_cache*>( zend_hash_index_find_ptr( Z_ARRVAL( stmt->field_cache ), i )) == NULL, "Field already cached." );
+                       core_sqlsrv_get_field( stmt, i, invalid, prefer_string, field_value, field_len, cache_field, sqlsrv_php_type_out TSRMLS_CC );
```
yitam: In quite a lot of places 'NULL' is replaced by '0'. Why don't we do the same here? 
Also, I assume the variable 'cached' is not necessary?

source/shared/core_stmt.cpp
```diff
@@ -1066,8 +1062,6 @@ void core_sqlsrv_post_param( sqlsrv_stmt* stmt, zend_ulong param_num, zval* para
     // PDO doesn't need the reference count, but sqlsrv does since the stream can be live after sqlsrv_execute by sending it
     // with sqlsrv_send_stream_data.
     if( zend_hash_index_exists( Z_ARRVAL( stmt->param_streams ), param_num )) {
-        sqlsrv_stream* stream_encoding = NULL;
-        stream_encoding = reinterpret_cast<sqlsrv_stream*>(zend_hash_index_find_ptr(Z_ARRVAL(stmt->param_streams), param_num));
```
yitam: Why are these two lines not needed?

source/shared/core_sqlsrv.h
Hadis-Fard: This was ifdef inux and the only change is changing to Win32

source/shared/core_stmt.cpp
```diff
@@ -1171,15 +1165,22 @@ void core_sqlsrv_set_query_timeout( sqlsrv_stmt* stmt, long timeout TSRMLS_DC )
         
         // a query timeout of 0 indicates "no timeout", which means that lock_timeout should also be set to "no timeout" which 
         // is represented by -1.
-        long lock_timeout = (( timeout == 0 ) ? -1 : timeout * 1000 /*convert to milliseconds*/ );
+        int lock_timeout = (( timeout == 0 ) ? -1 : timeout * 1000 /*convert to milliseconds*/ );
 
         // set the LOCK_TIMEOUT on the server.
         char lock_timeout_sql[ 32 ];
+#ifndef _WIN32		
+        int written = snprintf( lock_timeout_sql, sizeof( lock_timeout_sql ), "SET LOCK_TIMEOUT %d",
```
yitam: Since we are not using format characters, is it possible to use memcpy_s() instead, thus getting rid of ```ifndef``` here

source/shared/core_conn.cpp
Hadis-Fard: it hasn't been used

source/shared/core_stmt.cpp
```diff
@@ -1266,11 +1265,17 @@ bool core_sqlsrv_send_stream_packet( sqlsrv_stmt* stmt TSRMLS_DC )
...
-                int wsize = MultiByteToWideChar( stmt->current_stream.encoding, MB_ERR_INVALID_CHARS, 
-                                                 buffer, static_cast<int>( read ), wbuffer, static_cast<int>( sizeof( wbuffer ) / sizeof( wchar_t )));
-                if( wsize == 0 && GetLastError() == ERROR_NO_UNICODE_TRANSLATION ) {
+                SQLWCHAR wbuffer[ PHP_STREAM_BUFFER_SIZE + 1 ];
+				DWORD last_error_code = ERROR_SUCCESS;
+				// buffer_size is the # of wchars.  Since it set to stmt->param_buffer_size / 2, this is accurate
+#ifndef _WIN32
+                int wsize = SystemLocale::ToUtf16Strict( stmt->current_stream.encoding, buffer, static_cast<int>(read), wbuffer, static_cast<int>(sizeof( wbuffer ) / sizeof( SQLWCHAR )), &last_error_code );
+#else
+                int wsize = MultiByteToWideChar( stmt->current_stream.encoding, MB_ERR_INVALID_CHARS, buffer, static_cast<int>( read ), wbuffer, static_cast<int>( sizeof( wbuffer ) / sizeof( wchar_t )));
```
yitam: For consistency, should we use ```SQLWCHAR``` rather than ```wchar_t```?
If so, we could write a single line for the division of ```sizeof( wbuffer ) / sizeof( SQLWCHAR ))
Besides, should we consider floor() or ceil() before casting the result to an integer?

source/shared/core_conn.cpp
Hadis-Fard: it is Unix compatible structure defined in sys/utsname.h

source/shared/core_results.cpp
Hadis-Fard: hasn't been used!

source/shared/core_sqlsrv.h
Hadis-Fard: So far yes, it should be tested on other platform to see if it exhibits the same behaviors

source/shared/core_sqlsrv.h
Hadis-Fard: because str_conn_attr_func in not defined in core namespace. it should be after core is defined otherwise GCC complains about it

source/shared/core_util.cpp
```diff
@@ -261,8 +251,15 @@ bool core_sqlsrv_get_odbc_error( sqlsrv_context& ctx, int record_number, sqlsrv_
...
+#ifndef _WIN32
+            // In linux we need to calculate number of characters
+            SQLLEN wsqlstate_len = sizeof( wsqlstate ) / sizeof( SQLWCHAR );
+#else
+           // In Windows we need the size in bytes
+           SQLLEN wsqlstate_len = sizeof( wsqlstate );
+#endif // !_WIN32
             SQLLEN sqlstate_len = 0;
-            convert_string_from_utf16(enc, wsqlstate, sizeof(wsqlstate), (char**)&error->sqlstate, sqlstate_len);
+            convert_string_from_utf16(enc, wsqlstate, wsqlstate_len, (char**)&error->sqlstate, sqlstate_len);
```
yitam: should wsqlstate_len be SQL_SQLSTATE_BUFSIZE?

source/shared/core_sqlsrv.h
Hadis-Fard: the default is 13, later in the code if 13 hasn't been found the it sets it to 11

source/shared/core_sqlsrv.h
v-dareck: It was never used by the function.

source/shared/core_stmt.cpp
v-dareck: stream_encoding is not used.

source/shared/core_stmt.cpp
v-dareck: the string parameter is using %d...  `"SET LOCK_TIMEOUT %d"`

source/shared/core_sqlsrv.h
yitam: Just noticed that this helper method is not used anywhere in the driver...

source/shared/core_stmt.cpp
yitam: you're right... I missed that

source/shared/core_stmt.cpp
Hadis-Fard: zend_hash_index_find_ptr returns NULL

source/shared/core_stmt.cpp
Hadis-Fard: hasn't been used

source/shared/core_stmt.cpp
Hadis-Fard: We're using format characters in `SET LOCK_TIMEOUT %d`, however we can get rid of `ifdef` by using snprintf for both platforms.

source/shared/core_stmt.cpp
Hadis-Fard: +1 for `SQLWCHAR`

Issue #224, opened 2017-01-12
PDOStatement::getColumnMeta() crash/segfault


v-kigos:
PHP call for $stmt->getColumnMeta($index) will crash in Windows, and produce
segmentation fault in Linux if index is out of the range (exceeds the number of columns).

Repro script:
```
<?php
/* Connect to an SQL server */
$conn = new PDO("sqlsrv:server=$server", $user, $password);  

/* Retrieve metadata for a column */
$stmt = $conn->query("select name from sys.databases");  
$metadata = $stmt->getColumnMeta(2);
var_dump($metadata);  
?>
```

yitam:
It also crashes when prepare() is used instead of query() when the column is valid.

v-kigos:
MSDN does not define return values for this case, 
see https://msdn.microsoft.com/en-us/library/ff628182(v=sql.105).aspx

PHP help suggests: Returns FALSE if the requested column does not exist in the result set, or if no result set exists, see http://php.net/manual/en/pdostatement.getcolumnmeta.php

yitam:
Fixed by #261

eaglebush:
My development system is Windows 7 SP1 and the PDO_SQLSRV I use is 4.1.4. The only problem I am having is PDOStatement::getColumnMeta() crash/segfault #224, fixed in 4.3.0. However, this version does not support Windows 7 SP1. Can't we have an update on PDO_SQLSRV 4.1.4 of this bug?

yitam:
Yes @eaglebush , starting with v4.3.0 we stopped supporting Windows 7. That means we will no longer test our drivers in Windows 7 platforms, but it doesn't mean the drivers will break. Please feel free to try out v4.3.0 in your Windows 7 SP1.

PR #261, opened 2017-01-27
safeguard get_col_meta

yitam:
this is meant to fix issue #224


v-kigos:
MSDN does not define return values for this case,
see https://msdn.microsoft.com/en-us/library/ff628182(v=sql.105).aspx
PHP help suggests: Returns FALSE if the requested column does not exist in the result set, or if no result set exists, see http://php.net/manual/en/pdostatement.getcolumnmeta.php

yitam:
Yes, did you read the warning message on PHP.net? :)
So you suggest that we simply return FALSE without any message?

PR #269, opened 2017-02-01
Merge PHP-7.0-Linux changes into merged-codebase

v-dareck:


source/shared/core_conn.cpp
```diff
@@ -143,7 +162,20 @@ sqlsrv_conn* core_sqlsrv_connect( sqlsrv_context& henv_cp, sqlsrv_context& henv_
             }
 
             SQLSMALLINT output_conn_size;
-            r = SQLDriverConnectW( conn->handle(), NULL, reinterpret_cast<SQLWCHAR*>(wconn_string.get()), static_cast<SQLSMALLINT>( wconn_len ), NULL, 0, &output_conn_size, SQL_DRIVER_NOPROMPT );
+#ifdef __linux__
```
Hadis-Fard: please change to `_WIN32` if it is not specif to Linux, and might apply to Mac too, and add comments on `endif`s `#endif // !_WIN32` or `#endif // _WIN32`,  also apply it to other use of `_linux_`

PR #279, opened 2017-02-07
define SQL_GUID for PECL packages

yitam:


source/shared/core_sqlsrv.h
```diff
@@ -131,6 +131,9 @@ OACR_WARNING_POP
 #include <sqlext.h>
 #endif // _WIN32
 
+//From sqlext.h, add this constant for PECL
+#define SQL_GUID            (-11)
```
Hadis-Fard: undef the macro prior the defintion to prevent macro redefinition

yitam: Made the changes already. Please check.

Issue #267, opened 2017-02-01
[Linux] PDO driver should not throw exception on "closeCursor()"


deeky666:
The PDO driver throws the following exception when trying to close a prepared statement cursor with `PDOStatement::closeCursor()` if the statement has not been executed yet.

```
The statement must be executed before results can be retrieved.
```

While this is not completely wrong, the behaviour does not adhere to the [PDO interface](http://php.net/manual/de/pdostatement.closecursor.php). The contract clearly states that in case of an error `false` is to be returned. Exceptions not allowed.
This is especially important when trying to build database abstractions where consistent behaviour across different PDO implementation is desired. We currently have discovered that issue in our testsuite (Doctrine DBAL). While all PDO implementations we support return `false`, `pdo_sqlsrv` throws an exception.

```php
$connection = new \PDO(/* connection parameters */);
$statement = $connection->prepare('SELECT foo FROM bar');
$statement->closeCursor();
```

Any chance we can change the driver's behaviour here?

ulvii:
Hi @deeky666 ,

Could you confirm if this issue has been resolved with the latest release? https://github.com/Microsoft/msphpsql/releases/tag/v4.1.8-preview

PR #281, opened 2017-02-09
Github #267 close cursor error

v-dareck:
Fix for #267.
Check that statement has been executed before calling core_sqlsrv_next_result.

### DIFF
--- a/source/pdo_sqlsrv/pdo_stmt.cpp
+++ b/source/pdo_sqlsrv/pdo_stmt.cpp
@@ -401,14 +401,13 @@ int pdo_sqlsrv_stmt_close_cursor(pdo_stmt_t *stmt TSRMLS_DC)
 
         // to "close the cursor" means we make the statement ready for execution again.  To do this, we 
         // skip all the result sets on the current statement.
-		// If the statement has not been executed there are no next results to iterate over.
-		if ( driver_stmt->executed == true )
-		{
-			while( driver_stmt->past_next_result_end == false ) {
-
-				core_sqlsrv_next_result( driver_stmt TSRMLS_CC );
-			}
-		}
+        // If the statement has not been executed there are no next results to iterate over.
+        if ( driver_stmt->executed == true )
+        {
+            while( driver_stmt->past_next_result_end == false ) {
+                core_sqlsrv_next_result( driver_stmt TSRMLS_CC );
+            }
+        }
     }
     catch( core::CoreException& ) {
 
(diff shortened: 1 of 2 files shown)


source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -401,9 +401,12 @@ int pdo_sqlsrv_stmt_close_cursor(pdo_stmt_t *stmt TSRMLS_DC)
...
         // to "close the cursor" means we make the statement ready for execution again.  To do this, we 
         // skip all the result sets on the current statement.
-        while( driver_stmt->past_next_result_end == false ) {
-
-            core_sqlsrv_next_result( driver_stmt TSRMLS_CC );
+        // If the statement has not been executed there are no next results to iterate over.
+        if ( driver_stmt->executed == true )
+        {
+            while( driver_stmt->past_next_result_end == false ) {
+                core_sqlsrv_next_result( driver_stmt TSRMLS_CC );
```
deeky666: Btw isn't there a more efficient way using the internal equivalent to `sqlsrv_cancel()`?
http://php.net/manual/en/function.sqlsrv-cancel.php

v-dareck: I'm not sure as PHP does basically same.
http://php.net/manual/en/pdostatement.closecursor.php

Your original error was caused by the driver internally trying to perform operations on a statement that has not been executed. So since nothing has been executed there is no results to iterate over or close.

deeky666: Sure. But as stated in the documentation looping the cursor until the end is only the "fallback" solution if no more efficient approach is available. Since looking at [sqlsrv_cancel](https://github.com/Microsoft/msphpsql/blob/a02691e277d5804c3d8b28fcdfee31e6fd2df8ab/source/sqlsrv/stmt.cpp#L1308-L1335) seems to utilize an optimized native solution, couldn't that fit in here, too? That looks better than looping through until the end of the cursor is reached.

yitam: Hi @deeky666 the suggested fix by @v-dareck is actually more efficient to fix the problem you originally raised because there's nothing to cancel in your repro scenario. 

Besides, in normal circumstances when the query has actually been executed closeCursor() will do the clean up and finalize output parameters etc.

Please let us know if this have clarified things for you.

deeky666: @yitam I did not recognize the "if executed" condition, shame on me. Thanks for pointing this out! No complaints from my side then :P

yitam: Glad to be of help @deeky666 :)

Issue #231, opened 2017-01-13
String truncation when binding varchar(max)


david-puglielli:
When binding output parameters as varchar(max) or nvarchar(max), a string truncation error message may result. This does not happen if the argument is not 'max'. This problem is seen on both Windows and Linux.

```php
<?php 
$server = "myserver";
$connectionInfo = array( "Database"=>"$databaseName", "UID"=>"$uid", "PWD"=>"$pwd");

$tableName = "datatypes";
$columnNames = array( "c1","c2" );

$conn = sqlsrv_connect($server, $connectionInfo);

for ($k = 1; $k <= 6; $k++)
{
    $sqlType = GetSqlType($k);
    $dataType = "[$columnNames[0]] int, [$columnNames[1]] $sqlType";
    
    $sql = "CREATE TABLE [$tableName] ($dataType)";
    $stmt1 = sqlsrv_query($conn, $sql);
    
    $data = "LongStringForTesting";
    
    $dataValues = array($k, $data);
    
    $tsql = "INSERT INTO [$tableName] ($columnNames[0], $columnNames[1]) VALUES (?, ?)";    
    $stmt2 = sqlsrv_query( $conn, $tsql, $dataValues );

    ExecProc($conn, $tableName, $columnNames, $k, $data, $sqlType);
    
    $stmt3 = sqlsrv_query($conn, "DROP TABLE [$tableName]");
    sqlsrv_free_stmt($stmt1);
    sqlsrv_free_stmt($stmt2);
    sqlsrv_free_stmt($stmt3);
}

sqlsrv_close($conn);


function ExecProc($conn, $tableName, $columnNames, $k, $data, $sqlType)
{
    $phpDriverType = GetDriverType($k, strlen($data));
    
    $spArgs = "@p1 int, @p2 $sqlType OUTPUT";
    $spCode = "SET @p2 = ( SELECT c2 FROM $tableName WHERE c1 = @p1 )";
    $procName = "testBindOutSp";

    $stmt1 = sqlsrv_query($conn, "CREATE PROC [$procName] ($spArgs) AS BEGIN $spCode END");
    sqlsrv_free_stmt($stmt1);
    
    $callArgs = "?, ?";
    
    // Data to initialize $callResult variable. This variable should be different than inserted data in the table
    $initData = "ShortString";
    $callResult = $initData;
    
    $params = array( array( $k, SQLSRV_PARAM_IN ), 
[...]

PR #293, opened 2017-02-17
For issue #231, force column size to be determined

yitam:
Issue #231: When binding params, in the function determine_column_size_or_precision(), column size is set to SQL_SS_LENGTH_UNLIMITED (0). Then later in resize_output_buffer_if_necessary() when calculating the required size, the buffer length is smaller than required, which results in truncation error. This affects not only varchar(max), but also nvarchar(max), varbinary(max), and xml types


yitam:
The problem described in issue #231 also affects InOut parameters, and for varbinary(max), the encoding is not set properly. The new test will test both output and inout parameters plus these data types with unlimited sizes

Issue #260, opened 2017-01-27
PDO statement returns bool(false) when using temporary tables


v-kigos:
PDO_SQLSRV fails to successfully execute CREATE / SELECT statements 
when using temporary tables, see repro script. This behaviour 
is cross-platform (Win/Linux) on PHP7.0 and 7.1.

SQLSRV driver does not have this issue and returns: 
resource(7) of type (SQL Server Statement)

Repro script:

```
// Connect
$conn = new PDO("sqlsrv:server=$serverName", $username, $password);

// Create a temporary table
$sql = "CREATE TABLE #T1 (code VARCHAR(12))";
$stmt = $conn->query($sql);

// Insert data fails, and returns bool(false)
$sql = "INSERT INTO #T1 VALUES ('ONE')";
$stmt = $conn->query($sql);
var_dump($stmt);
```

Expected output:
resource(7) of type (SQL Server Statement)
string(11) "ONE"

Actual output:
bool(false)

yitam:
The main difference is that PDO invokes query() by first preparing the statement then execute it, whereas in sqlsrv we execute the query directly by SQLExecDirect().

As explained in this [article](https://technet.microsoft.com/en-us/library/ms175528(v=sql.105).aspx):
> Prepared statements cannot be used to create temporary objects on SQL Server. Prepared statements cannot reference system stored procedures that create temporary objects, such as temporary tables. These procedures must be executed directly.

PR #296, opened 2017-02-21
Remove uchar.h

v-dareck:
Replace uchar.h functions with SystemLocale FromUtf16 ToUtf16 calls.

### DIFF
--- a/source/shared/core_results.cpp
+++ b/source/shared/core_results.cpp
@@ -25,7 +25,6 @@
 
 #ifndef _WIN32
 #include <type_traits>
-#include <uchar.h>
 #endif // !_WIN32
 
 
@@ -160,6 +159,27 @@ SQLRETURN copy_buffer( _Out_ void* buffer, SQLLEN buffer_length, _Out_ SQLLEN* o
 }
 #endif // !_WIN32
 
+
+#ifndef _WIN32
+
+size_t charFromUtf16(const WCHAR src, char * dest, size_t cchDest, DWORD * pErrorCode)
+{
+    return SystemLocale::FromUtf16( CP_ACP, &src, 1, dest, cchDest, NULL, pErrorCode );
+}
+
+size_t charFromUtf16(const char src, char * dest, size_t cchDest, DWORD * pErrorCode)
+{
+    if (cchDest > 1)
+    {
+        dest[0] = src;
+        dest[1] = '\0';
+        return 1;
+    }
+    return 0;    
+}
+
+#endif // !_WIN32
+
 // convert a number to a string using locales
 // There is an extra copy here, but given the size is short (usually <20 bytes) and the complications of
 // subclassing a new streambuf just to avoid the copy, it's easier to do the copy
@@ -210,21 +230,20 @@ SQLRETURN number_to_string( Number* number_data, _Out_ void* buffer, SQLLEN buff
 
     if ( std::is_same<Char, char16_t>::value )
     {
+
         std::basic_string<char16_t> str;
-        char *str_num_ptr = &str_num[0], *str_num_end = &str_num[0] + str_num.size();
-
-        for ( const auto &mb : str_num )
+                
+        for (const auto &mb : str_num )
         {
-            char16_t ch16;
-            std::mbstate_t mbs = std::mbstate_t();
-
-            int len = mbrtoc16( &ch16, &mb, str_num_end - str_num_ptr, &mbs );
-            if ( len > 0 || len == -3 )
+            size_t cch = SystemLocale::NextChar( CP_ACP, &mb ) - &mb;
+            if ( cch > 0 )
             {
-                str.push_back( ch16 );
-                if ( len > 0 )
+                WCHAR ch16;
+                DWORD rc;                
+                size_t cchActual = SystemLocale::ToUtf16( CP_ACP, &mb, cch, &ch16, 1, &rc);
+                if (cchActual > 0)
                 {
-                    str_num_ptr += len;
+                    str.push_back ( ch16 );
                 }
             }
         }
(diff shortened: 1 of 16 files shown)


source/shared/core_results.cpp
```diff
@@ -160,6 +159,27 @@ SQLRETURN copy_buffer( _Out_ void* buffer, SQLLEN buffer_length, _Out_ SQLLEN* o
 }
 #endif // !_WIN32
 
+
+#ifndef _WIN32
+
+size_t charFromUtf16(const WCHAR src, char * dest, size_t cchDest, DWORD * pErrorCode)
+{
+    return SystemLocale::FromUtf16( CP_ACP, &src, 1, dest, cchDest, NULL, pErrorCode );
```
yitam: Is CP_ACP the only choice even if 'src' is WCHAR?

source/shared/core_results.cpp
```diff
@@ -210,21 +230,20 @@ SQLRETURN number_to_string( Number* number_data, _Out_ void* buffer, SQLLEN buff
...
+        std::basic_string<char16_t> str;
+                
+        for (const auto &mb : str_num )
         {
-            char16_t ch16;
-            std::mbstate_t mbs = std::mbstate_t();
-
-            int len = mbrtoc16( &ch16, &mb, str_num_end - str_num_ptr, &mbs );
-            if ( len > 0 || len == -3 )
+            size_t cch = SystemLocale::NextChar( CP_ACP, &mb ) - &mb;
```
yitam: Where's the third parameter of SystemLocale::NextChar() ? A default? What's its value?

source/shared/core_results.cpp
```diff
@@ -160,6 +159,27 @@ SQLRETURN copy_buffer( _Out_ void* buffer, SQLLEN buffer_length, _Out_ SQLLEN* o
...
+size_t charFromUtf16(const WCHAR src, char * dest, size_t cchDest, DWORD * pErrorCode)
+{
+    return SystemLocale::FromUtf16( CP_ACP, &src, 1, dest, cchDest, NULL, pErrorCode );
+}
+
+size_t charFromUtf16(const char src, char * dest, size_t cchDest, DWORD * pErrorCode)
+{
+    if (cchDest > 1)
+    {
+        dest[0] = src;
```
yitam: What if dest is null? Or cchDest is 0?
Looks like this method is never used, correct?

test/pdo_sqlsrv/pdo_fetch_cursorBuffered_binary.phpt
```diff
@@ -0,0 +1,105 @@
...
+require_once("autonomous_setup.php");
+$conn = new PDO( "sqlsrv:server=$serverName", $username, $password);
+$conn->setAttribute( PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION );
+$sample = 'asdgasdgasdgsadg';
+
+$query = 'CREATE TABLE #TESTTABLE (exist varbinary(max))';
+$stmt = $conn->exec($query);
+$query = 'INSERT INTO #TESTTABLE VALUES(:p0)';
+$stmt = $conn->prepare($query);
+$stmt->bindParam(':p0', $sample, PDO::PARAM_LOB, 0, PDO::SQLSRV_ENCODING_BINARY);
```
yitam: Unfortunately PDO doesn't support temporary tables...

test/pdo_sqlsrv/pdo_fetch_cursorBuffered_multicolumns.phpt
```diff
@@ -0,0 +1,271 @@
...
+$conn = new PDO( "sqlsrv:server=$serverName", $username, $password);
+$conn->setAttribute( PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION );
+$sample = 1234567890.1234;
+$sample1 = -1234567890.1234;
+$sample2 = 1;
+$sample3 = -1;
+$sample4 = 0.5;
+$sample5 = -0.55;
+
+$query = 'CREATE TABLE #TESTTABLE (a float(53), neg_a float(53), b int, neg_b int, c decimal(16, 6), neg_c decimal(16, 6), zero int, zerof float(53), zerod decimal(16,6))';
```
yitam: Would you consider adding a 'real' column or a 'numeric' column?

test/pdo_sqlsrv/pdo_fetch_cursorBuffered_multicolumns.phpt
```diff
@@ -0,0 +1,271 @@
...
+$stmt = $conn->exec($query);
+$query = 'INSERT INTO #TESTTABLE VALUES(:p0, :p1, :p2, :p3, :p4, :p5, 0, 0, 0)';
+$stmt = $conn->prepare($query);
+$stmt->bindValue(':p0', $sample, PDO::PARAM_INT);
+$stmt->bindValue(':p1', $sample1, PDO::PARAM_INT);
+$stmt->bindValue(':p2', $sample2, PDO::PARAM_INT);
+$stmt->bindValue(':p3', $sample3, PDO::PARAM_INT);
+$stmt->bindValue(':p4', $sample4, PDO::PARAM_INT);
+$stmt->bindValue(':p5', $sample5, PDO::PARAM_INT);
+$stmt->execute();
```
yitam: for bindValue() we don't have to supply the third parameter (it's actually confusing because most of these values are not integers anyway)

source/shared/core_results.cpp
v-dareck: it's the destination CP and internally it is changed to UTF8.
charFromUtf16 is used to get around string_to_number compile errors.

test/pdo_sqlsrv/pdo_fetch_cursorBuffered_multicolumns.phpt
v-dareck: real is just float(24)

test/pdo_sqlsrv/pdo_fetch_cursorBuffered_multicolumns.phpt
v-dareck: If they aren't bound as int then they are retrieved as strings which may skip the conversion functions.

source/shared/core_results.cpp
v-dareck: mb is null terminated so it is not necessary.

source/shared/core_results.cpp
v-dareck: removed.

test/pdo_sqlsrv/pdo_fetch_cursorBuffered_binary.phpt
yitam: See issue #260

source/shared/core_results.cpp
```diff
@@ -257,6 +236,39 @@ SQLRETURN number_to_string( Number* number_data, _Out_ void* buffer, SQLLEN buff
...
+    while ( source[i] )
+    {        
+        memset( c_str, 0, sizeof( c_str ));        
+        DWORD rc;    
+        int cch = 0;
+        errno_t err = mplat_wctomb_s( &cch, c_str, sizeof( c_str ), source[i++] );
+        if (cch > 0 && err == 0)
+        {
+            str.append( std::string( c_str, cch ) );
+        }
```
yitam: Great! What if err != 0? Would you return an error?

v-dareck: no it that case the character is skipped. ERROR_SUCCESS is used instead of 0.

Issue #270, opened 2017-02-02
[Linux] Fetching VARBINARY value with pdo_sqlsrv does not work


deeky666:
I am not quite sure if this is a bug or I just cannot figure out how to do this. Fetching a value from a `VARBINARY` column does not return the same value that was inserted before when using `pdo_sqlsrv`. Expected result would be the binary representation that was inserted into the database but instead the hexadecimal representation is returned.

**Test**

```php
$sql = "CREATE TABLE foo (icon VARBINARY(MAX))";
$conn->exec($sql);

$icon = base64_decode(<<<EOF
H4sICJRACVgCA2RvY3RyaW5lLmljbwDtVNtLFHEU/ia1i9fVzVWxvJSrZmoXS6pd0zK7QhdNc03z
lrpppq1pWqJCFERZkUFEDybYBQqJhB6iUOqhh+whgl4qkF6MfGh+s87O7GVmO6OlBfUfdIZvznxn
fpzznW9gAI4unQ50XwirH2AAkEygEuIwU58ODnPBzXGv14sEq4BrwzKKL4sY++SGTz6PodcutN5x
IPvsFCa+K9CXMfS/cOL5OxesN0Wceygho0WAXVLwcUJBdDVDaqOAij4Rrz640XlXQmAxQ16PHU63
iqdvXbg4JOHLpILBUSdM7XZEVDDcfuZEbI2ASaYguUGAroSh97GMngcSeFFFerMdI+/dyGy1o+GW
Ax5FxfAbFwoviajuc+DCIwn+RTwGRmRIThXxdQJyu+z4/NUDYz2DKCsILuERWsoQfoQhqpLhyhMZ
XfcknBmU0NLvQArpTm0SsI5mqKqKuFoGc8cUcjrtqLohom1AgtujQnapmJJU+BbwCLIwhJXyiKlh
MB4TkFgvIK3JjrRmAefJm+77Eiqvi+SvCq/qJahQyWuVuEpcIa7QLh7Kbsourb9b66/pZdAd1voz
fCNfwsp46OnZQPojSX9UFcNy+mYJNDeJPHtJfqeR/nSaPTzmwlXar5dQ1adpd+B//I9/hi0xuCPQ
Nkvb5um37Wtc+auQXZsVxEVYD5hnCilxTaYYjsuxLlsxXUitzd2hs3GWHLM5UOM7Fy8t3xiat4fb
[...]

PR #297, opened 2017-02-22
issue #270 fix

Hadis-Fard:
Fix to retrieve the data as inserted in binary columns

### DIFF
--- a/source/pdo_sqlsrv/pdo_stmt.cpp
+++ b/source/pdo_sqlsrv/pdo_stmt.cpp
@@ -721,12 +721,10 @@ int pdo_sqlsrv_stmt_get_col_data(pdo_stmt_t *stmt, int colno,
         sqlsrv_phptype sqlsrv_php_type;
         SQLSRV_ASSERT( colno >= 0 && colno < static_cast<int>( driver_stmt->current_meta_data.size()),
                        "Invalid column number in pdo_sqlsrv_stmt_get_col_data" );
-        sqlsrv_php_type = driver_stmt->sql_type_to_php_type( static_cast<SQLUINTEGER>( driver_stmt->current_meta_data[ colno ]->field_type ),
-                                                             static_cast<SQLUINTEGER>( driver_stmt->current_meta_data[ colno ]->field_size ),
-                                                             true );
 
         // set the encoding if the user specified one via bindColumn, otherwise use the statement's encoding
-        sqlsrv_php_type.typeinfo.encoding = driver_stmt->encoding();
+        sqlsrv_php_type = driver_stmt->sql_type_to_php_type( static_cast<SQLUINTEGER>( driver_stmt->current_meta_data[colno]->field_type ),
+                                                            static_cast<SQLUINTEGER>( driver_stmt->current_meta_data[colno]->field_size ), true );
 
         // if a column is bound to a type different than the column type, figure out a way to convert it to the 
         // type they want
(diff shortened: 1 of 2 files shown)


test/pdo_sqlsrv/pdo_270_fetch_binary.phpt
```diff
@@ -0,0 +1,95 @@
...
+require_once("autonomous_setup.php");
+
+$tableName = 'test_binary';
+$columns = array( 'col1', 'col2', 'col3', 'col4');
+
+// Connect
+$conn = new PDO( "sqlsrv:server=$serverName", "$username", "$password" );
+
+// CREATE database
+$conn->query("CREATE DATABASE ". $dbName) ?: die();
```
yitam: No need to create a database. It suffices to create a table name unique enough (e.g. using timestamp).

test/pdo_sqlsrv/pdo_270_fetch_binary.phpt
```diff
@@ -0,0 +1,95 @@
...
+	$result = "";
+	$stmt = $conn->query($sql);  
+	$result = $stmt->fetchObject();
+	if( strncmp($result->$columnName, $input, $len) !== 0){
+		print_r("\nRetrieving using fetchObject failed");
+	}
+
+	$result = "";
+	$stmt = $conn->query($sql);  
+	$result = $stmt->fetchALL( PDO::FETCH_COLUMN );
```
yitam: typo? fetchAll()

test/pdo_sqlsrv/pdo_270_fetch_binary.phpt
```diff
@@ -0,0 +1,95 @@
...
+//free connection
+$conn=null;
+
+print_r("Test finished successfully");
+
+//calls various fetch methods
+function test_fetch($conn, $tableName, $columnName, $input){
+		
+    $len = strlen($input);
+	$result = "";
```
yitam: indentation a little off?

test/pdo_sqlsrv/pdo_270_fetch_binary.phpt
```diff
@@ -0,0 +1,95 @@
...
+	$sql = "SELECT $columnName from $tableName";
+	
+	$stmt = $conn->query($sql);  
+	$stmt->bindColumn(1, $result, PDO::PARAM_LOB);
+	$stmt->fetch(PDO::FETCH_BOUND);
+	//binary is fixed size, to evaluate output, compare it using strncmp
+	if( strncmp($result, $input, $len) !== 0){
+		print_r("\nRetrieving using bindColumn failed");
+	}
```
yitam: Perhaps testing equality like ($result === $input) is enough?

source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -1306,6 +1304,8 @@ sqlsrv_phptype pdo_sqlsrv_stmt::sql_type_to_php_type( SQLINTEGER sql_type, SQLUI
                        "Invalid encoding on the connection.  Must not be invalid or default." );
     }                
 
+    sqlsrv_phptype.typeinfo.encoding = local_encoding;
+
     switch( sql_type ) {
```
yitam: Does this change mean that each field can have its own encoding set via bindColumn?

Hadis-Fard: it sets the encoding to stmt or conn encoding and in case of binary or invalid it is set to binary or invalid encoding respectively.

test/pdo_sqlsrv/pdo_270_fetch_binary.phpt
Hadis-Fard: binary column is fixed size, this is just testing that value is not hex number anymore

test/pdo_sqlsrv/pdo_270_fetch_binary.phpt
Hadis-Fard: it is better to have self-contained exportable tests, and since the db is dropped at the end of the test, doesn't matter if the table name is unique or not

PR #306, opened 2017-02-25
removed unnecessary conditional in sqlsrv/conn.cpp

luagod:


source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -1305,6 +1325,9 @@ int pdo_sqlsrv_dbh_quote( pdo_dbh_t* dbh, const char* unquoted, size_t unquoted_
 // This method is not implemented by this driver.
 pdo_sqlsrv_function_entry *pdo_sqlsrv_get_driver_methods( pdo_dbh_t *dbh, int kind TSRMLS_DC )
 {
+	if(!dbh)
+		return NULL;
+
```
v-kaywon: These checks for if(!dbh) is not necessary since dbh is always passed in from the pdo side and it is always not null

source/sqlsrv/conn.cpp
```diff
@@ -37,12 +37,7 @@ struct date_as_string_func {
 
         ss_sqlsrv_conn* ss_conn = static_cast<ss_sqlsrv_conn*>( conn );
 
-        if( zend_is_true( value )) {
-            ss_conn->date_as_string = true;
-        }
-        else {
-            ss_conn->date_as_string = false;
-        }
+        ss_conn->date_as_string = zend_is_true( value );
```
v-kaywon: OK

source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -1417,6 +1440,9 @@ void validate_stmt_options( sqlsrv_context& ctx, zval* stmt_options, _Inout_ Has
 
 void pdo_bool_conn_str_func::func(connection_option const* option, zval* value, sqlsrv_conn* /*conn*/, std::string& conn_str TSRMLS_DC )
 {
+    if(!option)
+	return;
+
```
v-kaywon: OK

v-kaywon:
Hi @leremix , I reviewed your PR and yes removing the unnecessary conditional in sqlsrv conn.cpp does make the code cleaner, thank you for you contribution. As for the checks for if(!dbh), I don't think they are necessary since dbh is built from the pdo side (see [pdo_dbh.c](https://github.com/v-kaywon/php-src/blob/master/ext/pdo/pdo_dbh.c#L354) from the pdo extension for constructing dbh). If dbh does not exist, it should be handled in the PDO side.

Issue #278, opened 2017-02-06
PDO::lastInsertId() not implemented correctly for sequences


deeky666:
`\PDO::lastInsertId($name = null)` should return the current value for a given sequence object name. [See the documentation](http://php.net/manual/en/pdo.lastinsertid.php).

Currently this is not working because the PDO implementation is using `SELECT IDENT_CURRENT(%s)` which is not intended for sequences but for tables. [See here](https://github.com/Microsoft/msphpsql/blob/master/source/pdo_sqlsrv/pdo_dbh.cpp#L33).

The current workaround is to extend `\PDO` like [we have done it in Doctrine](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php#L43-L53).

While this is working, it should rather be fixed in the driver itself IMO.

meet-bhagdev:
@deeky666 Your recommendation makes sense. We are still looking into it. If we do change the behavior, we might end up breaking applications that rely on this behavior. Do you think a lot of folks in the community would be using `PDO::lastInsertId($name = null)`  to return the current value of a table?

deeky666:
@meet-bhagdev I do not see why anyone should rely on a behaviour that is not defined. The method is not supposed to accept a table name. It is clearly documented in PHP. If people are really using it for tables, they are misusing the API. I see your objection about BC but IMO keep BC for a bug has nothing to do with BC ;) Does that make sense?

Also I have not come across another PDO implementation that was NOT using this API for sequences.

Hadis-Fard:
This issue is resolved in [5.0.0-preview](https://github.com/Microsoft/msphpsql/releases/tag/v5.0.0-preview) release.

PR #311, opened 2017-03-01
Github #278: change lastInsertId to return the last sequence number

v-kaywon:
Fix for #278.
Change lastInsertId to return the last sequence number when the sequence name is provided. Previously when a table name is provided, the driver returns the last id inserted of that table. Now when a table name is given, an empty string is returned.


v-kaywon:
@meet-bhagdev: this PR does not include the connection option, it simply change the lastInsertId function to return the last sequence number. Depending on the decision if a connection option for sequence number is wanted, the code will/will not be changed.

v-kaywon:
@yitam :  did that already :)

yitam:
So when the user is running an old sql server but supplies a name to lastInsertId() the method simply returns an empty string?

v-kaywon:
yes, in fast when the user tries to create a sequence, the stmt returned would be false

source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -30,7 +30,7 @@ namespace {
 
 const char LAST_INSERT_ID_QUERY[] = "SELECT @@IDENTITY;";
 const size_t LAST_INSERT_ID_BUFF_LEN = 10;    // size of the buffer to hold the string value of the last insert id integer
-const char TABLE_LAST_INSERT_ID_QUERY[] = "SELECT IDENT_CURRENT(%s)";
+const char TABLE_LAST_INSERT_ID_QUERY[] = "SELECT CURRENT_VALUE FROM SYS.SEQUENCES WHERE NAME=%s";
```
Hadis-Fard: it's better to change the `TABLE_LAST_INSERT_ID_QUERY` to Sequence.

Hadis-Fard:
@david-puglielli can we close this PR now?

david-puglielli:
Yes this PR can be closed.

Issue #282, opened 2017-02-09
Versions/tags/releases/stability


dzuelke:
Hi,

Now that the Windows/Linux codebases have been merged, I'd like to encourage some more structure and discipline ;) around version numbers, PECL release stabilities, and tags.

The version number recently jumped from 4.0.8 to 4.1.6. Unlike 4.0.7, which was marked "preview", the 4.0.8 release was called "production". Not sure where the 4.1.0-4.1.5 releases went? :)

No machine readable metadata indicates the stability on GitHub (e.g. using an "RC1" suffix), just on PECL, where the stability is "devel".

4.1.6 is, again, a "preview release"; there is a 4.1.6.1 on PECL, but no tag for it on GitHub. Was that just a re-packaging? Even if it was, I assume `package.xml` changed, so there should be a tag for it (but i the spirit of SemVer, "4.1.7" would be the better choice for a number).

Would it be possible to clean this all up a little?

If 4.1.6 brings new features, but is not yet stable, then why not call it 4.1.0beta1, …RC1, …RC2 and so forth instead, until a final 4.1.0 is out, where then maintenance updates can be made with minor fixes, while work on 4.2.0 begins?

Remember, the driver is not an end user software or product; it's a library. As such, it would really benefit from adhering to semantic versioning rules.

Or at least have release lines? Where 4.0.x and 4.2.x are the stable release lines, and 4.1.x and 4.3.x are the unstable/preview ones, from which stuff gets either backported, or folded into a future "even numbered" release, similar to how Nginx do it with their "stable" and "mainline" series?

meet-bhagdev:
@dzuelke Great point on using beta1, RC1, RC2 etc tags. We will look into it and start doing it if everything works out in our test labs.

The version number jump was because the Windows and Linux codebase merged after a long time :) We will not do this in the future.

4.1.6.1 was just repackaging due to a PECL issue. In hindsight, you are correct 4.1.7 would have been better 👍 Point taken.

dzuelke:
The other benefit of that is that version requirements on `ext-sqlsrv` and `ext-pdo_sqlsrv` in Composer packages become more reliable with beta/RC stability in the version number.

meet-bhagdev:
@dzuelke starting with our March releases, we will be using tags as per your recommendation. We will be starting with 4.1.7-preview.1 followed by 4.1.7-preview.2, ..... , 4.1.7-RC.1, 4.1.7-RC.2, 4.1.7. What do you think?

PR #315, opened 2017-03-04
updated versioning to semantic versioning 

Hadis-Fard:
related to #282


source/shared/version.h
```diff
@@ -24,13 +24,24 @@
...
+#define SEMVER_BUILDMETA
+
+#if SQLVERSION_BUILD > 0
+#undef SEMVER_BUILDMETA
+#define SEMVER_BUILDMETA "+" STRINGIFY( SQLVERSION_BUILD )
+#endif
+
+#define VER_FILEVERSION_STR     STRINGIFY( SQLVERSION_MAJOR ) "." STRINGIFY( SQLVERSION_MINOR ) "." STRINGIFY( SQLVERSION_PATCH ) SEMVER_PRERELEASE SEMVER_BUILDMETA
+#define _FILEVERSION            SQLVERSION_MAJOR,SQLVERSION_MINOR,SQLVERSION_PATCH,SQLVERSION_BUILD
+
```
yitam: so when SQLVERSION_BUILD is 1, VER_FILEVERSION_STR is 4.1.7-preview+1?

Hadis-Fard: that's true and when it is 0, `VER_FILEVERSION_STR`  is `4.1.7-preview`, and build number is not counted in precedence order.

yitam: How about Product name? Both drivers have the same name but file descriptions are different?

Hadis-Fard: will remain untouched

PR #319, opened 2017-03-08
Mac instructions

ulvii:
Added initial Mac install instructions.


meet-bhagdev:
@ulvii @Hadis-Fard Can we merge this ASAP? We don't have any public facing documentation for Mac right now.

Issue #330, opened 2017-03-19
Unexpected result for sqlsrv_num_rows


jansor:
Hi, I am not sure is this is a bug or the behavior is correct but when i execute below code the web server crashes. It provides me with Bad Gateway 

The original script that i used was much bigger but when i execute this test script i got the same issue:
`<?php
$server = "localhost";
$connectionInfo = ["UID" => "localhost", "PWD" => "localhost", 'Database' => 'localhost'];
$conn = sqlsrv_connect($server, $connectionInfo);

$stmt = sqlsrv_query($conn, "IF EXISTS (SELECT * FROM [sys].[objects] WHERE (name LIKE 'tbl_Test%') AND type in (N'U'))
                BEGIN
                    select 0
                END", [], ['Scrollable' => SQLSRV_CURSOR_CLIENT_BUFFERED]);

if ($stmt) {
    $hasRows = sqlsrv_has_rows($stmt);
    $numRows = sqlsrv_num_rows($stmt);
    var_dump($hasRows, $numRows);
}
?>`

Following scenario is expexted, when the table name like tbl_Test exists it would return the amount of tables that are found, hens sqlsrv_has_rows would be true and sqlsrv_num_rows would return the amount
When no tables are found sqlsrv_has_rows must be false, sqlsrv_num_rows would be zero. 

Thanks to look into this

PS i am using following setup
IIS10, PHP7.1.2  and SQL Driver 4.1.7-preview on an SQL 2016 server

v-kaywon:
Hello @jansor, we are able to reproduce the issue. Thank you for reporting to us.

The behavior that you expect when table name like 'tbl_Test%' exists is incorrect. For example, say you have three tables 'tbl_Test1', 'tbl_Test2' and 'tbl_Test3'. The query you have will return 0 due to the `select 0`. Thus `sqlsrv_num_rows` will return `1` and not `3` as you have expected. If you want to count the number of tables with name like 'tbl_Test%', you can try the following query:
`SELECT COUNT(*) from sys.objects where (name LIKE 'tbl_Test%') AND (type = 'U')`

jansor:
Yes you are correct, in my testing previous i had the select query from the if statement also inside the select 0 statement. Thanks for the confirmation its a bug, i was confused with the result set.

v-kaywon:
[...]

PR #333, opened 2017-03-20
fix crash from getting the numrow of a null buffered result set

v-kaywon:
fit for #330
fix crash from sqlsrv_num_rows to get the numrow in a client buffered result set that is null
added test

### DIFF
--- a/source/shared/core_results.cpp
+++ b/source/shared/core_results.cpp
@@ -911,6 +911,7 @@ SQLLEN sqlsrv_buffered_result_set::row_count( TSRMLS_D )
 		return zend_hash_num_elements( cache );
 	}
 	else {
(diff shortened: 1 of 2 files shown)


source/shared/core_results.cpp
```diff
@@ -907,7 +907,12 @@ SQLLEN sqlsrv_buffered_result_set::row_count( TSRMLS_D )
 {
     last_error = NULL;
 
-    return zend_hash_num_elements( cache );
+	if ( cache ) {
+		return zend_hash_num_elements( cache );
+	}
+	else {
+		return -1;
```
Hadis-Fard: Please add a comment that what does -1 mean

Issue #332, opened 2017-03-20
ISO C++ forbids converting a string constant to ‘char*’


shakaw:
/var/tmp/sqlsrv/conn.cpp:152:75: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
 char* ss_sqlsrv_conn::resource_name = static_cast<char *>("ss_sqlsrv_conn");

/var/tmp/sqlsrv/init.cpp: In function ‘int zm_activate_sqlsrv(int, int)’:
/opt/remi/php71/root/usr/include/php/Zend/zend_ini.h:139:62: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
 #define INI_INT(name) zend_ini_long((name), sizeof(name)-1, 0)
                                                              ^
/opt/remi/php71/root/usr/include/php/Zend/zend_ini.h:142:37: note: in expansion of macro ‘INI_INT’
 #define INI_BOOL(name) ((zend_bool) INI_INT(name))
                                     ^
/var/tmp/sqlsrv/init.cpp:639:45: note: in expansion of macro ‘INI_BOOL’
     SQLSRV_G( warnings_return_as_errors ) = INI_BOOL( INI_PREFIX INI_WARNINGS_RETURN_AS_ERRORS);
                                             ^
/opt/remi/php71/root/usr/include/php/Zend/zend_ini.h:139:62: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
 #define INI_INT(name) zend_ini_long((name), sizeof(name)-1, 0)
                                                              ^
/var/tmp/sqlsrv/init.cpp:640:32: note: in expansion of macro ‘INI_INT’
     SQLSRV_G( log_severity ) = INI_INT( INI_PREFIX INI_LOG_SEVERITY);
                                ^
/opt/remi/php71/root/usr/include/php/Zend/zend_ini.h:139:62: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
 #define INI_INT(name) zend_ini_long((name), sizeof(name)-1, 0)
                                                              ^
/var/tmp/sqlsrv/init.cpp:641:34: note: in expansion of macro ‘INI_INT’
     SQLSRV_G( log_subsystems ) = INI_INT( INI_PREFIX INI_LOG_SUBSYSTEMS);
                                  ^
[...]

PR #337, opened 2017-03-23
convert const char[] to char[] to avoid conversion warnings

v-kaywon:
#332

### DIFF
--- a/source/sqlsrv/init.cpp
+++ b/source/sqlsrv/init.cpp
@@ -636,14 +636,14 @@ PHP_RINIT_FUNCTION(sqlsrv)
     LOG_FUNCTION( "PHP_RINIT for php_sqlsrv" );
 
     // read INI settings
-	// need to convert const char[] to char[] to avoid converting string to char* warnings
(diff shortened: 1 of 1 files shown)


source/sqlsrv/init.cpp
```diff
@@ -636,7 +636,14 @@ PHP_RINIT_FUNCTION(sqlsrv)
     LOG_FUNCTION( "PHP_RINIT for php_sqlsrv" );
 
     // read INI settings
-    SQLSRV_G( warnings_return_as_errors ) = INI_BOOL( INI_PREFIX INI_WARNINGS_RETURN_AS_ERRORS );
+	// need to convert const char[] to char[] to avoid converting string to char* warnings
+	// cannot simply cast const char[] to char* since INI_INT needs the sizeof(param) and the size of char* is always 4 / 8 bytes
+	char warnings_as_errors[] = INI_PREFIX INI_WARNINGS_RETURN_AS_ERRORS;
```
Hadis-Fard: Indentations :smile:

PR #338, opened 2017-03-24
Connection resiliency 2

david-puglielli:


source/pdo_sqlsrv/config.m4
```diff
@@ -1,68 +1,68 @@
-PHP_ARG_WITH(pdo_sqlsrv, for pdo_sqlsrv support,
```
ulvii: Please exclude this file if you have not changed it.

source/shared/core_sqlsrv.h
```diff
@@ -1099,24 +1103,28 @@ const char SERVER[] = "Server";
...
-   SQLSRV_CONN_OPTION_MARS,
-   SQLSRV_CONN_OPTION_QUOTED_ID,
-   SQLSRV_CONN_OPTION_TRACE_FILE,
-   SQLSRV_CONN_OPTION_TRACE_ON,
-   SQLSRV_CONN_OPTION_TRANS_ISOLATION,
-   SQLSRV_CONN_OPTION_TRUST_SERVER_CERT,
-   SQLSRV_CONN_OPTION_WSID,
-   SQLSRV_CONN_OPTION_ATTACHDBFILENAME,
-   SQLSRV_CONN_OPTION_APPLICATION_INTENT,
-   SQLSRV_CONN_OPTION_MULTI_SUBNET_FAILOVER,
```
v-kaywon: Please fix indentation (somehow is has one less space thus it's saying you've modified these lines, while you'll only added the last 4 lines

source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -98,6 +102,24 @@ struct pdo_txn_isolation_conn_attr_func
     static void func( connection_option const* /*option*/, zval* value_z, sqlsrv_conn* conn, std::string& /*conn_str*/ TSRMLS_DC );
 };
 
+#ifdef _WIN32
+struct pdo_int_conn_str_func {
+
+    static void func( connection_option const* option, zval* value, sqlsrv_conn* /*conn*/, std::string& conn_str TSRMLS_DC )
+    {
+        TSRMLS_C;
+        SQLSRV_ASSERT( Z_TYPE_P( value ) == IS_STRING, "Wrong zval type for this keyword" )
```
v-kaywon: In conn.cpp, your assert message is "An integer is expected for this keyword", then should you change this one to "An string is expected for this keyword" just to be consistent?

test/sqlsrv/connection_resiliency_keywords.phpt
```diff
@@ -0,0 +1,245 @@
...
+        \(
+            \[0\] => IMSSP
+            \[SQLSTATE\] => IMSSP
+            \[1\] => -8
+            \[code\] => -8
+            \[2\] => An invalid connection option key type was received. Option key types must be strings.
+            \[message\] => An invalid connection option key type was received. Option key types must be strings.
+        \)
+
+\)
```
v-kaywon: These two tests are very long although they are just doing the same thing with different ConnectRetryCount and ConnectRetryInterval 9 times for each test. Probably looks cleaner if you can make a function and all it nine times with different params.

source/shared/core_sqlsrv.h
david-puglielli: But most of this file has 4 space indentations - this is one of only two places where the indentation was 3 spaces. Actually I think I should fix the indentation in the other place too (the SQLSRV_STMT_OPTIONS enum at line 1062).

source/pdo_sqlsrv/pdo_dbh.cpp
david-puglielli: Well, an integer is expected for it, but it's a string in the connection string. This function is based on pdo_int_conn_attr_func::func immediately below, which is used for handling the login timeout, also an integer, and uses a similar assert. In conn.cpp, the login timeout is handled as an integer in the zval, so I did it that way. Why the difference between the sqlsrv and pdo source, I don't know.

source/pdo_sqlsrv/config.m4
david-puglielli: I can only surmise that upon executing 'git push' the line endings did not get changed to CRLF. I fixed it by downloading the original and uploading it back using the web site instead.

test/sqlsrv/connection_resiliency_keywords.phpt
david-puglielli: Done!

david-puglielli:
Closing because we have a newer PR for connres.

PR #342, opened 2017-03-27
Update readme

ulvii:
On Ubuntu 16, we need to add extension to apache php.ini file too


ulvii:
Yes, to make sure that we have drivers installed before updating apache php.ini

ulvii:
Updating Ubuntu 16 Apache Configuration

ulvii:
Updating Ubuntu 16 Apache Configuration

Issue #189, opened 2016-11-23
Performance issue, using odbc is faster than sqlsrv with PDO


sean256:
While screwing around with trying to get connection pooling working I discovered that sqlsrv is much slower during execute than using odbc.

- OS: CentOS 7.2
- PHP: PHP 7.0.12
- SQL Driver: ODBC Driver 13 for SQL Server

Test code (in milliseconds)

```
$startTime = microtime(true);
function printElapsedTimeSinceLastEvent($event=''){
	global $startTime;
	$elapsed = microtime(true) - $startTime;
	echo $event . round($elapsed*1000) ."\n";
	$startTime = microtime(true);
}

// Connection sqlsrv driver
$pdo = new PDO ("sqlsrv:server=10.12.12.123;database=someDatabase",'someUser','somePassword');
printElapsedTimeSinceLastEvent("SQLSRV Connected: ");
$stmt = $pdo->prepare("SELECT TOP 10 * FROM users");
$stmt->execute();
printElapsedTimeSinceLastEvent("SQLSRV Execute: ");
$array = $stmt->fetchAll(PDO::FETCH_ASSOC);
printElapsedTimeSinceLastEvent("SQLSRV Fetch: ");

//Connection using ODBC driver
$pdo = new PDO ('odbc:DSN=someODBCDSN;UID=someUser;PWD=somePassword');
printElapsedTimeSinceLastEvent("ODBC Connected: ");
$stmt = $pdo->prepare("SELECT TOP 10 * FROM users");
$stmt->execute();
printElapsedTimeSinceLastEvent("ODBC Execute: ");
$array = $stmt->fetchAll(PDO::FETCH_ASSOC);
printElapsedTimeSinceLastEvent("ODBC Fetch: ");
```

Output:

```
SQLSRV Connected: 151
SQLSRV Execute: 109
SQLSRV Fetch: 1
ODBC Connected: 150
ODBC Execute: 58
ODBC Fetch: 0
```

meet-bhagdev:
@sean256 Very interesting find. Let us see what is going on here in our test labs. My hunch is that there is some conversion logic when using sqlsrv that causes the perf differences when comparing it to vanilla ODBC

richerlariviere:
I have the same slowness with PDO_SQLSRV. [Call graph](http://imgur.com/a/rJlDE). My profiling points to the PDOStatement::fetch method.

OS: Ubuntu 16.04.1 (tested on CentOS 7.2 too)
[...]

PR #343, opened 2017-03-27
Column meta cache 

ulvii:
Implemented Cache to store Column SQL types and Column Display Sizes, so that PDO SQLSRV does not call ODBC driver for every fetch. This change should improve the driver performance by 15-20% when fetching strings.

### DIFF
--- a/source/shared/core_stmt.cpp
+++ b/source/shared/core_stmt.cpp
@@ -1034,6 +1034,9 @@ void core_sqlsrv_next_result( sqlsrv_stmt* stmt TSRMLS_DC, bool finalize_output_
 
         close_active_stream( stmt TSRMLS_CC );
 
+        //Clear column sql types and sql display sizes.
+        zend_hash_clean( Z_ARRVAL( stmt->col_cache ));	    
+	    
         SQLRETURN r;
         if( throw_on_errors ) {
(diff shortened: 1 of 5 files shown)


v-kaywon:
Do you want to quickly run Dr. Memory or Valgrind on a simple fetching test to make sure the cache is cleaned up?

ulvii:
@v-kaywon , done.

source/shared/core_stmt.cpp
```diff
@@ -1014,6 +1034,9 @@ void core_sqlsrv_next_result( sqlsrv_stmt* stmt TSRMLS_DC, bool finalize_output_
 
         close_active_stream( stmt TSRMLS_CC );
 
+        //Clear column sql types and sql display sizes.
+        zend_hash_clean( Z_ARRVAL( stmt->col_cache ));	    
+
```
v-kaywon: Where does the cache get reinitialized after core_sqlsrv_next_result? The only initialization I see for the col_cache is in the sqlsrv_stmt constructor.

ulvii: zend_hash_clean() does not release the buckets, it removes the elements from a hashtable.

ulvii:
#189

Issue #336, opened 2017-03-22
PDO: Got error when exec 'DELETE' on something doesn't exist


Gounlaf:
Hi,

I got a weird behavior when I try to delete something that doesn't exist (table exist, but the row I want to deleted doesn't), with pdo driver.
On PHP 5.6, there were no problem.

Bellow a test script and results obtained on this system:

```
PHP 7.0.15-0ubuntu0.16.04.4 (cli) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.15-0ubuntu0.16.04.4, Copyright (c) 1999-2017, by Zend Technologies
```

```
$ pecl info sqlsrv
About pecl.php.net/sqlsrv-4.0.8
===============================
Release Type          PECL-style PHP extension (source code)
Name                  sqlsrv
Channel               pecl.php.net
Summary               Microsoft Drivers for PHP for SQL Server
                      (SQLSRV)
Description           The Microsoft Drivers for PHP for SQL Server are
                      PHP extensions that allow for the reading and
                      writing of SQL Server data from within PHP
                      scripts. The SQLSRV extension provides a
                      procedural interface while the PDO_SQLSRV
                      extension implements PDO for accessing data in
                      all editions of SQL Server 2005 and later
                      (including Azure SQL DB). These drivers rely on
                      the Microsoft ODBC Driver for SQL Server to
                      handle the low-level communication with SQL
                      Server.
                      *This package contains only the SQLSRV driver.*
Maintainers           Meet Bhagdev <meetb@microsoft.com> (lead)
                      Jay Kint <jaykint@microsoft.com> (lead,
                      inactive)
                      Marie Barwin <v-mabarw@microsoft.com> (lead)
                      Hadis Kakanejadi Fard <v-hakaka@microsoft.com>
                      (lead)
Release Date          2016-12-19 13:12:21
Release Version       4.0.8 (stable)
API Version           4.0.8 (stable)
License               The MIT License (MIT)
                      (https://opensource.org/licenses/mit)
Release Notes         [Added]
                      - Unicode Column name support.
                      [Fixed]
                      - Fixed issue with buffered cursor in PDO_SQLSRV
                      driver when CharacterSet is UTF-8.
                      - Fixed issue with empty output parameters on
                      stored procedure.
                      - Fixed memory leaks in buffered queries.
[...]

PR #344, opened 2017-03-28
workaround for unixODBC bug returning error when getting numcol

v-kaywon:
FIx for #336 
unixODBC returns error when getting the numcol after pdo::exec on a query that returns SQL_NO_DATA
workaround by returning 0 so pdo::exec doesn't throw an exception


v-kaywon:
@yitam , it should not do anything when it is NULL. I should put 

```
if ( r == -1 && stmt->executed && strcmp( reinterpret_cast<const char*>( error->sqlstate ), "HY010" ) == 0 )
  	 return 0;
```
inside the if block `if ( stmt->current_results != NULL ) `
to prevent sigmentation fault at `error->sqlstate`.

Thanks you for your sharp eyes :)

source/shared/core_sqlsrv.h
```diff
@@ -2040,7 +2040,21 @@ namespace core {
...
+
+        // Workaround for a bug in unixODBC: after SQLExecDirect returns SQL_NO_DATA,
+        // r = ::SQLNumResultCols( stmt->handle(), &num_cols );  
+        // returns r=-1 (SQL_ERROR) and error HY010 (Function sequence error)
+        // but it should have succeeded with r=0 (SQL_SUCCESS) and no error
+        // instead of throwing an exception, return 0 if the r=-1, stament has been executed, and has a HY010 error
+        // (HY010 error should not return if stmt->execute is true)
+#ifndef _WIN32
+        sqlsrv_error_auto_ptr error;
+        if ( stmt->current_results != NULL ) {
```
yitam: You don't have to make changes, but you can actually declare the variable error inside the IF block ;)

test/pdo_sqlsrv/pdo_336_pho_exec_empty_result_set_error.phpt
```diff
@@ -0,0 +1,53 @@
+--TEST--
+GitHub issue #336 - PDO::exec should not return an error with query returning SQL_NO_DATA
+--DESCRIPTION--
+Verifies GitHub issue 336 is fixed, PDO::exec on query returning SQL_NO_DATA will not give an error
+--SKIPIF--
+--FILE--
+<?php
+require_once("pdo_tools.inc");
+
```
yitam: Looks like you don't have to include "pdo_tools.inc", but since you have it, you might want to create temporary tables instead (no need to check for a table's existence then drop it later)

test/pdo_sqlsrv/pdo_336_pho_exec_empty_result_set_error.phpt
```diff
@@ -0,0 +1,53 @@
...
+$Statement = $conn->exec("IF OBJECT_ID('foo_table', 'U') IS NOT NULL DROP TABLE foo_table");
+$Statement = $conn->exec("CREATE TABLE foo_table (id BIGINT PRIMARY KEY NOT NULL IDENTITY, intField INT NOT NULL)");
+$Statement = $conn->exec("INSERT INTO foo_table (intField) VALUES(3)");
+
+//test prepare, not args
+$stmt = $conn->prepare($sql);
+$stmt->execute();
+if ($conn->errorCode() == 00000)
+    echo "prepare OK\n";
```
yitam: don't you need quotes around 00000?

source/shared/core_sqlsrv.h
```diff
@@ -2040,7 +2040,21 @@ namespace core {
...
         r = ::SQLNumResultCols( stmt->handle(), &num_cols );
-        
+
+        // Workaround for a bug in unixODBC: after SQLExecDirect returns SQL_NO_DATA,
+        // r = ::SQLNumResultCols( stmt->handle(), &num_cols );  
+        // returns r=-1 (SQL_ERROR) and error HY010 (Function sequence error)
+        // but it should have succeeded with r=0 (SQL_SUCCESS) and no error
+        // instead of throwing an exception, return 0 if the r=-1, stament has been executed, and has a HY010 error
+        // (HY010 error should not return if stmt->execute is true)
+#ifndef _WIN32
```
ulvii: What happens if you execute a large  batch of SQL statements, for example 10000 inserts? According to [this](https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlnumresultcols-function) page, SQLNumResultCols() might return HY010 when executing function is still executing.

test/pdo_sqlsrv/pdo_336_pho_exec_empty_result_set_error.phpt
```diff
@@ -0,0 +1,50 @@
...
+
+//test prepare, with args
+$stmt = $conn->prepare($sqlWithParameter);
+$stmt->execute(array(':id' => $sqlParameter));
+if ($conn->errorCode() == "00000")
+    echo "prepare with args OK\n";
+
+//test direct exec
+$stmt = $conn->exec($sql);
```
ulvii: I think the test would be more accurate, if we check the actual return value of `$conn->exec($sql);` too.

PR #351, opened 2017-03-31
Adding a script to test sqlsrv connection pooling on Linux and Mac

ulvii:


test/sqlsrv/isPooled.php
```diff
@@ -0,0 +1,25 @@
+<?php
+include_once 'autonomous_setup.php';
+$conn1 = sqlsrv_connect( $serverName, $connectionInfo);
+$connId1 = ConnectionID($conn1);
+$conn1 = null;
+
```
yitam: any reason you set $conn1 to null?
did you mean to use sqlsrv_close()?
same with stmt in ConnectionID() below... sqlsrv_free_stmt() should be used

test/sqlsrv/sqlsrv_ConnPool_Unix.phpt
```diff
@@ -0,0 +1,35 @@
...
+file_put_contents($odbcinst_ini, $current);
+
+//Creating a new php process, because for changes in odbcinst.ini file to affect pooling, drivers must be reloaded.
+print_r(shell_exec("php ./test/sqlsrv/isPooled.php"));
+
+//disable pooling by modifying the odbcinst.ini file
+$current = file_get_contents($odbcinst_ini);
+$current = str_replace("CPTimeout=5\n[ODBC]\nPooling=Yes\n",'',$current);
+file_put_contents($odbcinst_ini, $current);
+
```
yitam: would you save the original content of odbcinst.ini and use it to replace the modified content?

test/sqlsrv/isPooled.php
ulvii: in fact, does it matter if you close first connection or not in your test?
- That's how you pool connections.

test/sqlsrv/sqlsrv_ConnPool_Unix.phpt
ulvii: I think this is simple enough :)

PR #353, opened 2017-03-31
More tests

yitam:
new tests for negative scenarios (from test framework) + fix Katmai test


test/output.py
```diff
@@ -51,6 +51,8 @@ def TestFilename(line):
     currentPos = 0
     while True:
         currentPos = currentPos - 1
+        if currentPos == 0:
```
ulvii: can currentPos even be equal to 0?

test/pdo_sqlsrv/pdo_fetch_complex_transactions.phpt
```diff
@@ -0,0 +1,159 @@
...
+        $v1 = $i + 1;
+        $v2 = $v1 * 1.5;
+
+        $stmt->bindValue(1, $v1);
+        $stmt->bindValue(2, $v2);
+        $stmt->execute(); 
+    }
+}
+
+function FetchData($stmt, $tableName, $numRows, $fetchMode = false)
```
ulvii: what is this for? Looks like you set fetch mode back to PDO::FETCH_LAZY anyways

test/pdo_sqlsrv/pdo_fetch_complex_transactions.phpt
```diff
@@ -0,0 +1,159 @@
...
+
+    $conn->beginTransaction();
+    $stmtDelete->execute(); 
+
+    echo "Deletion aborted\n";  
+
+    return $numRows;    
+}
+
+function InsertData($conn, $tableName, $count)
```
ulvii: Don't we have a common file that we can put these functions in instead of redefining them for every test?

test/sqlsrv/sqlsrv_streams_null_nchar.phpt
```diff
@@ -0,0 +1,111 @@
...
+function NullStream_Char2Stream($conn)
+{
+    $tableName = GetTempTableName();
+    
+    // create another test table a char(512) column
+    $stmt = sqlsrv_query($conn, "CREATE TABLE $tableName ([c1_int] int, [c2_nchar] nchar(512))");
+    sqlsrv_free_stmt($stmt);
+
+    $fname = null;
+    $stmt = sqlsrv_query($conn, "INSERT INTO $tableName (c1_int, c2_nchar) VALUES (?, ?)", array(-187518515, &$fname));
```
ulvii: Can't we just create one table with all these types, instead of so much duplication?

test/pdo_sqlsrv/pdo_fetch_complex_transactions.phpt
yitam: no, near the end of the test I did set it to true

test/sqlsrv/sqlsrv_streams_null_nchar.phpt
yitam: ok I will look into this

test/pdo_sqlsrv/pdo_fetch_complex_transactions.phpt
ulvii: Down below, 
```
if ($fetchMode) {
    $stmt->setFetchMode(PDO::FETCH_LAZY);  
}
else{
    $result = $stmt->fetch(PDO::FETCH_LAZY)
}
```
How are if and else blocks different?

v-kaywon: It could be the test is testing to see if $stmt->setFetchMode(PDO::FETCH_LAZY) actually works to change to fetch mode

test/sqlsrv/sqlsrv_fetch_invalid_types.phpt
```diff
@@ -0,0 +1,142 @@
...
+}
```
v-kaywon: Is this function really needed here? Since it is just a one liner (although a very long time) and it's only being called in PopulateTestTable

test/sqlsrv/sqlsrv_streams_empty_char.phpt
```diff
@@ -0,0 +1,119 @@
...
+    sqlsrv_execute($stmt);
+    $fld = $value;
+    $result = sqlsrv_fetch($stmt);
+    $stream = sqlsrv_get_field($stmt, $fld, SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY));
+    var_dump($stream);
+    
+    sqlsrv_execute($stmt);
+    $result = sqlsrv_fetch($stmt);
+    $value = sqlsrv_get_field($stmt, $fld, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_BINARY));
+    var_dump($value);
```
v-kaywon: A little confusing to use $value as field index and then putting result from sqlsrv_get_field into $value. Why not just rename $value to $fld in the function definition and then var_dump a value at the end

test/pdo_sqlsrv/pdo_fetch_complex_transactions.phpt
yitam: yes, @ulvii these are different APIs

test/sqlsrv/sqlsrv_streams_empty_char.phpt
yitam: Good point. Thanks!

PR #356, opened 2017-04-01
Connection resiliency changes

david-puglielli:


source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -98,6 +102,24 @@ struct pdo_txn_isolation_conn_attr_func
     static void func( connection_option const* /*option*/, zval* value_z, sqlsrv_conn* conn, std::string& /*conn_str*/ TSRMLS_DC );
 };
 
+#ifdef _WIN32
+struct pdo_int_conn_str_func {
+
+    static void func( connection_option const* option, zval* value, sqlsrv_conn* /*conn*/, std::string& conn_str TSRMLS_DC )
+    {
+        TSRMLS_C;
+        SQLSRV_ASSERT( Z_TYPE_P( value ) == IS_STRING, "Wrong zval type for this keyword" )
```
ulvii: I think it makes more sense to assert  "!= IS_LONG "

david-puglielli: Other integer attributes are handled this way for PDO_SQLSRV. It is the underlying ODBC driver that will sort out if the data types are correct.

PR #362, opened 2017-04-07
Fixed sed commands for pecl php ini file configuration

CodyCodeman:
Mac sed doesn't support the \s character class.


ulvii:
Hi @CodyCodeman ,
Thank you for the PR. Would you please resend it on the dev branch?

yitam:
hi @CodyCodeman, thanks for your suggestions. It turns out this step is not necessary in Mac (see Pull Request #375 )
Also, we appreciate your feedback, and in case you have any new idea in the future, please avoid using the master branch. Use the dev branch instead. Thank you.

Issue #140, opened 2016-08-25
Binary column binding is broken with EMULATE_PREPARES


david-garcia-garcia:
Something similar was originally reported (and fixed) here:

https://github.com/Azure/msphpsql/issues/63

See attached repro script (similar to the one posted in https://github.com/Azure/msphpsql/issues/121).

http://pastebin.com/GYNtYewC

Current result:

Uncaught PDOException: SQLSTATE[42000]: [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]Implicit conversion from data type varchar to varbinary(max) is not allowed. Use the CONVERT function to run this query.

Expected result:

Hello world!

The issue is again with EMULATE_PREPARES, if you comment out the following line:

`$pdo_options[PDO::ATTR_EMULATE_PREPARES] = TRUE;`

The script will work as expected.

meet-bhagdev:
@david-garcia-garcia Thanks for reporting this. We are looking into it.

david-garcia-garcia:
Any update on this? I am comming accross this issue all the time, it makes it impossible to mix EMULATE_PREPARES with binary columns....

meet-bhagdev:
@david-garcia-garcia Thanks for pinging the thread. We are able to reproduce this and are looking into fixing this. Stay tuned for updates.

Hadis-Fard:
@david-garcia-garcia We are looking into this issue and #92, I have a couple of questions about emulate prepare, and would appreciate if you can help me to understand how emulate prepare works in other PDO drivers.
- Are you able to set the encoding for parameters with emulated prepare in other PDO Drivers? As PDO replaces the placeholders with values itself.
- do  other PDO drivers with emulated prepare provide `$driver_options` in [PDOStatement::bindParam](https://msdn.microsoft.com/en-us/library/ff628166(v=sql.105).aspx), if so what are those options?
- Are you able to use bind param without encoding options, and encode the values before using them in bindParam?

david-garcia-garcia:
> help me to understand how emulate prepare works in other PDO drivers.

I'll do my best, but the truth is that the PDO for MS SQL is the only one I've been involved with - ever. To my understanding, switching on EMULATE_PREPARES should only impact how the underlying sql statements are built and sent to the database engine, but not how the developer interacts with the PDO driver itself.

> Are you able to set the encoding for parameters with emulated prepare in other PDO Drivers? As PDO replaces the placeholders with values itself.

[...]

Issue #92, opened 2016-05-08
Using emulate prepares breaks string parameters that contain unicode characters


david-garcia-garcia:
Someone reported this bug here:

https://www.drupal.org/node/2693763

The issue is that a statement without EMULATE_PREPARES generates the following SQL:

`SELECT * FROM users WHERE name = @P1 AND status = 1',N'@P1 nvarchar(6)',N'가각'`

But when using EMULATE_PREPARES it turns it into:

`SELECT * FROM users WHERE name = '가각' AND status = 1`

When it should be (notice the additional N):

`SELECT * FROM users WHERE name = N'가각' AND status = 1`

There is another issue related to EMULATE_PREPARES not working as expected:

https://github.com/Azure/msphpsql/issues/63

meet-bhagdev:
@david-garcia-garcia we believe this particular issue with the SQL queries listed above is fixed in our latest release - https://github.com/Azure/msphpsql/releases/tag/4.1.0. We are reviewing the other issues you have opened. In the meanwhile can you help us validate this?

david-garcia-garcia:
I can confirm this is not solved :(

I've built a complete test script:

http://pastebin.com/TsLt8Za3

Current Output:

> FOUND: ??sample2
> FOUND: 가각ácasa

Expected Output:

> FOUND: 가각ácasa
> FOUND: 가각ácasa
> FOUND: 가각ácasa
> FOUND: 가각ácasa

david-garcia-garcia:
Repasted the demo script, so that it won't expire as it has happened in other issues:

https://pastebin.com/P8emQQj4

david-garcia-garcia:
I can confirm that latest release fixes this issue. Thanks!!!!!

PR #371, opened 2017-04-21
implemented emulate prepare and added tests

v-kaywon:
Fix for #92 and #140 .

Implement emulate prepare by:
1. When a statement is prepared with emulate prepare, parse out placeholders and store it in a placeholders hashtable in stmt
2. Get statement information in the quote function
3. Going through the placeholders in order, get the encoding from stmt->bound_params
4. Set the encoding to param level encoding, stmt level encoding, or connection level encoding; param encoding dominate, then stmt level, then connection level. If one level's encoding is invalid, then go to the higher level encoding.
5. "quote" the parameter base on the encoding
6. Replace placeholders in the sql string with proper quoted/encoded parameters

Note: Emulate prepare only works for input bound parameters.
When user specify PDO::PARAM_INPUT_OUTPUT in PDO::bindParam when emulate prepare is on, the driver throws a PDO exception.
When user binds an output parameter, the sql query should be a placeholder preceding an '=' sign (e.g., "SELECT **? =** COUNT(*) FROM Employee WHERE Department='Human Resources'"). When while parsing out the placeholders out from the sql string, it also checks if the sql query has an '=' immediately after a placeholder. If it does, throws an exception saying output parameter with emulate prepared statement is not supported. If user binds an output parameter to a stored procedure, an error would not be raise.


david-garcia-garcia:
Thaaaaanks!! Just an question... What happens with duplicate placeholders in the updated implementation? On the previous driver you could not have a placeholder twice, yet on other PDO drivers such as MySQL they will let you re-use parameteres by allowing a placeholder to be repeated.

v-kaywon:
@david-garcia-garcia ,

Can you clarify on what you mean by duplicated placeholders?

Do you mean having two placeholders with the same name in a sql query?
For example.
```
$statement = $connection->prepare("SELECT * FROM Table1 WHERE uid=:id AND pid=:id");
$statement->bindParam(':id', 1234);
```
And both ':id' placeholders in the sql query is replaced with 1234.

Or do you mean binding to the same placeholder twice?
For example,
```
$statement = $connection->prepare("SELECT * FROM Table1 WHERE uid=:id");
$statement->bindParam(':id', 1234);
$statement->bindParam(':id', 6789);
```

david-garcia-garcia:
I'm talking about your first example werwe the placeholder is present twice in the query itself.

v-kaywon:
I did some preliminary testing. Seems like duplicated placeholders works for parameters of any encoding types that's not binary.

source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -1279,13 +1289,67 @@ int pdo_sqlsrv_dbh_quote( pdo_dbh_t* dbh, const char* unquoted, size_t unquoted_
     PDO_VALIDATE_CONN;
     PDO_LOG_DBH_ENTRY;
 
-    pdo_sqlsrv_dbh* driver_dbh = reinterpret_cast<pdo_sqlsrv_dbh*>( dbh->driver_data );
-    SQLSRV_ENCODING encoding = driver_dbh->bind_param_encoding;
+    SQLSRV_ENCODING encoding = SQLSRV_ENCODING_CHAR;
+	
+    // get the current object in PHP; this distinguishes pdo_sqlsrv_dbh_quote being called from:
+	// 1. PDO::quote() - object name is PDO
+	// 2. PDO::execute() - object name is PDOStatement
```
Hadis-Fard: I think you meant PDOStatement::execute 😉

Issue #378, opened 2017-04-28
sqlsrv output param shows garbage when the variable is initialized to a different type


yitam:
Tested in Windows, 32bit and 64bit, both NTS. 
First, create a simple stored procedure like this
```
create proc test_out @p1 integer, @p2 integer, @p3 integer output
as
begin
	select @p3 = @p1 + @p2
	print @p3
end
go
```

Then, in the script, set warnings return as errors then initialize the output param to a non-integer as shown below:

```
sqlsrv_configure( 'WarningsReturnAsErrors', true );

$v1 = 1;
$v2 = 2;
$v3 = 'str';

$stmt = sqlsrv_query( $conn, "{call test_out( ?, ?, ? )}", array( $v1, $v2, array( &$v3, SQLSRV_PARAM_OUT )));
 
echo "Output: $v3\n";
```

The output is ```Output: 3``` followed by lots of garbage.

To fix this issue, simply do the following:
```sqlsrv_configure( 'WarningsReturnAsErrors', false );```

Then we will get ```Output: 3```, as expected.

Hadis-Fard:
Fixed in in PR #381

PR #381, opened 2017-05-03
issue #378

Hadis-Fard:
#378

### DIFF
--- a/test/pdo_sqlsrv/pdo_378_out_param_error.phpt
+++ b/test/pdo_sqlsrv/pdo_378_out_param_error.phpt
@@ -1,0 +1,60 @@
+--TEST--
+This test verifies that GitHub issue #378 is fixed in pdo_sqlsrv.
+--DESCRIPTION--
+GitHub issue #378 - output parameters appends garbage info when variable is initialized with different data type
+steps to reproduce the issue:
+1- create a store procedure with print and output parameter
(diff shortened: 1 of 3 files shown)


test/pdo_sqlsrv/pdo_378_out_param_error.phpt
```diff
@@ -0,0 +1,60 @@
...
+	$stmt = $conn->exec($sp_sql);
+	if ($stmt === false) { print("Failed to create stored procedure"); }
+}
+
+function executeSP($conn, $procName){
+	$expected = 3;
+	$stmt = $conn->prepare("{call $procName( ?, ?, ? )}");
+	$stmt->bindParam(1, $v1);
+	$stmt->bindParam(2, $v2);
+	$stmt->bindParam(3, $v3, PDO::PARAM_INT, 10);
```
v-kaywon: I remember you've mentioned this before, but does the user has to specify the length when binding output parameters? (the msdn example also did that same thing)

Hadis-Fard: Yeah, the length should be set, otherwise driver won't consider it as output param

v-kaywon: should probably update the MSDN page to let the users know

PR #382, opened 2017-05-03
Added Authentication keyword for Azure Active Directory Authentication

david-puglielli:


source/shared/core_sqlsrv.h
```diff
@@ -2352,4 +2354,26 @@ struct str_conn_attr_func {
...
 
+struct str_conn_attr_auth_func {
+
+    static void func( connection_option const* option, zval* value, sqlsrv_conn* conn, std::string& conn_str TSRMLS_DC )
+    {
+        try {
+            SQLSRV_ASSERT( Z_TYPE_P( value ) == IS_STRING, "str_conn_attr_auth_func: Unexpected zval type." );
+            
+            std::string aads = Z_STRVAL_P( value );
+            SQLSRV_ASSERT( aads == "SqlPassword" || aads == "ActiveDirectoryPassword" || aads == "ActiveDirectoryIntegrated",
```
Hadis-Fard: `ActiveDirectoryIntegrated` is only supported in windows, `SqlPassword` is the default in ODBC if you don't specify anything, is there any reason we should support it?

source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -200,6 +201,15 @@ const connection_option PDO_CONN_OPTS[] = {
...
         conn_str_append_func::func 
     },
+    { 
+        PDOConnOptionNames::Authentication,
+        sizeof( PDOConnOptionNames::Authentication ),
+        SQLSRV_CONN_OPTION_AUTHENTICATION,
+        ODBCConnOptions::Authentication,
+        sizeof( ODBCConnOptions::Authentication ),
+        CONN_ATTR_STRING,
+        str_conn_attr_auth_func::func
```
Hadis-Fard: Please use `conn_str_append_func`

source/shared/core_sqlsrv.h
```diff
@@ -2352,4 +2354,26 @@ struct str_conn_attr_func {
     }
 };
 
+struct str_conn_attr_auth_func {
```
Hadis-Fard: no need to add a function for authentication, validation and error checking should be done in driver level not core or header, see validate_conn_options function (https://github.com/Microsoft/msphpsql/blob/dev/source/sqlsrv/conn.cpp#L151)

source/sqlsrv/conn.cpp
```diff
@@ -274,6 +275,15 @@ const connection_option SS_CONN_OPTS[] = {
         conn_str_append_func::func
     },
     {
+        SSConnOptionNames::Authentication,
+        sizeof( SSConnOptionNames::Authentication ),
+        SQLSRV_CONN_OPTION_AUTHENTICATION,
+        ODBCConnOptions::Authentication,
+        sizeof( ODBCConnOptions::Authentication ),
+        CONN_ATTR_STRING,
+        str_conn_attr_auth_func::func
```
Hadis-Fard: please use conn_str_append_func

source/shared/core_sqlsrv.h
```diff
@@ -2352,4 +2354,26 @@ struct str_conn_attr_func {
...
+
+    static void func( connection_option const* option, zval* value, sqlsrv_conn* conn, std::string& conn_str TSRMLS_DC )
+    {
+        try {
+            SQLSRV_ASSERT( Z_TYPE_P( value ) == IS_STRING, "str_conn_attr_auth_func: Unexpected zval type." );
+            
+            std::string aads = Z_STRVAL_P( value );
+            SQLSRV_ASSERT( aads != "ActiveDirectoryIntegrated",
```
Hadis-Fard: Use exception not assert, using predefined macros such as CHECK_CUSTOM_ERROR

Hadis-Fard: the error messages should be registered in error array, you can find similar definitions in util and pdo_util (https://github.com/Microsoft/msphpsql/blob/dev/source/sqlsrv/util.cpp#L51)

PR #386, opened 2017-05-08
investigating Authentication=SqlPassword issue on appveyor

yitam:
the equivalent test in pdo_sqlsrv passed in appveyor but for unknown reason it failed to connect using sqlsrv

### DIFF
--- a/test/pdo_sqlsrv/pdo_azure_ad_authentication.phpt
+++ b/test/pdo_sqlsrv/pdo_azure_ad_authentication.phpt
@@ -4,13 +4,13 @@ Test the Authentication keyword with options SqlPassword and ActiveDirectoryInte
 
 --FILE--
 <?php
-require_once("autonomous_setup.php");
+require_once("MsSetup.inc");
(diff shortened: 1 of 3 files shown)


test/pdo_sqlsrv/pdo_azure_ad_authentication.phpt
```diff
@@ -20,15 +20,15 @@ catch( PDOException $e )
     echo "\n";
 }
 
-$stmt = $conn->query( "SELECT name FROM master.dbo.sysdatabases" );
+$stmt = $conn->query( "SELECT count(*) FROM cd_info" );
```
v-kaywon: Why change from fetching string to integer? It may change what the test is originally testing (although I'm sure our driver has no problem fetching strings)

yitam: this meant to test if connection actually works by using a simple query, but if you like I can change to fetching texts

PR #395, opened 2017-05-12
fix uninitialized read if pos of string parser is -1

v-kaywon:
Fix uninitialized read in sql_string_parser if the pos of the parser is -1

### DIFF
--- a/source/pdo_sqlsrv/pdo_parser.cpp
+++ b/source/pdo_sqlsrv/pdo_parser.cpp
@@ -415,8 +415,9 @@ void conn_string_parser:: parse_conn_string( TSRMLS_D )
 // Primary function which parses out the named placeholders from a sql string.
 void sql_string_parser::parse_sql_string( TSRMLS_D ) {
     try {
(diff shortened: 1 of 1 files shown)


source/pdo_sqlsrv/pdo_parser.cpp
```diff
@@ -417,7 +417,9 @@ void sql_string_parser::parse_sql_string( TSRMLS_D ) {
     try {
         while ( !this->is_eos() ) {
             int start_pos = -1;
-
+            if ( pos == -1 ) {
+                next();
```
yitam: where is 'pos' defined?

v-kaywon: In the constructor of the sql_string_parser (https://github.com/Microsoft/msphpsql/blob/dev/source/pdo_sqlsrv/pdo_parser.cpp#L41), and pos is initialized as -1. I didn't want to initialize it to 0 since conn_string_parser has it initilized as -1

PR #397, opened 2017-05-15
Xplat cleanup

david-puglielli:


source/shared/xplat.h
```diff
@@ -162,59 +85,19 @@ using std::max;
...
-typedef unsigned __int64 windowsULongLong_t;
-#else // *nix (!WwoWH)
 #include <stdint.h> // Use standard bit-specific types (signed/unsigned integrated)
 typedef int32_t windowsLong_t;
 typedef uint32_t windowsULong_t;
 typedef int64_t windowsLongLong_t;
 typedef uint64_t windowsULongLong_t;
-#endif
-typedef windowsLong_t LONG, *PLONG, *LPLONG;
+typedef windowsLong_t LONG;
```
yitam: so I suppose we don't need *PLONG or *LPLONG?

source/shared/xplat.h
```diff
@@ -234,162 +117,48 @@ typedef windowsULongLong_t ULONGLONG;
 // source code shared with Windows SNAC (which is ODBC 3.8) compiles.
 #define ODBCVER 0x0380
```
yitam: Do we need to define ODBCVER?

source/shared/xplat.h
david-puglielli: They're not used anywhere in the driver code, so right now those typedefs are not needed.

source/shared/xplat.h
david-puglielli: Actually no, I'll remove it.

ulvii: Please be careful when removing ODBCVER. We still should be able to compile the drivers without having to specify --with-odbcver option

yitam: I believe in appveyor it configures without that odbcver option already, so we will see

ulvii: its not about appveyor. PECL build does not use --with-odbcver configure option.

Issue #127, opened 2016-08-06
PHP Fatal error:  Unexpected SQL type encountered in calc_string_size


xalopp:
On Linux, I get with the `sqlsrv` driver a Fatal error with following Fatal error

```
PHP Fatal error:  Unexpected SQL type encountered in calc_string_size
```

when I execute this query:

``` sql
select current_value from sys.sequences
```

The same query works well with the PDO driver `pdo_sqlsrv`.

Linux Version: Ubuntu 16.04, PHP installed as package.
PHP Version: `PHP 7.0.9-1+deb.sury.org~xenial+1 (cli) ( NTS )`
`pdo_sqlsrv` `ExtensionVer => 4.0.2.0`
`sqlsrv` `ExtensionVer => 4.0.2.0`

xalopp:
I use the scripts https://github.com/xalopp/vagrant-php7sqlsrv/blob/master/tests/pdo_sqlsrv_query.php and https://github.com/xalopp/vagrant-php7sqlsrv/blob/master/tests/sqlsrv_query.php for testing.

```
$ MSSQL_SQL='select current_value from sys.sequences' /vagrant/tests/sqlsrv_query.php
PDO sqlsrv connection successfull: yes
PDO sqlsrv query successfull: yes
PHP Fatal error:  Unexpected SQL type encountered in calc_string_size. in /vagrant/tests/sqlsrv_query.php on line 31
```

meet-bhagdev:
Hello @xalopp,

We are trying to repro this in our test lab with no luck. Can you add $count to your script and share your results?


$count = 0;
while ($row = sqlsrv_fetch_array($result, SQLSRV_FETCH_ASSOC)) {
                $count++;
                // var_dump($row);
}
sqlsrv_free_stmt($result);
echo "rows = ".$count."\n";

xalopp:
I forget to mention I'm using:
```
Microsoft SQL Server 2016 (RTM) - 13.0.1601.5 (X64) 
	Apr 29 2016 23:23:58 
	Copyright (c) Microsoft Corporation
	Enterprise Evaluation Edition (64-bit) on Windows Server 2012 R2 Essentials 6.3 <X64> (Build 9600: ) (Hypervisor)
``

**Comment by xalopp** [comment]
```
sqlcmd -S 192.168.56.102 -d AdventureWorks2016CTP3 -U test -P test -Q 'create sequence test123 start with 123'
```

```
[...]

Issue #51, opened 2015-05-21
Unable to handle SQL_VARIANT types


david-garcia-garcia:
The PDO driver is chocking on SQL_VARIANT types, but this can contain themselves simple types such as char, intenger, etc...

Steps to reproduce:
1. Create an UDF:

CREATE FUNCTION [dbo].[GREATEST](@op1 sql_variant, @op2 sql_variant) RETURNS sql_variant AS
BEGIN
  DECLARE @result sql_variant
  SET @result = CASE WHEN @op1 >= @op2 THEN @op1 ELSE @op2 END
  RETURN @result
END
GO
1. Run this query through the PDO

SELECT dbo.GREATEST(5, 6)
1. Theses are the errors you get:

Without Buffered Queries:

SQLSTATE[IMSSP]: Invalid type.

With Buffered Queries:

Fatal error: Unknown type in sqlsrv_buffered_query::sqlsrv_buffered_query

I can manage to get it to work if I use PDO::PARAM_INT on the column before the fetch, but there is no way that I can infer this informatino from the PDO metadata (getColumnMeta) that is reporting:

sql_srv_type = sql_variant
native_type = string

Possible solutions.....
1. At least let me retrieve the data as a string by default, instead of Fatal crashing.

stevebauman:
This is still an issue today. Is there a fix in the works?

Hadis-Fard:
@stevebauman PHP drivers rely on ODBC to handle SQL data types, and SQL_VARIANT is not supported in ODBC. We're looking into David's suggestion to see if SQL variants can be handled with string inferred type.

stevebauman:
@Hadis-Fard Awesome! Thanks for the reply :).

Hadis-Fard:
@stevebauman @david-garcia-garcia we added support for sql variants in [4.2.0-preview] (https://github.com/Microsoft/msphpsql/releases/tag/v4.2.0-preview), you can get the driver from our PECL repositiry or GitHub releases page. Note that sqlvariant for output parameters is not supported.

PR #399, opened 2017-05-17
sql variant support

Hadis-Fard:
Limitation

- sqlvariant is not supported for output parameters, and operand type clash error is expected.


test/pdo_sqlsrv/pdo_param_output_variants.phpt
```diff
@@ -0,0 +1,140 @@
...
+}
+
+RunTest();
+
+?>
+--EXPECT--
+﻿
+SQLSTATE[22018]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Operand type clash: nvarchar(max) is incompatible with sql_variant
+SQLSTATE[22018]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Operand type clash: nvarchar(max) is incompatible with sql_variant
+
```
yitam: Where did nvarchar(max) come from?? The error messages don't make sense here...?

test/sqlsrv/sqlsrv_simple_fetch_variants_.phpt
```diff
@@ -0,0 +1,194 @@
+--TEST--
+Test simple insert and fetch sql_variants as strings using inputs of various data types
```
yitam: This test looks so much similar to sqlsrv/sqlsrv_simple_fetch_variants.phpt? What's the difference?

test/pdo_sqlsrv/pdo_fetch_variants_diff_styles.phpt
```diff
@@ -0,0 +1,296 @@
...
+        // insert column data from a row of the original table                    
+        $stmtOriginal = $conn->prepare("SELECT * FROM $tableName WHERE c1_int = :row");       
+        
+        for ($i = 1; $i <= $numRows; $i++)
+        {
+            $c1_int = $i;
+            
+            echo "Insert all columns from row $c1_int into one column of type sql_variant\n";            
+            $stmtOriginal->bindValue(':row', $c1_int, PDO::PARAM_INT);
+
```
yitam: What if the user calls bindValue() without specifying the type? Will there be an error?

test/sqlsrv/sqlsrv_simple_fetch_variants_.phpt
Hadis-Fard: Oops, accidentally checked this in 😃

test/pdo_sqlsrv/pdo_param_output_variants.phpt
Hadis-Fard: the initialized data is php string that will fall into sql varchar* type group, in the [core stmt](https://github.com/Microsoft/msphpsql/blob/dev/source/shared/core_stmt.cpp#L546) when binding output parameters the size is assigned to max to avoid silent truncation. so when it is passed to ODBC the type is max, and the server receives the max char type. For error handling in this part we're benefiting from server side errors, as we cannot identify the type of the variant column beforehand (the same reason why we cannot support output parameters), and we can't constraint the error `SQLSTATE[22018]` to only variant, and it might occur when querying other columns than variant.

Hadis-Fard:
issue #51 and #127

PR #402, opened 2017-05-19
extracted select output param into a different test

yitam:
split the test pdo_param_output_variants.phpt into two separate tests to get different error messages


test/pdo_sqlsrv/pdo_param_output_select_variant.phpt
```diff
@@ -0,0 +1,85 @@
+--TEST--
+Test parametrized insert and sql_variant as an output parameter.
+--DESCRIPTION--
+parameterized queries is not supported for Sql_Variant columns, this test, verifies a proper error message is returned
```
Hadis-Fard: only output or inout parameters are not supported

PR #412, opened 2017-05-30
Fix failing tests

david-puglielli:


test/pdo_sqlsrv/PDO_ConnPool_Unix.phpt
```diff
@@ -1,35 +1,63 @@
 --TEST--
-PDO Connection Pooling Test on Unix
+SQLSRV Connection Pooling Test on Unix
```
v-kaywon: SQLSRV? Shouldn't it be PDO?

test/pdo_sqlsrv/PDO_ConnPool_Unix.phpt
```diff
@@ -4,32 +4,60 @@ PDO Connection Pooling Test on Unix
...
 <?php
+// On Bamboo we must use sudo to fiddle with odbcinst.ini
+// On travis-ci we can't use sudo
+$sudo = '';
+$user = posix_getpwuid(posix_geteuid());
+if (strtolower($user['name']) == 'bamboo')
+{
+    $sudo = 'sudo ';
+}
+
```
yitam: Can we generalize this by checking if it's `root` instead? So if `$user['name'] == 'root'` then do not use `sudo`
Also please modify the comment to reflect this change.

PR #419, opened 2017-06-02
added database setup files for ae testing and some priliminary tests

v-kaywon:


test/ae_database/AESetup.inc
```diff
@@ -0,0 +1,251 @@
...
+    $conn = sqlsrv_connect($server, $connectionInfo);
+    sqlsrv_assert($conn === false, "Could not connect.\n");
+    return ($conn);
+}
+
+function DropAETable($conn){
+    $sql_drop = "IF OBJECT_ID('dbo.test_AE', 'U') IS NOT NULL DROP TABLE [dbo].[test_AE]";
+    $stmt = sqlsrv_query($conn, $sql_drop);
+    sqlsrv_assert($stmt === false, "Error in dropping table.\n");
+}
```
yitam: Perhaps think of another function name rather than sqlsrv_assert()? This sounds like one of the existing sqlsrv api...

test/ae_database/AESetup.inc
```diff
@@ -0,0 +1,251 @@
...
+                                [encRandChar] [char] (10)
+                                $encRand";
+                break;
+            case "varchar":
+                $sql_create .= "[normVarchar][varchar] (50) NULL,
+                                [encDetVarchar] [varchar] (50) COLLATE Latin1_General_BIN2
+                                $encDet
+                                [encRandVarchar] [varchar] (50)
+                                $encRand";
+                break;
```
yitam: is there a way to specify the size of the char/varchar columns?

yitam: similarly with nchar, binary, etc.

test/ae_database/databasesetup.php
```diff
@@ -0,0 +1,254 @@
...
+                                                       [normDecimal] [decimal](18,5), [encDetDecimal] [decimal](18,5), [encRandDecimal] [decimal](18,5),
+                                                       [normNumeric] [numeric](18,5), [encDetNumeric] [numeric](18,5), [encRandNumeric] [numeric](18,5),
+                                                       [normMoney] [money], [encDetMoney] [money], [encRandMoney] [money],
+                                                       [normSmallmoney] [smallmoney], [encDetSmallmoney] [smallmoney], [encRandSmallmoney] [smallmoney],
+                                                       [normBit] [bit], [encDetBit] [bit], [encRandBit] [bit])";
+
+$stmt = sqlsrv_query($conn, $sql_create);
+if ($stmt === false) {
+    echo "Failed to create table.\n";
+    die(print_r(sqlsrv_errors(), true));
```
yitam: Since this setup script contains more than one table, consider making all the error messages more descriptive. For example, here it should be 'failed to create table test_AE_exnum' or something like that. When errors occurred we will know what exactly went wrong.

test/ae_database/AESetup.inc
```diff
@@ -0,0 +1,251 @@
...
+    
+    $stmt = sqlsrv_query($conn, $sql_create);
+    sqlsrv_assert($stmt === false, "Error in creating table.\n");
+}
+
+function CreateAETable($conn){
+    CreateAETableTypes($conn, "nvarchar", "int", "float", "datetime2", "bit");
+}
+
+function PopulateAETable($conn){
```
yitam: Why do we need CreateAETable() and PopulateAETable() here? Are they a mix of column types?

test/ae_database/encrypttable.ps1
```diff
@@ -0,0 +1,31 @@
+# Import the SqlServer module.
+Import-Module "SqlServer"
+
+# Connect to your database.
+$serverName = "yourServer"
+$databaseName = "AEDemo"
+$userName = "yourUsername";
+$password = "yourPassword";
+
+#For Windows Authentication
```
yitam: Do all these files reference the same credentials? If so, consider creating a file only with credentials so that the other scripts can simply reference this file?

test/ae_database/encrypttable.ps1
```diff
@@ -0,0 +1,31 @@
...
+#$database = $server.Databases[$databaseName]
+
+#For SQL Server Authentication
+Add-Type -AssemblyName "Microsoft.SqlServer.Smo"
+$MySQL = new-object('Microsoft.SqlServer.Management.Smo.Server') $serverName 
+$MySQL.ConnectionContext.LoginSecure = $false
+$MySQL.ConnectionContext.set_Login($userName)
+$MySQL.ConnectionContext.set_Password($password)
+$database = $MySQL.Databases[$databaseName]
+
```
yitam: Perhaps use a different variable name than 'MySQL'? ;)

test/ae_database/databasesetup.php
```diff
@@ -0,0 +1,254 @@
...
+sqlsrv_configure( 'WarningsReturnAsErrors', 1 );
+sqlsrv_configure( 'LogSeverity', SQLSRV_LOG_SEVERITY_ALL );  
+
+$databaseName = 'AEDemo';
+$uid = 'yourUsername';
+$pwd = 'yourPassword';
+$server = 'yourServer';
+
+$connectionInfo = array( "Database"=>$databaseName, "UID"=>$uid, "PWD"=>$pwd);
+
```
yitam: How and where do you create this particular database?

test/ae_database/AESetup.inc
v-kaywon: I was thinking of the same thing too. There is a way to do it, but it won't work for the data in AEData.inc anymore

yitam: you can define a constant or variable in the beginning and use that instead of hard-coding?
it's more flexible even if it may not work with AEData.inc because you can always add another data file
or you can suggest what's the minimum values for these varchar columns in order to work with AEData.inc

test/ae_database/databasesetup.php
v-kaywon: I'm not going to change this for now since I don't really know how the directory structure is going to look like at the end. Maybe these tests will be put into the functional test folder and they can all just use MSSetup.inc.

Hadis-Fard:
@v-kaywon @yitam Could you review this PR too?

v-kaywon:
this PR include tests for testing AE insertion, let's leave it for now until AE insertion has been implemented.

PR #421, opened 2017-06-02
Performance tests

ulvii:
Adding Performance tests and environment setup scripts.


test/Performance/README.md
```diff
@@ -0,0 +1,30 @@
...
+    sudo env “PATH=$PATH” bash setup_env_unix.sh RedHat7 <PHP_VERSION - 7.x.x> <PHP_THREAD - ts or nts> <Absolute path to driver source folder>
+### Sierra
+`brew` cannot be run with `sudo` on Sierra. Either enable passwordless `sudo` on the machine or enter the password when prompted. 
+
+    bash setup_env_unix.sh Sierra <PHP_VERSION - 7.x.x> <PHP_THREAD - ts or nts> <Absolute path to driver source folder>
+## Run benchmarks - Subject to change once the process is automated
+
+Run sqlsrv benchmarks:
+
+    ./vendor/bin/phpbench run benchmark/sqlsrv/regular --time-unit="milliseconds" --iterations [num_of_iterations] --report=aggregate
```
ulvii: Will be done in run-tests script

Hadis-Fard: if it can be done here using the framework, why should it rely on the script?

ulvii: I mean this command will be ran in the run-test script, I will make sure the `time-unit` is set to seconds.

test/Performance/benchmark/sqlsrv/regular/SqlsrvCreateDbTableProcBench.php
```diff
@@ -0,0 +1,31 @@
...
+    /*
+    * Each iteration creates a database, a table and a stored procedure in that database and drops the database at the end.
+    * Not that, execDirect function are used to execute all the queries. 
+    */
+    public function benchCreateDbTableProc(){
+        $randomNum = rand();
+        $databaseName = "test_db_$randomNum";
+        $tableName = "test_table_$randomNum";
+        $procName = "test_proc_$randomNum";
+        SqlsrvUtil::createDbTableProc( $this->conn, $databaseName, $tableName, $procName );
```
ulvii: SqlsrvUtil is a helper class with static functions. You'd have to create an instance of the class if the function is not static.

Hadis-Fard: I mean why the function should be in the class at all, if it's only used here, and not gonna be used anywhere else?

ulvii: Mainly to keep the benchmark clean.

test/Performance/setup_env_windows.ps1
```diff
@@ -0,0 +1,112 @@
...
+
+Copy-Item php.ini-production php.ini
+Add-Content php.ini "extension=C:\php\ext\php_sqlsrv.dll"
+Add-Content php.ini "extension=C:\php\ext\php_pdo_sqlsrv.dll"
+Add-Content php.ini "extension=C:\php\ext\php_openssl.dll"
+Move-Item php.ini C:\Windows -force
+Copy-Item C:\php-sdk\phpdev\vc14\$ARCH\deps\bin\ssleay32.dll C:\Windows -force
+Copy-Item C:\php-sdk\phpdev\vc14\$ARCH\deps\bin\libeay32.dll C:\Windows -force
+
+cd $startingDir
```
v-kaywon: Can you put this line just before "Remove-Item temp -Recurse -Force -ErrorAction Ignore" (line 111)? Putting this line here looks like the whole block of code (including line 107-110) requires cd $startingDir.

test/Performance/setup_env_windows.ps1
```diff
@@ -0,0 +1,112 @@
...
+$env:Path += ";C:\php\"
```
v-kaywon: Do we want to change the environment path only in the current session? or permanently?

test/Performance/run-perf_tests.py
```diff
@@ -0,0 +1,792 @@
...
+            self.database_name = database_name
+            self.username = username
+            self.password = password
+ 
+class XMLResult( object ):
+    """
+    A class to keep a result set of a benchmark generated by PHPBench as an XML file.
+    Attributes:
+        benchmark_name (str): The name or the benchmark.
+        success (int): 0 or 1. 0 if the benchmark to execute, 1 if the execution was successful.
```
v-kaywon: 0 if the benchmark to execute **fails**?

test/Performance/run-perf_tests.py
```diff
@@ -0,0 +1,792 @@
...
+        , database=db.database_name
+        , user=db.username
+        , password=db.password
+        , autocommit = True)
+ 
+def get_server_version( server ):
+    """
+    This module returns the version of the given server
+    Args:
+        server (obj): Server object to connect to
```
v-kaywon: use "database object" instead? to be consistent with connect( db )

test/Performance/setup_env_windows.ps1
ulvii: Should be permanent.

test/Performance/setup_env_windows.ps1
ulvii: No, we actually want Composer to be installed in $startingDir. I didnt install composer system-wide. 
You'd use composer.phar in the current directory.

test/Performance/run-perf_tests.py
ulvii: yes :)

test/Performance/run-perf_tests.py
ulvii: I am using this class as server when database_name is not required. Example, when fetching @@Version. We could probably name this class better though.

test/Performance/run-perf_tests.py
```diff
@@ -0,0 +1,792 @@
...
+def parse_results( dump_file ):
+    """
+    This module parses the .xml files generated by PHPBench
+    @TO-DO: Currently only limited detailes are parsed, such as duration and peak memory.
+    PHPBench reports a lot more information that can be helpful.
+    Consider looking at the xml files.
+    Args:
+        dump_file (str): The name of the XML file to be parsed.
+    Returns:
+        N/A
```
v-kaywon: returns array of parsed results, each entry has information like duration and memory

ulvii: Right :)

test/Performance/run-perf_tests.py
```diff
@@ -0,0 +1,792 @@
...
+    This module inserts a new entry into Drivers table
+    Args:
+        conn (obj): Connection object to the Results database
+        name (str): Full path to the driver
+        driver_hash (bin): sha1sum hash of the driver 
+    Returns:
+        N/A    
+    """
+    file_date = time.strftime( fmt, time.gmtime( os.path.getmtime( driver_path )))
+    query = "INSERT INTO Drivers ( Arch, FileDate, SHA1, HashVer ) VALUES ( ?, ?, {0}, 1 )"
```
v-kaywon: What is HashVer? Is it because sha**1** was used? If that is the case, I think it's better to make the HashVer a global variable so we can change it whenever a different sha version is used.

Also, is there a reason why you used bind param here? While for most of the other cases you just use query.format(...).

test/Performance/run-perf_tests.py
```diff
@@ -0,0 +1,792 @@
...
+    server_id = get_server_id( conn, test_db )
+    client_id = get_client_id( conn )
+    team_id   = get_team_id( conn )
+    driver_id = get_driver_id( conn, driver )
+ 
+   
+    php_arch    = get_php_arch()
+    php_thread  = get_php_thread()
+    php_version = get_php_version()
+    driver_version = get_driver_version( driver )
```
v-kaywon: This variable is not used in this function.

test/Performance/run-perf_tests.py
```diff
@@ -0,0 +1,792 @@
...
+    query = "INSERT INTO PerformanceResults( TestId, ClientId, DriverId, ServerId, TeamId, Success ) OUTPUT INSERTED.ResultId VALUES( {0}, {1}, {2}, {3}, {4}, {5} )"
+    cursor = conn.cursor()
+    cursor.execute( query.format( test_id, client_id, driver_id, server_id, team_id, success ))
+    result_id = cursor.fetchone()
+    cursor.close()
+    if result_id is not None:
+        return result_id[0]
+    return id
+ 
+def insert_key_value( conn, table_name, result_id, key, value ):
```
v-kaywon: suggest changing the name of this function to something like insert_result_key_value(). insert_key_value() sounds very generic (i.e., sounds like it can be used to insert something into the server table, or the client table).

test/Performance/lib/PDOSqlsrvUtil.php
```diff
@@ -0,0 +1,228 @@
...
+    private function fetch( $stmt ){
+        return $stmt->fetch();
+    }
+
+    private function prepare( $conn, $sql ){
+        try{
+            $stmt = $conn->prepare( $sql );
+            if( $stmt === false ){
+                die( "Failed to prepare\n");
+            }
```
v-kaywon: this `($stmt === false)` block is not needed. At connection, you did `$conn->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);`, so if prepare fails, it returns a PDOException object

test/Performance/benchmark/pdo_sqlsrv/regular/PDOCreateDbTableProcBench.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+class PDOCreateDbTableProcBench{
+    private $conn;
+
+    public function connect(){
+        $this->conn = PDOSqlsrvUtil::connect();
+    }
+    /*
+    * Each iteration creates a database, a table and a stored procedure in that database and drops the database at the end.
+    * Not that, execDirect function are used to execute all the queries.
```
v-kaywon: *Note

test/Performance/lib/PDOSqlsrvUtil.php
ulvii: These functions are generic. I check possible failure options and handle them. There can be another connection that does not have `ERRMODE_EXCEPTION` set.

test/Performance/report.sql
```diff
@@ -0,0 +1,24 @@
+--The script can be run to read the results
```
v-kaywon: add ". You can filter out the results ran on a certain date by changing the date at the end."

test/Performance/benchmark/sqlsrv/regular/SqlsrvUpdateBench.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function generateUpdateParams(){
+        $this->updateParams = SqlsrvUtil::generateUpdateParams();
+    }
+    /**
+     * Each iteration inserts a row into the table, updateWithPrepare() updates that row 1000 times.
+     * Note that, every update calls prepare, bindParam and execute APIs.
+     */
+    public function benchUpdateWithPrepare(){
+        for( $i=0; $i<1000; $i++ ){
```
v-kaywon: Seems like this 1000 is used at a lot of places. Do you think it's a better idea if you store this number as a variable in SqlsrvUtil and PDOSqlsrvUtil?

test/Performance/benchmark/sqlsrv/regular/SqlsrvCreateDbTableProcBench.php
v-kaywon: This function may also be used if more tests are added.

test/Performance/setup_env_windows.ps1
v-kaywon: by doing this only makes it temporary (I tried 'echo %PATH%' using another cmd, and C:\php\ is not shown there). This post (https://stackoverflow.com/questions/714877/setting-windows-powershell-path-variable) has a solution to make it permanent.

test/Performance/benchmark/pdo_sqlsrv/regular/PDOCreateDbTableProcBench.php
ulvii: yeap

test/Performance/benchmark/sqlsrv/regular/SqlsrvUpdateBench.php
ulvii: 1000 is the fixed number of rows in the requirements for CRUD tests.

test/Performance/setup_env_windows.ps1
ulvii: I know, I have fixed it already, will be pushed with the next commit :)

test/Performance/run-perf_tests.py
ulvii: Sure

test/Performance/run-perf_tests.py
ulvii: good catch. It should be inserted as a key value too.

test/Performance/run-perf_tests.py
ulvii: I highly doubt we will want to change the sha version. We are inserting HashVer, because the result database we are using is meant be used for any test results, it can also contain different versions of has algorithms. To determine which binary is used when running the tests, sha1sum is good enough :)

Well, I tried using bind param for everything, but couldnt get pyodbc binary binding working.

test/Performance/report.sql
ulvii: Done

test/Performance/setup_env_unix.sh
```diff
@@ -0,0 +1,142 @@
+#!/bin/bash
+set -e
+if [[ ("$1" = "Ubuntu16" || "$1" = "RedHat7" || "$1" = "Sierra") ]]; then
+    PLATFORM=$1
+else
+    echo "First argument must be one of Ubuntu16, RedHat7, Sierra. Exiting..."
+    exit 1
+fi
+if [[ "$2" = "" ]]; then
```
v-kaywon: check for ![ "$2" = "7*" ] instead?

ulvii: Changed to `if [[ "$2" != 7.*.* ]]; then`
For Unix, php download page is the same for both latest and older versions of PHP.
For Windows, the latest source is in http://windows.php.net/downloads/releases/,
the older versions are in http://windows.php.net/downloads/releases/archives/

test/Performance/setup_env_unix.sh
```diff
@@ -0,0 +1,142 @@
...
+elif [ $PLATFORM = "RedHat7" ]; then
+    echo "Update..."
+    yes | sudo yum update >> env_setup.log 2>&1
+    echo "Enabling EPEL repo..."
+    yes | sudo yum install epel-release -y >> env_setup.log 2>&1
+    echo "OK"
+    echo "Installing python34-setuptools..."
+    yes | sudo yum install python34-setuptools -y >> env_setup.log 2>&1
+    echo "OK"
+    echo "Installing gcc, git, zip libxml, openssl, EPEL, python3, pip3..."
```
v-kaywon: remove EPEL and pip3 as they are install at earlier and later steps respectively

test/Performance/setup_env_unix.sh
```diff
@@ -0,0 +1,142 @@
...
+php --ri pdo_sqlsrv
+echo "Installing Composer..."
+wget https://getcomposer.org/installer -O composer-setup.php >> env_setup.log 2>&1
+php composer-setup.php >> env_setup.log 2>&1
+echo "OK"
+echo "Installing PHPBench..."
+php composer.phar install >> env_setup.log 2>&1
+echo "OK"
+echo "Cleaning up..."
+rm -rf php-$PHP_VERSION* compser-setup.php
```
v-kaywon: why remove php-$PHP_VERSION?

test/Performance/setup_env_unix.sh
v-kaywon: I mean why doesn't Windows have a check to exit if the provided PHP version is not 7.x.x

test/Performance/setup_env_unix.sh
ulvii: You dont need the source folder once the binaries are installed.

test/Performance/setup_env_unix.sh
ulvii: I havent had a chance to properly test the script on RedHat yet, will address this

test/Performance/setup_env_unix.sh
ulvii: I added the check to the windows script too.

test/Performance/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+    bash setup_env_unix.sh Sierra <PHP_VERSION - 7.x.x> <PHP_THREAD - ts or nts> <Absolute path to driver source folder>
+## Run benchmarks
+PHPBench is used to run the benchmarks. Visit http://phpbench.readthedocs.io/en/latest/introduction.html to have an idea how the tool works.
+
+### 1. Modify lib/connect.php with the test database credetials
+### 2. Execute run-perf_tests.py. 
+On Linux and Mac, the script must be executed with `sudo python3` because to enable pooling it needs to modify odbcinst.ini system file. As an improvement, the location of the odbcinst.ini file can be changed so that, sudo is not requiered. 
+    
+    py.exe run-perf_tests.py -platform <PLATFORM> -iterations <ITERATIONS> -iterations-large <ITERATIONS_LARGE> -result-server <RESULT_SERVER> -result-db <RESULT_DB> -result-uid <RESULT_UID> -result-pwd <RESULT_PWD>
```
v-kaywon: put these into subsections? 
Linux and Mac:
  sudo python3 run-perf_tests.py ...
Windows:
  py.exe run-perf_tests.py ...

PR #424, opened 2017-06-07
Use the latest msodbcsql.h from ODBC 13.1

yitam:


yitam:
Decided not to adopt msodbcsql.h as is.

PR #426, opened 2017-06-07
Add TNIR conn option and tests

v-kaywon:


Hadis-Fard:
side note, check the coverall results on new code make sure it is covered more than 90%

v-kaywon:
If using ODBC < 13.1, sqlsrv_error (or PDOException object) returns ODBC error about invalid attribute. If value is not enabled or disabled, sqlsrv_error (or PDOException object) returns ODBC error about invalid value for the attribute. Connection will not be established in both cases. Since error is returned from the ODBC driver, do we still need to incorporate error handling on our side?

Hadis-Fard:
it is better to add the validation in PHP side before making the call to ODBC which results in a better perf

PR #429, opened 2017-06-08
Update msodbc constants

yitam:
Updated msodbc constants based on the latest header file from ODBC driver, in particular the ones for Column Encryption

### DIFF
--- a/source/shared/msodbcsql.h
+++ b/source/shared/msodbcsql.h
@@ -36,7 +36,7 @@
 
 
 #if !defined(SQLODBC_VER)
-#define SQLODBC_VER 1100
+#define SQLODBC_VER 1300
 #endif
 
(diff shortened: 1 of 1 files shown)


source/shared/msodbcsql.h
```diff
@@ -365,57 +361,68 @@
...
-} AEKEYSTOREPROVIDER;
-
-/* Data is defined to be past the end of the structure header.
- This is accepted by MSVC, GCC, and C99 standard but former emits
- unnecessary warning, hence it has to be disabled.
-*/
+        unsigned short ecekLen,
+        unsigned char **cekOut,
+        unsigned short *cekLen);
+    void (__stdcall *Free)();
```
lilgreenbird: what about EncryptCEK?

PR #435, opened 2017-06-12
Azure AD tests using ActiveDirectoryPassword

yitam:
Testing Azure AD on an Azure database instance


test/pdo_sqlsrv/pdo_azure_ad_authentication.phpt
```diff
@@ -50,8 +55,37 @@ catch( PDOException $e )
...
+// your credentials to test, or this part is skipped.
+//
+$azureServer = $adServer;
+$azureDatabase = $adDatabase;
+$azureUsername = $adUser;
+$azurePassword = $adPassword;
+
+if ($azureServer != 'TARGET_AD_SERVER')
+{
+    $connectionInfo = "Authentication = ActiveDirectoryPassword; TrustServerCertificate = false";
```
lilgreenbird: TrustServerCertificate option is yes/no in ODBC

yitam: in our drivers we take care of that... true means yes and false means no...

Issue #417, opened 2017-06-01
Extension Load Order issue in MacOS


lilgreenbird:
The ordering of the extension may cause a problem in MacOS.
This can be reproduced by specifying the extension order in php.ini as follows:

```
   extension=php_sqlsrv_71_nts.so
   extension=php_pdo_sqlsrv_71_nts.so
```

With this order  logging severity can not be set. ie the following has no effect:

```
   sqlsrv_configure(` 'LogSeverity', SQLSRV_LOG_SEVERITY_ALL );
```

but` if the order is reversed (ie pdo 1st then sqlsrv 2nd) then this is ok.

This is reproducible both using an INI file and dynamic loading and is seen in El Capitan and Sierra only. The ordering is not an issue in CentOS 7 or Ubuntu 15 or 16.

This is related to [PHP bug #74293](https://bugs.php.net/bug.php?id=74293)

meet-bhagdev:
@lilgreenbird Good catch. This is definitely an issue and we are taking a look at it.

ulvii:
Fixed #438

PR #438, opened 2017-06-16
Fixed load ordering issue in MacOS caused by flat namespace

lilgreenbird:
added MACOSX_DEPLOYMENT_TARGET to not use linker flat_namespace flag


Hadis-Fard:
#417

PR #440, opened 2017-06-16
fixed regex for expected error messages

yitam:
for calling sqlsrv_fetch right after sqlsrv_cancel

### DIFF
--- a/test/bvt/sqlsrv/msdn_sqlsrv_cancel.phpt
+++ b/test/bvt/sqlsrv/msdn_sqlsrv_cancel.phpt
@@ -46,4 +46,4 @@ echo "$count sales accounted for the first $$salesTotal in revenue.\n";
 sqlsrv_cancel( $stmt);
 ?>
 --EXPECT--
-58 sales accounted for the first $104205.9607 in revenue.
+57 sales accounted for the first $104171.7607 in revenue.
--- a/test/sqlsrv/sqlsrv_statement_cancel.phpt
+++ b/test/sqlsrv/sqlsrv_statement_cancel.phpt
@@ -94,7 +94,7 @@ Repro();
 --EXPECTREGEX--
 ﻿
 Test begins...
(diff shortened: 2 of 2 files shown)


test/sqlsrv/sqlsrv_statement_cancel.phpt
```diff
@@ -94,7 +94,7 @@ Repro();
 --EXPECTREGEX--
 ﻿
 Test begins...
-\[Microsoft\](\[ODBC Driver 13 for SQL Server\]|\[ODBC Driver Manager\])([ ]{0,1}Function sequence error)
+(.*\[Driver Manager\]|\[Microsoft\](\[ODBC Driver 13 for SQL Server\]|\[ODBC Driver Manager\]))([ ]{0,1}Function sequence error)
```
ulvii: Why do we need to do this?

test/bvt/sqlsrv/msdn_sqlsrv_cancel.phpt
```diff
@@ -46,4 +46,4 @@ echo "$count sales accounted for the first $$salesTotal in revenue.\n";
 sqlsrv_cancel( $stmt);
 ?>
 --EXPECT--
-58 sales accounted for the first $104205.9607 in revenue.
\ No newline at end of file
+57 sales accounted for the first $104171.7607 in revenue.
```
ulvii: Why?

yitam: It's the problem with the sql query itself. See #436 for details. Back then I put the incorrect expected output.

test/sqlsrv/sqlsrv_statement_cancel.phpt
yitam: in Mac the error message is different from the others even though the real message `Function sequence error` is the same in all platforms

ulvii: UnixODBC 2.3.4 adds extraneous DM prefix to diagnostic messages.

PR #443, opened 2017-06-16
Fix leak

v-kaywon:
prevent leak in zval containing the connection and statement resource


test/pdo_sqlsrv/PDO81_MemoryCheck.phpt
```diff
@@ -23,6 +23,7 @@ function MemCheck($noPasses, $noRows1, $noRows2, $startStep, $endStep, $leakThre
 
     Trace("Execution setup: $noPasses passes over a table with $noRows1 => ".($noRows1 + $noRows2)." rows.\n");
     $conn1 = Connect();
+    $conn1->setAttribute(PDO::SQLSRV_ATTR_ENCODING, PDO::SQLSRV_ENCODING_UTF8);
     SetUTF8Data(true);
```
yitam: I think in pdo tests we almost always use utf8 data...

test/pdo_sqlsrv/MsCommon.inc
```diff
@@ -1221,7 +1221,15 @@ function NumRows($conn, $tableName)
...
+    if (UseUTF8data())
+    {
+        include_once 'MsData_UTF8.inc';
+        return (InsertQueryExUTF8($tableName, rand(1, 20)));
+    }
+    else 
+    {
+        return (InsertQueryEx($tableName, rand(1, 20)));
+    }
 }
```
yitam: do we really need ANSI version of data?

PR #454, opened 2017-06-23
rearrange test folders as functional

yitam:


Hadis-Fard:
The setup files in setup folders are related to functional tests, so they should be moved into the related folder(functional tests). so in future we're gonna have, `bvt`, `functional`, `performance`, `stress` and etc. in the same level of directory hierarchy , and each of these test categories, would have their own setup files.

PR #458, opened 2017-06-27
Azure tests

yitam:


test/functional/pdo_sqlsrv/isPooled.php
```diff
@@ -1,10 +1,10 @@
 <?php
 include 'MsSetup.inc';
-$conn1 = new PDO("sqlsrv:Server=$server", $uid, $pwd);
+$conn1 = new PDO("sqlsrv:Server=$server; database=$databaseName", $uid, $pwd);
 $connId1 = ConnectionID($conn1);
 $conn1 = null;
 
-$conn2 = new PDO("sqlsrv:Server=$server", $uid, $pwd);
+$conn2 = new PDO("sqlsrv:Server=$server; database=$databaseName", $uid, $pwd);
```
v-kaywon: Do we need to connect to a database here?

test/functional/pdo_sqlsrv/pdo_passwords.phpt
```diff
@@ -3,8 +3,7 @@ Test password with non alphanumeric characters
 --DESCRIPTION--
 The first three cases should have no problem connecting. Only the last case fails because the 
 right curly brace should be escaped with another right brace.
-In Azure we can't set DEFAULT_DATABASE for a login user. For this test to psss must connect
-to the test database defined in MsSetup.inc 
+In Azure for this test to psss do not specify any particular database when connecting
```
v-kaywon: *pass

v-kaywon: I there a reason why it won't pass if the database is specified?

test/functional/pdo_sqlsrv/isPooled.php
yitam: yes it's required for it to pass in Azure

test/functional/sqlsrv/test_non_alpha_password.phpt
```diff
@@ -3,42 +3,52 @@ password with non alphanumeric characters
 --DESCRIPTION--
 The first three cases should have no problem connecting. Only the last case fails because the 
 right curly brace should be escaped with another right brace.
-In Azure we can't set DEFAULT_DATABASE for a login user. For this test to psss must connect
-to the test database defined in MsSetup.inc
+In Azure for this test to psss do not specify any particular database when connecting
```
v-kaywon: *pass

test/functional/pdo_sqlsrv/pdo_passwords.phpt
yitam: It used to work the other way but for some reason we had to revert. We can create a backlog issue for this.

test/functional/sqlsrv/test_connectionOption.phpt
```diff
@@ -11,7 +11,7 @@ Test MultipleActiveResultSets connection setting off
 
     require( 'MsCommon.inc' );
 
-    $conn = ConnectSpecial(array( 'MultipleActiveResultSets' => false ));
+    $conn = Connect(array( 'MultipleActiveResultSets' => false ));
```
david-puglielli: What is the difference between Connect and ConnectSpecial?

test/functional/sqlsrv/test_non_alpha_password.phpt
```diff
@@ -51,5 +62,5 @@ print "Test successful";
 ?> 
 --EXPECTREGEX--
 An unescaped right brace \(}\) was found in either the user name or password.  All right braces must be escaped with another right brace \(}}\)\.
-Warning: sqlsrv_close\(\) expects parameter 1 to be resource, boolean given in .+(\/|\\)test_non_alpha_password\.php on line 38
+Warning: sqlsrv_close\(\) expects parameter 1 to be resource, boolean given in .+(\/|\\)test_non_alpha_password\.php on line 50
```
david-puglielli: Do we want to specify a line number? Maybe make the number a regex? Or maybe put in a conditional like `if ($conn) sqlsrv_close($conn);` so that we don't output this warning at all  when the test is working properly, otherwise seeing this message in a diff or log file makes it look at first glance like there's a problem.

test/functional/sqlsrv/test_connectionOption.phpt
yitam: Connect() basically merges the given array to the existing options and ConnectSpecial() makes sure uid, pwd and database are set using the defaults if not already defined

test/functional/sqlsrv/test_non_alpha_password.phpt
yitam: I think for this test the line number is important because all the test cases before should pass

test/functional/sqlsrv/test_connectionOption.phpt
yitam: Actually ConnectSpecial is not used in any test now so I'm going to remove it

test/functional/sqlsrv/test_non_alpha_password.phpt
david-puglielli: Even so I think it would be better to put in a message in the main body of the test - `if ($conn) {sqlsrv_close($conn); echo "Error, $conn should not be valid\n";}` instead of using PHP's standard error message here.

PR #459, opened 2017-06-29
fix for fortify warnings

v-kaywon:


source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -1343,7 +1343,7 @@ int pdo_sqlsrv_dbh_quote( pdo_dbh_t* dbh, const char* unquoted, size_t unquoted_
         }
         // get the placeholder at the current position in driver_stmt->placeholders ht
         zval* placeholder = NULL;
-        if (( placeholder = zend_hash_get_current_data( driver_stmt->placeholders )) != NULL && zend_hash_move_forward( driver_stmt->placeholders ) == SUCCESS ) {
+        if (( placeholder = zend_hash_get_current_data( driver_stmt->placeholders )) != NULL && zend_hash_move_forward( driver_stmt->placeholders ) == SUCCESS && stmt->bound_params != NULL ) {
```
yitam: I'd suggest you check if stmt->bound_params is NULL first before calling the zend hash methods

v-kaywon: decided not to change the ordering because "placeholder = zend_hash_get_current_data( driver_stmt->placeholders )) != NULL" is more likely to fail and if it fails the if statement can just move on without doing other checks

PR #460, opened 2017-06-29
Fixing issues reported by binary analysis - Hardening checks

ulvii:
- Adding compiler options `-D_FORTIFY_SOURCE` and `-fstack-protector`
- Adding linker option  `-Wl,-z,now` for Linux and `-Wl,-bind_at_load` for Max
- Fixing the issue with detecting the platform


ulvii:
Failed because pdo_sqlsrv.so was loaded before pdo.so after the changes

ulvii:
Adding optimizations flag decreases the coverage, because some lines are not hit after the optimization. Turning on optimization flags makes the compiler attempt to improve the performance and/or code size at the expense of compilation time and possibly the ability to debug the program. This flag should be turned off when debugging.

source/pdo_sqlsrv/config.m4
```diff
@@ -52,8 +52,14 @@ if test "$PHP_PDO_SQLSRV" != "no"; then
   fi
     AC_MSG_RESULT($pdo_sqlsrv_inc_path)
     
-           
+  HOST_OS_ARCH=`uname`         
+  if test "${HOST_OS_ARCH}" = "Darwin"; then
+      MACOSX_DEPLOYMENT_TARGET=`sw_vers -productVersion` 
+  fi
```
v-kaywon: why does this guy need to be moved up?

source/sqlsrv/config.w32
```diff
@@ -3,7 +3,7 @@
 //
 // Contents: JScript build configuration used by buildconf.bat
 //
-// Microsoft Drivers 4.3 for PHP for SQL Server
+// Microsoft Drivers 4.2 for PHP for SQL Server
```
v-kaywon: shouldn't it be 4.3??

source/pdo_sqlsrv/config.m4
ulvii: It should be inside the if block.
Also, the previous way of checking the OS was wrong.

source/sqlsrv/config.w32
ulvii: it should :)

PR #463, opened 2017-07-04
adding test users for testing in Azure 

yitam:


Hadis-Fard:
@yitam can this be merged?

test/functional/setup/create_logins_azure.sql
```diff
@@ -0,0 +1,18 @@
+--for this script to work in Azure, use sqlcmd to connect to master database 
+IF NOT EXISTS (SELECT name FROM sys.sql_logins WHERE name = 'test_password')
+BEGIN
+    CREATE LOGIN test_password WITH PASSWORD='! ;4triou';
```
ulvii: Issue with this line: `test_password` might exist in the database with a different password. This will cause the tests fail. Same applies for `test_password2` and `test_password2`

Hadis-Fard: @yitam can we drop the test_password, and create it again? and at the end of the test, delete the pass too

yitam: that was the original problem of running tests in parallel (dropping one another's test logins and passwords)

Hadis-Fard: @yitam can you generate password using random? and drop it at the end?

yitam: unfortunately no, the specific test expects this particular password to make sure it fails and check for the appropriate error message

Hadis-Fard: what are those tests dependent on this pass? You can generate the pass in a parent setup script and then pass it on to dependent scripts

ulvii: The users and logins can be created inside the tests too, it does not have to be the setup scripts.

PR #481, opened 2017-07-22
test fixes for PHP 7.2

lilgreenbird:


test/functional/sqlsrv/sqlsrv_fetch_field_twice_data_types.phpt
```diff
@@ -72,7 +72,7 @@ function FetchField($stmt, $idx, $numFields, $errorExpected)
 function PrintError($errorExpected = true)
 {
     $errors = sqlsrv_errors(SQLSRV_ERR_ALL);
-    if (count($errors) > 0)
+    if (!empty($errors) && count($errors) > 0)
     {
```
yitam: already checking if it's not empty why count() too?

yitam: same for the other two tests

lilgreenbird: good point, will remove

PR #482, opened 2017-07-25
Aesetup

v-kaywon:
Modify setup_dbs.py for setting up self signed certificate, column master key and column encryption key. Add in code for handling ColumnEncryption connection string keyword. Add test for testing certificate setup, and connect string keyword.


test/functional/pdo_sqlsrv/MsCommon.inc
```diff
@@ -14,6 +14,16 @@
...
+{
+    $msodbcsql_ver = $conn->getAttribute(PDO::ATTR_CLIENT_VERSION)["DriverVer"];
+    $server_ver = $conn->getAttribute(PDO::ATTR_SERVER_VERSION);
+    $msodbcsql_maj = explode(".", $msodbcsql_ver)[0];
+    if ($msodbcsql_maj < 13 || explode('.', $server_ver)[0] < 13)
+        return false;
+    return true;
+}
+
 function connect($options=array())
```
yitam: Please double check... currently, does the client version return 13.1 or 13.0? What if it's 13.0?

test/functional/pdo_sqlsrv/test_ae_keys_setup.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+retrieval of names of column master key and column encryption key generated in the database setup
+--SKIPIF--
+<?php if (strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN') die("Skipped: Test for Windows"); ?>
+--FILE--
```
yitam: you can use `skipif_unix` instead

test/functional/pdo_sqlsrv/test_ae_keys_setup.phpt
```diff
@@ -0,0 +1,36 @@
...
+    }
+    else {
+        die("Column Master Key and Column Encryption Key not created.\n");
+    }
+    unset($stmt);
+}
+else {
+    echo "Test Successfully.\n";
+}
+unset($conn);
```
yitam: did you mean `Test Successfully done.`?

test/functional/setup/setup_dbs.py
```diff
@@ -75,5 +83,8 @@ def executeBulkCopy(conn_options, dbname, tblname, datafile):
     setupTestDatabase(conn_options, args.DBNAME, args.AZURE)    
     # populate these tables
     populateTables(conn_options, args.DBNAME)
+    # setup AE (certificate, column master key and column encryption key)
+    setupAE(conn_options, args.DBNAME)
+
```
yitam: I believe we should check if setting up for Azure here? If so, should we skip this step even running in Windows?

PR #483, opened 2017-07-28
added the docker logs 

Hadis-Fard:
added the docker logs to sql and client containers
changed line ending lf


Hadis-Fard:
Closing this as it requires deeper investigation

PR #488, opened 2017-08-02
Fix Travis

Hadis-Fard:
- Run server image before building client

- Added sleep to before_script in travis to let the server fire up

- Added entrypoint script to test connection at client container run time


.travis.yml
```diff
@@ -15,22 +15,27 @@ env:
...
 script: 
-  - docker run -e TRAVIS_JOB_ID -t -d -w $PHPSQLDIR --link $TEST_PHP_SQL_SERVER --name=client msphpsql-dev
-  - docker ps -a 
+  - docker run -e TRAVIS_JOB_ID -t -d -w $PHPSQLDIR --name=client --link $TEST_PHP_SQL_SERVER msphpsql-dev
+  - docker ps -a
+  - docker logs client
   - docker exec client python ./test/functional/setup/setup_dbs.py -dbname $SQLSRV_DBNAME
   - docker exec client python ./test/functional/setup/setup_dbs.py -dbname $PDOSQLSRV_DBNAME
-  - docker exec client php ./source/pdo_sqlsrv/run-tests.php ./test/functional/pdo_sqlsrv/*.phpt
+  - docker exec client php ./source/sqlsrv/run-tests.php ./test/functional/pdo_sqlsrv/*.phpt
```
david-puglielli: This uses the same run-tests.php for sqlsrv and pdo_sqlsrv tests?

entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+set -e
+
+testConnection="/opt/mssql-tools/bin/sqlcmd -S sql -U sa -P Password123"
+
+until $testConnection; do
+>&2 echo "SQL Server is starting up.."
+sleep 1
+done
```
david-puglielli: Is there a chance this could result in an infinite loop if there is some other remote server problem?

.travis.yml
```diff
@@ -15,22 +15,27 @@ env:
   - PHPSQLDIR=/REPO/msphpsql-dev
   - TEST_PHP_SQL_SERVER=sql
   - SQLSRV_DBNAME=msphpsql_sqlsrv
-  - PDOSQLSRV_DBNAME=msphpsql_pdosqlsrv  
+  - PDOSQLSRV_DBNAME=msphpsql_pdosqlsrv
```
lilgreenbird: what changed here??

entrypoint.sh
lilgreenbird: this seems to be a very tight loop and as David suggested may result in an infinite loop if the server startup failed for whatever reason. I would set a limit and increase the wait.
Actually there's a travis_retry can we make use of that?

.travis.yml
Hadis-Fard: space at the end

.travis.yml
Hadis-Fard: good catch

entrypoint.sh
Hadis-Fard: updated please review! travis_retry is used when you install dependencies within travis and it gives timeout. in this case we install dependencies within docker

Hadis-Fard:
@david-puglielli @lilgreenbird Could you review please? would like to push this before other PRs?

PR #497, opened 2017-08-08
Implementation of fetching AE max columns

v-kaywon:


v-kaywon:
Performance measurement for fetching varchar(max) columns with and without AE:

Performance test information:
Driver: PDO_SQLSRV
Table columns: char(5), char(5, deterministic encrypted), char(5, randomize encrypted), varchar(max), varchar(max, deterministic encrypted), varchar(max, randomize encrypted)
Number of rows in table: 2
Number of iterations: 1000
Number of measurements: 10 (each measurement is the sum of all 1000 iterations)

Mean(with AE): 98.71267
Mean(without AE): 85.88191

Hadis-Fard:
@meet-bhagdev we have the performance measurement on AE retrieval, can you review?

source/shared/core_results.cpp
```diff
@@ -1563,6 +1563,19 @@ SQLPOINTER read_lob_field( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT field_in
...
 
+        if ( stmt->conn->ce_option.enabled == true ) {
+            // cursor type SQLSRV_CURSOR_BUFFERED has to be FORWARD_ONLY
+            // thus has to close and reopen cursor to reset the cursor buffer
+            core::SQLCloseCursor(stmt);
+            core::SQLExecute(stmt);
+            // FETCH_NEXT until the cursor reaches the row that it was at
+            for (int i = 0; i <= row_count; i++) {
+                core::SQLFetchScroll(stmt, SQL_FETCH_NEXT, 0);
+            }
```
yitam: Why not use the `offset` argument for SQLFetchScroll()?
Also, consider renaming input parameter `row_count` to something like `rows_read`

v-kaywon: FORWARD_ONLY cursor ignores the offset argument. SQLSRV_CURSOR_BUFFERED has to be FORWARD_ONLY.

source/shared/core_stmt.cpp
```diff
@@ -796,6 +798,9 @@ bool core_sqlsrv_fetch( _Inout_ sqlsrv_stmt* stmt, _In_ SQLSMALLINT fetch_orient
         // move to the record requested.  For absolute records, we use a 0 based offset, so +1 since 
         // SQLFetchScroll uses a 1 based offset, otherwise for relative, just use the fetch_offset provided.
         SQLRETURN r = stmt->current_results->fetch( fetch_orientation, ( fetch_orientation == SQL_FETCH_RELATIVE ) ? fetch_offset : fetch_offset + 1 TSRMLS_CC );
+        if ( stmt->cursor_type == SQL_CURSOR_FORWARD_ONLY && stmt->conn->ce_option.enabled == true ) {
+            stmt->fwd_row_index++;
+        }
```
yitam: Do we reset this `fwd_row_index` at some point? For example, when `past_fetch_end` is true? Or is it not necessary?

source/shared/core_stmt.cpp
```diff
@@ -2171,6 +2176,7 @@ void get_field_as_string( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT field_ind
 		else {
 			// Get the SQL type of the field. unixODBC 2.3.1 requires wide calls to support pooling
 			core::SQLColAttributeW( stmt, field_index + 1, SQL_DESC_CONCISE_TYPE, NULL, 0, NULL, &sql_field_type TSRMLS_CC );
+            SQLLEN sql_field_len = 0;
```
yitam: Probably `sql_field_len` is not used after this?

source/shared/core_stmt.cpp
```diff
@@ -2218,145 +2223,195 @@ void get_field_as_string( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT field_ind
...
-
-						} while( r == SQL_SUCCESS_WITH_INFO && is_truncated_warning( state ));
-					}
-					else {
-
-						// We got the field_len_temp from SQLGetData call.
-						field_value_temp = static_cast<char*>( sqlsrv_realloc( field_value_temp, field_len_temp + extra + 1 ));
-
-						// We have already recieved INITIAL_FIELD_STRING_LEN size data.
-						field_len_temp -= INITIAL_FIELD_STRING_LEN;
```
yitam: Looks like the block (lines 2261 to 2269) is the same as the code in the else block (lines 2274 to 2282)?
If so, is it possible to refactor the code?

meet-bhagdev:
@Hadis-Fard Sure! Let me know where the report is and i will take a look.

source/shared/core_stmt.cpp
```diff
@@ -2218,145 +2223,195 @@ void get_field_as_string( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT field_ind
                 if( is_truncated_warning( state ) ) {
 #endif // !_WIN32
```
yitam: will we get truncated warning if a client buffer is used, (the data has already been fetched beforehand), would it ever reach this part?
similarly, few lines after this, checking if AE is enabled, can we assume client buffer is not used?

source/shared/core_stmt.cpp
```diff
@@ -2218,145 +2223,195 @@ void get_field_as_string( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT field_ind
...
+                            if( r == SQL_SUCCESS_WITH_INFO ) {
+                                core::SQLGetDiagField( stmt, 1, SQL_DIAG_SQLSTATE, state, SQL_SQLSTATE_BUFSIZE, &len
+                                                       TSRMLS_CC );
+                            }
+
+                        } while( r == SQL_SUCCESS_WITH_INFO && is_truncated_warning( state ));
+                    }
+                    else {
+
+                        // We got the field_len_temp from SQLGetData call.
```
yitam: Please add a comment here what's this else block is (if not blah blah ?? )

source/shared/core_stmt.cpp
```diff
@@ -2218,145 +2223,195 @@ void get_field_as_string( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT field_ind
...
+                                core::SQLExecute( stmt );
+                                // FETCH_NEXT until the cursor reaches the row that it was at
+                                for ( int i = 0; i <= stmt->fwd_row_index; i++ ) {
+                                    core::SQLFetchScroll( stmt->current_results->odbc, SQL_FETCH_NEXT, 0 );
+                                }
+                            }
+                            else {
+                                core::SQLFetchScroll( stmt->current_results->odbc, SQL_FETCH_NEXT, 0 );
+                                core::SQLFetchScroll( stmt->current_results->odbc, SQL_FETCH_PRIOR, 0 );
+                            }
```
yitam: Please investigate whether it's possible to refactor the above specific code for AE's max fields

source/shared/core_results.cpp
yitam: Please investigate the feasibility of not using FORWARD_ONLY when AE is on

Hadis-Fard:
@meet-bhagdev It's mixed in the discussion thread, pasting it here:

Performance measurement for fetching varchar(max) columns with and without AE:

Performance test information:
Driver: PDO_SQLSRV
Table columns: char(5), char(5, deterministic encrypted), char(5, randomize encrypted), varchar(max), varchar(max, deterministic encrypted), varchar(max, randomize encrypted)
Number of rows in table: 2
Number of iterations: 1000
Number of measurements: 10 (each measurement is the sum of all 1000 iterations)

Mean(with AE): 98.71267s (SD = 19.78386)
Mean(without AE): 85.88191s (SD = 9.94258)

source/shared/core_results.cpp
v-kaywon: I've double checked. I get an error "Protocol error in TDS stream" when trying to fetch an encrypted max type column with non FORWARD_ONLY cursor (CURSOR_STATIC AND CURSOR_DYNAMIC). https://github.com/Microsoft/msphpsql/blob/dev/source/shared/core_stmt.cpp#L1126 is where the driver sets the cursor to SQL_CURSOR_FORWARD_ONLY when SQLSRV_CURSOR_BUFFERED is on.

yitam: Thanks for checking.

source/shared/core_stmt.cpp
```diff
@@ -796,6 +798,9 @@ bool core_sqlsrv_fetch( _Inout_ sqlsrv_stmt* stmt, _In_ SQLSMALLINT fetch_orient
         // move to the record requested.  For absolute records, we use a 0 based offset, so +1 since 
         // SQLFetchScroll uses a 1 based offset, otherwise for relative, just use the fetch_offset provided.
         SQLRETURN r = stmt->current_results->fetch( fetch_orientation, ( fetch_orientation == SQL_FETCH_RELATIVE ) ? fetch_offset : fetch_offset + 1 TSRMLS_CC );
+        if ( stmt->cursor_type == SQL_CURSOR_FORWARD_ONLY && stmt->conn->ce_option.enabled == true ) {
```
v-kaywon: even is the column itself is not encrypted, it still goes into this part of the code. Need further investigation to see if we can determine the encryption status of a column in a result set.

source/shared/core_stmt.cpp
v-kaywon: It's not necessary to reset fwd_row_index explicitly. past_fetch_end is set to true so the driver can raise an error when the user tries to fetch again even after SQL_NO_DATA. past_fetch_end and fwd_row_index is reset when the statement is re-executed and a new result set is created (https://github.com/v-kaywon/msphpsql/blob/22b4817f658a16edb8d24eba7eba2b2e44e2b01c/source/shared/core_stmt.cpp#L217)

source/shared/core_stmt.cpp
v-kaywon: we do get truncated warning (the driver try to mimic the same behavior as server cursor). The new commit doesn't check for client buffer anymore.

source/shared/core_stmt.cpp
```diff
@@ -796,6 +799,13 @@ bool core_sqlsrv_fetch( _Inout_ sqlsrv_stmt* stmt, _In_ SQLSMALLINT fetch_orient
         // move to the record requested.  For absolute records, we use a 0 based offset, so +1 since 
         // SQLFetchScroll uses a 1 based offset, otherwise for relative, just use the fetch_offset provided.
         SQLRETURN r = stmt->current_results->fetch( fetch_orientation, ( fetch_orientation == SQL_FETCH_RELATIVE ) ? fetch_offset : fetch_offset + 1 TSRMLS_CC );
+        
+        // when AE is on, need to store the current row number being fetched so a cursor can be reset back to it's original position when getting stream data
+        // fwd_row_index is used to store the current row index, which is incremented everytime SQLFetchScroll is called with SQLSRV_CURSOR_FORWARD_ONLY
```
yitam: When AE is *enabled*, will keep track of the numbers of rows being fetched *so far such that* ... back to *its* original position...

PR #501, opened 2017-08-11
Python scripts for building extensions for PHP 7+ in Windows

yitam:


buildscripts/buildtools.py
```diff
@@ -0,0 +1,437 @@
...
+            
+            file = branch + '.zip'
+            url = 'https://github.com/' + repo + '/msphpsql/archive/' + branch + '.zip'
+
+            print('Downloading ' + url + ' ...')
+            try:
+                with urllib.request.urlopen(url) as response, open(file, 'wb') as out_file:
+                    shutil.copyfileobj(response, out_file)
+            except:
+                print ("Resort to skip ssl verifcation...")
```
david-puglielli: verification

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,291 @@
...
+from buildtools import BuildUtil
+
+class BuildDriver(object):
+    """Build sqlsrv and/or pdo_sqlsrv drivers with PHP source with the following properties:
+    
+    Attributes:
+        util            # BuildUtil object whose constructor takes phpver, driver, arch, thread, debug 
+        repo            # GitHub repository
+        branch          # GitHub repository branch
+        download_source # download source from GitHub or not
```
lilgreenbird: this seems redundant, if you specified repo and branch it should mean that you are downloading source from github?

Hadis-Fard: on the other hand, it conveys the message that the users have a choice of getting it from GitHub or Local

yitam: in Python **all** the instance attributes have to be declared in the INIT function, and repo and branch are possible members that might be ignored if the user chooses *not* to download

Hadis-Fard:
@yitam please update the build windows section in readme file as well, and commit it to this PR

buildscripts/builddrivers.py
lilgreenbird: all these options seem so intertwined they overlap and have conflicting meanings. Eg. specifying github/branch and download_source = no and/or local/source_path makes no sense at all these are mutually exclusive options. The flags should be internal only they do not need to be externally visible. I think you only need 3 options (actually 2 if you combine gthub/branch) to convey this information. If Github/branch then download_source is obviously yes there's no need to specify that.  If source_path then obviously it's local no need to ask whether build is local or not. Actually, there's no need for the concept or "local" or "remote" at all. Basically there are 2 ways you can get the source, you either specify the path to the source (which could be local or remote unc path), or you specify the github repo.

yitam: No I think you've misunderstood. Some of these are flags or member variables in the script, such that I don't have to pass the arguments around or testing if some conditions are true again and again. Also, local or remote doesn't mean where to get the source from. Please go to the main function to get more details.

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,291 @@
...
+        package         # package name for the binaries (will be ignored for local builds)
+        remote_path     # remote destination to where the drivers will be placed (None for local builds)
```
lilgreenbird: this name is confusing. 1. it doesn't mention that this is the "destination" path 2. it's not necessarily a "remote" path, it could be any where you specify (any unc path, local or remote mount etc)

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,291 @@
...
+
+        if dest is None:       
+            # This indicates the script is NOT running locally, and that 
+            # the drivers should be in the working directory
+
+            # Make sure drivers path is defined
+            if self.remote_path is None:
+                print('Errors: Drivers destination should be defined! Do nothing.')
+            else:
+                OS_folder = "Windows" # hardcode this since this script is only run in Windows
```
lilgreenbird: why hardcode at all, makes it hard to maintain should this ever change and is unnecessary as the remote_path (which should really be named dest_path) will specify the location

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,291 @@
...
+    parser.add_argument('-v', '--PHPVER', help="PHP version, e.g. 7.1.*, 7.2.* etc.")
+    parser.add_argument('-a', '--ARCH', choices=['x64', 'x86'])
+    parser.add_argument('-t', '--THREAD', choices=['nts', 'ts'])
+    parser.add_argument('-d', '--DRIVER', choices=['all', 'sqlsrv', 'pdo_sqlsrv'])
+    parser.add_argument('-m', '--DEBUG', default='no', choices=['yes', 'no'], help="enable debug mode")
+    parser.add_argument('-r', '--REPO', default='Microsoft', help="GitHub repository")
+    parser.add_argument('-b', '--BRANCH', default='dev', help="GitHub repository branch")
+    parser.add_argument('-g', '--GITHUB', default='yes', help="get source from GitHub or not")
+    parser.add_argument('-k', '--PACKAGE', default='Latest', help="the package name for the drivers")
+    parser.add_argument('-p', '--PATH', default=None, help="the remote destination for the drivers (do not use this when building locally)")
```
lilgreenbird: DESTINATION_PATH (or DESTPATH) would be more descriptive as path could be any old path. It's also not necessarily "remote". And I MAY choose to do this when "building locally" that's up to me to decide how to use this script

buildscripts/buildtools.py
```diff
@@ -0,0 +1,465 @@
...
+    
+    def major_version(self):
+        """Return the major version number based on the PHP version."""
+        return self.phpver[0:3]
+        
+    def version_label(self):
+        """Return the version label based on the PHP version."""
+        major_ver = self.major_version()
+        
+        if major_ver == '7.0':
```
lilgreenbird: can we not hardcode 7 ie any version x.0

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,291 @@
...
+    branch = args.BRANCH
+    download = args.GITHUB.lower() == 'yes'
+    path = args.PATH
+    package = args.PACKAGE
+
+    if phpver is None:
+        # assuming it is building drivers locally when required to prompt
+        # thus will not prompt for drivers' destination path, which is None by default
+        phpver = input("PHP Version (e.g. 7.1.* or 7.2.*): ")
+        arch_version = input("Want to build 64-bit [y/n]: ")
```
lilgreenbird: the words "Want to build" is unnecessary" and is inconsistent with the rest of the UI

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,291 @@
+#!/usr/bin/python
```
lilgreenbird: the point might be moot in windows as I think this is ignored. But you said this requires python 3.4 so it should be /usr/bin/python3

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,291 @@
...
+#########################################################################################
+#
+# Description:  This script helps to build drivers in a Windows environment for PHP 7+ (32-bit/64-bit)
+#
+# Requirement:
+#               python 3.4
```
lilgreenbird: not necessarily 3.4, could be any 3.x

buildscripts/buildtools.py
```diff
@@ -0,0 +1,465 @@
+#!/usr/bin/python
```
lilgreenbird: /usr/bin/python3

buildscripts/buildtools.py
```diff
@@ -0,0 +1,465 @@
...
+#########################################################################################
+#
+# Description:  The class BuildUtil will build Microsoft SQL Server PHP 7+ Drivers 
+#               for 32 bit and 64 bit.
+#
+# Requirement:
+#               python 3.4
```
lilgreenbird: 3.x

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,291 @@
...
+        if dest is None:       
+            # This indicates the script is NOT running locally, and that 
+            # the drivers should be in the working directory
+
+            # Make sure drivers path is defined
+            if self.remote_path is None:
+                print('Errors: Drivers destination should be defined! Do nothing.')
+            else:
+                OS_folder = "Windows" # hardcode this since this script is only run in Windows
+                dest_drivers = os.path.join(self.remote_path, 'PHP', 'Drivers', self.package, OS_folder, self.util.major_version(), self.util.arch)
```
lilgreenbird: this is taking an argument and adding hardcoded parts to it!!
just leave it alone, let the user specify

buildscripts/builddrivers.py
yitam: My goal is to minimize the number of arguments, so `remote_path` is indeed *remote* because its default value **None** signifies this is a local build, which does things differently when it's not. For local build the expected behavior is to put the drivers in the 'ext' folder. So far there's no such requirement or use case to provide the flexibility to move the drivers anywhere else. Besides, the user (or php developer) can copy them to anywhere s/he likes afterwards. 

Since the user can run this script locally by providing arguments, if you insist I have to make this a generic *destination* path, I need some way to tell this is a local build. This means yet another argument, which seems to be something you're against based on your previous comments.

buildscripts/builddrivers.py
yitam: Please see my other comment above.

buildscripts/builddrivers.py
yitam: Please clarify how you want to reword the question.

buildscripts/builddrivers.py
yitam: I've asked @Hadis-Fard about this and she agreed. You might want to double check with her.

buildscripts/builddrivers.py
yitam: Again, it's to minimize the number of input arguments. Since `REPO` and `BRANCH` options have default values, for the user to choose *not* to download from GitHub (which is different from the default case), the user only needs to specify `--GITHUB=no` or simply `-g=no`

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,287 @@
...
+
+        self.util.build_drivers(self.make_clean, dest, logfile)
+
+        if dest is None:       
+            # This indicates the script is NOT running locally, and that 
+            # the drivers should be in the working directory
+
+            # Make sure drivers path is defined
+            if self.remote_path is None:
+                print('Errors: Drivers destination should be defined! Do nothing.')
```
david-puglielli: 'Doing nothing.' sounds better, otherwise it reads like an instruction to the user, which is confusing.

buildscripts/builddrivers.py
lilgreenbird: I still don't think it's necessary, BUT, I don't feel strongly about that, it's fine :)

yitam: Just curious, how else do you think the user can specify *not* to download from GitHub? ;)

buildscripts/builddrivers.py
lilgreenbird: it was nit-picking :) Just looks inconsistent as sometimes it asks if you "want to" do something sometimes it doesn't. Seem the 2 words "want to" is unnecessary.

buildscripts/builddrivers.py
lilgreenbird: if github/branch not specified then it's local :)

yitam: but those arguments have default values when not specified...

buildscripts/buildtools.py
```diff
@@ -0,0 +1,465 @@
...
+        return 'Build_' + datetime.datetime.now().strftime("%Y%m%d_%H%M") + '.log'
+    
+    @staticmethod
+    def update_file_content(file, search_str, new_str):
+        """Find *search_str* and replace it by *new_str* in a *file*"""
+        os.chmod(file, stat.S_IWRITE)
+        f = open(file, 'r')
+        filedata = f.read()
+        updatedata = filedata.replace(search_str, new_str)
+        f = open(file, 'w')
```
david-puglielli: Is it not necessary to close the file before opening it again writing?

yitam: Good point. Will fix this function

README.md
```diff
@@ -37,7 +37,7 @@ Thank you for taking time to take our February survey. Let us know how we are do
 
 ## Announcements
 
- Please visit the [blog][blog] for more announcements.
+Please visit the [blog][blog] for more announcements.
 
 
 ## Build (Windows)
```
lilgreenbird: I couldn't click on the line below..
"you can use the pre-compiled binarIES"

README.md
```diff
@@ -46,23 +46,29 @@ Note: if you prefer, you can use the pre-compiled binary found [HERE](https://gi
 
 #### Prerequisites
 
-You must first be able to build PHP 7 without including these extensions.  For help with doing this, see the [official PHP website][phpbuild] for building your own PHP on Windows.
+You must first be able to build PHP 7 without including these extensions. For help with doing this, see the [official PHP website][phpbuild] for building your own PHP on Windows.
 
 #### Compile the drivers
```
lilgreenbird: Usually in our other READMEs the commands that a user is expected to run are in code blocks it makes it easier to read and c&p the commands and also keep things consistent

README.md
```diff
@@ -46,23 +46,29 @@ Note: if you prefer, you can use the pre-compiled binary found [HERE](https://gi
...
+5. Run `configure.bat --with-odbcver=0x0380 and the desired driver options (as below) [plus other options such as --disable-zts for the Non Thread Safe build]` to generate the makefile.  You can run `configure.bat --help` to see what other options are available.
+    * For SQLSRV use: `--enable-sqlsrv=shared`
+    * For PDO_SQLSRV use: `--enable-pdo --with-pdo-sqlsrv=shared`
 
-This software has been compiled and tested under PHP 7.0.20 and 7.1.6 using the Visual C++ 2015 compiler.
+6. Run `nmake`. Optionally, you can run `nmake clean` first.
+
+7. To install the resulting build, run `nmake install` or just copy php_sqlsrv.dll and/or php_pdo_sqlsrv.dll to your PHP extension directory.
+
+The extensions have been compiled and tested with the PHP 7.0.* and 7.1.* using the Visual C++ 2015 compiler as well as PHP 7.2.0 beta using the Visual C++ 2017 V15.0.
```
lilgreenbird: "the" PHP is a typo?

README.md
```diff
@@ -46,23 +46,29 @@ Note: if you prefer, you can use the pre-compiled binary found [HERE](https://gi
...
+    * For PDO_SQLSRV use: `--enable-pdo --with-pdo-sqlsrv=shared`
 
-This software has been compiled and tested under PHP 7.0.20 and 7.1.6 using the Visual C++ 2015 compiler.
+6. Run `nmake`. Optionally, you can run `nmake clean` first.
+
+7. To install the resulting build, run `nmake install` or just copy php_sqlsrv.dll and/or php_pdo_sqlsrv.dll to your PHP extension directory.
+
+The extensions have been compiled and tested with the PHP 7.0.* and 7.1.* using the Visual C++ 2015 compiler as well as PHP 7.2.0 beta using the Visual C++ 2017 V15.0. 
+
+For an example of how to automate building the Microsoft Drivers for PHP for SQL Server, please browse the *buildscripts* directory.
```
lilgreenbird: ?? we ARE in the build scripts directory that's where this README file is

lilgreenbird: hmm those are tools only not really an example of how to do it. How about something like  "The buildscripts directory contain sample scripts that help automate the building of the blah driver."?

buildscripts/README.md
```diff
@@ -0,0 +1,56 @@
+## Windows
+
+### Prerequisites
+To use the Python build scripts `builddrivers.py` and `buildtools.py`, install Python 3.x.
```
lilgreenbird: keep it consistent with the other README and also easier to read to have the list in bullets

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,287 @@
...
+            if build_choice == 'r':
+                print('Will rebuild the binaries')
+                # only the old binaries based on the current configuration will be removed
+                self.util.remove_prev_build(root_dir)
+            elif build_choice == 'c':
+                print('Will make clean')
+                self.make_clean = True
+                # all old builds are removed, and this step is necessary because 
+                # the user might have changed the configuration
+                self.util.remove_old_builds(root_dir)
```
david-puglielli: Does this option ever call 'nmake clean'? It seems that remove_old_builds removes the binaries but not the object files. Does PHP rebuild the object files if the configuration changes?

buildscripts/README.md
```diff
@@ -0,0 +1,56 @@
+## Windows
+
+### Prerequisites
+To use the Python build scripts `builddrivers.py` and `buildtools.py`, install Python 3.x. 
+
+To build extensions for PHP 7.0* or 7.1*, install Visual Studio 2015 and make sure C++ tools are enabled.
```
lilgreenbird: see previous comments on using code blocks for commands.

buildscripts/builddrivers.py
david-puglielli: OK I see this is done when creating the batch script.

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,287 @@
...
+class BuildDriver(object):
+    """Build sqlsrv and/or pdo_sqlsrv drivers with PHP source with the following properties:
+    
+    Attributes:
+        util            # BuildUtil object whose constructor takes phpver, driver, arch, thread, debug 
+        repo            # GitHub repository
+        branch          # GitHub repository branch
+        download_source # download source from GitHub or not
+        remote_path     # remote destination to where the drivers will be placed (None for local builds)
+        local           # a boolean flag - whether the build is local
```
david-puglielli: Isn't php always going to be built locally with this script? How is a 'remote' build done? Is it just a matter of switching to a remote destination after building the binaries locally?

yitam: For example, a Bamboo build plan can use the same Python scripts to build the drivers and put them in a designated folder. In README it explains the differences.

david-puglielli:
Some of the Readme changes conflict with those in #503. Why don't I incorporate your changes into that PR and remove the Readme changes here to avoid conflicts?

buildscripts/README.md
```diff
@@ -84,13 +84,11 @@ When something goes wrong during the build, the log file will be launched (you c
 
 In addition to the log files in `C:\php-sdk`, you can examine the contents of `C:\php-sdk\phpsdk-build-task.bat`, which is overwritten every time you run the build scripts.
 
-#### Building the extensions unattended
+#### Variation
```
david-puglielli: I still don't like this name 'variation'... maybe 'Setting alternative destination directory' or something like that?

buildscripts/README.md
```diff
@@ -0,0 +1,97 @@
...
+        * Locate the compiled php_sqlsrv.dll and/or php_pdo_sqlsrv.dll 
+        * Copy the dll(s) to the `ext` subfolder  
+
+### Using the sample build scripts
+
+The sample build scripts, `builddrivers.py` and `buildtools.py`, are expected to build our extensions for PHP in Windows.
+
+#### Overview
+
+When asked to provide the PHP version, you should enter values like `7.1.7`. If it's alpha, beta, or RC, make sure the name you provide matches the PHP tag name without the prefix `php-`. For example, for PHP 7.2 beta 2, the tag name is `php-7.2.0beta2`, so you should enter `7.2.0beta2`. Visit [PHP SRC]( https://github.com/php/php-src) to find the appropriate tag names.
```
david-puglielli: 'If it is an alpha, beta, or RC version, make sure...'

buildscripts/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+It's recommended that the PHP SDK is unzipped into the shortest possible path, preferrably somewhere near the root drive. Therefore, this script will create a `php-sdk` folder in the C:\ drive. This `php-sdk` folder will remain unless you remove it yourself. For ongoing development, it's suggested you keep it around. The build scripts will handle updating the PHP SDK if new version is available.
```
david-puglielli: 'It's' here too... though this could be changed to say who 'it' refers to. PHP developers presumably? Something like 'PHP recommends that the PHP SDK...', 'For ongoing development, PHP suggests..' or 'we suggest...' if 'it' refers to us instead.

buildscripts/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+When something goes wrong during the build, the log file will be launched (you can find the log files in `C:\php-sdk`). Otherwise, the log file will not be shown, and they remain in `C:\php-sdk` until you remove them manually.
+
+In addition to the log files in `C:\php-sdk`, you can examine the contents of `C:\php-sdk\phpsdk-build-task.bat`, which is overwritten every time you run the build scripts.
+
+#### Variation
+
+If you want the PHP drivers to be copied to somewhere else, you can invoke `py builddrivers.py` by providing the option `--DESTPATH=<some valid path>`. 
+
+In this case, you will find a copy of the drivers (unless the build fails) and the `php-sdk` folder created in the same directory of these Python scripts. It's your choice whether to remove the `php-sdk` folder and/or the drivers' binaries afterwards.
```
david-puglielli: It is your choice...

buildscripts/README.md
```diff
@@ -0,0 +1,97 @@
...
+        * Find the directory where the newly compiled `php.exe` is
+        * Locate the compiled php_sqlsrv.dll and/or php_pdo_sqlsrv.dll 
+        * Copy the dll(s) to the `ext` subfolder  
+
+### Using the sample build scripts
+
+The sample build scripts, `builddrivers.py` and `buildtools.py`, are expected to build our extensions for PHP in Windows.
+
+#### Overview
+
```
david-puglielli: The it's below are too informal - I don't think contraction show up anywhere else in this document.

buildscripts/README.md
```diff
@@ -0,0 +1,97 @@
...
+It's recommended that the PHP SDK is unzipped into the shortest possible path, preferrably somewhere near the root drive. Therefore, this script will create a `php-sdk` folder in the C:\ drive. This `php-sdk` folder will remain unless you remove it yourself. For ongoing development, it's suggested you keep it around. The build scripts will handle updating the PHP SDK if new version is available. 
+
+#### Steps
+
+1. Launch a regular `cmd` prompt 
+
+2. Change to the directory where the Python scripts `builddrivers.py` and `buildtools.py` are
+
+3. Interactive mode: 
+    * Run `py builddrivers.py` to use the interactive mode. Use lower cases to answer the following questions:
```
lilgreenbird: lower case is not necessary

yitam: why not? the scripts expect lower cases...

buildscripts/README.md
```diff
@@ -0,0 +1,97 @@
...
+        * PHP Version (i.e. the version number like `7.1.7` or `7.2.0beta2`)
```
lilgreenbird: (e.g. 7.1.7 or 7.2.0beta2)

buildscripts/README.md
```diff
@@ -0,0 +1,97 @@
...
+        * 64-bit?
+        * Thread safe?
+        * Driver?
+        * Debug enabled?
+        * Download source from GitHub?
+    * For `yes/no` questions, you can simply hit `ENTER` key for `yes`. Other questions are self-explanatory.
```
lilgreenbird: aren't the options presented after the choices? this line isn't necessary

buildscripts/README.md
```diff
@@ -0,0 +1,97 @@
...
+    
+4. Use Command-line arguments
+    * Run `py builddrivers.py -h` to get a list of options and their descriptions
+    * For example, 
+        * `py builddrivers.py -v=7.1.7 -a=x86 -t=ts -d=all -g=no`
+        * `py builddrivers.py --PHPVER=7.0.22 --ARCH=x64 --THREAD=nts --DRIVER=sqlsrv --GITHUB=yes`
+
+5. Based on the given configuration, if the script detects the presence of the PHP source directory, you can choose whether to rebuild, clean or superclean:
+    * `rebuild` if you have always used the same configuration (32 bit, thread safe, etc.)
```
lilgreenbird: "to build again using the same configuration..."

buildscripts/README.md
yitam: hmmm... I'm not sure I understand... the choices are `yes` or `no` but the user doesn't necessarily know they can hit the `ENTER` as `yes`?

buildscripts/README.md
lilgreenbird: this is confusing and is overloading the meaning of the DESTPATH argument.

buildscripts/builddrivers.py
lilgreenbird: This is why the whole local/remote terminology is confusing. Local and remote really just refers to where the source is, has nothing to do with where the driver binaries will be placed which in the case of the test lab this is actually not remote it's a net mount.
Anyway, the build script arg has been changed to destpath which avoids this confusion so this should be renamed to match

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,287 @@
...
+        :param  root_dir: the C:\ drive
+        :param  work_dir: the directory of this script
+        :outcome: the old binaries, if exist, will be removed
+        """
+        phpsrc = self.util.phpsrc_root(root_dir)
+        if os.path.exists( phpsrc ):
+            print(phpsrc + " exists.")
+            print("Choose rebuild(r) if using the same configuration. Choose clean(c) otherwise. If unsure, choose superclean(s).") 
+            build_choice = validate_input("Want to rebuild (r), clean (c) or superclean (s)? ", "r/c/s")
+            self.make_clean = False
```
lilgreenbird: can we shorten this and not use 2nd person narrative?
eg
"(r)ebuild to use same configuration, (c)lean otherwise, (s)uperclean if unsure"

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,287 @@
...
+
+            try:
+                dest = None
+                if self.local:
+                    dest = root_dir
+                    
+                self.build_extensions(dest, logfile)
+                print('Build Completed')
+            except:
+                print('Something went wrong. Build incomplete.')
```
lilgreenbird: Build failed

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,287 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-v', '--PHPVER', help="PHP version, e.g. 7.1.*, 7.2.* etc.")
+    parser.add_argument('-a', '--ARCH', choices=['x64', 'x86'])
+    parser.add_argument('-t', '--THREAD', choices=['nts', 'ts'])
+    parser.add_argument('-d', '--DRIVER', default='all', choices=['all', 'sqlsrv', 'pdo_sqlsrv'], help="driver to build (default: all)")
+    parser.add_argument('-m', '--DEBUG', default='no', choices=['yes', 'no'], help="enable debug mode (default: no)")
+    parser.add_argument('-r', '--REPO', default='Microsoft', help="GitHub repository (default: Microsoft)")
+    parser.add_argument('-b', '--BRANCH', default='dev', help="GitHub repository branch (default: dev)")
+    parser.add_argument('-g', '--GITHUB', default='yes', choices=['yes', 'no'], help="get source from GitHub (default: yes)")
+    parser.add_argument('-p', '--DESTPATH', default=None, help="the remote destination for the drivers (default: None)")
```
lilgreenbird: remove word "remote" as it's not necessarily remote and is not the case in our test lab.
it's the unc path to where the driver binaries will be copied to

buildscripts/builddrivers.py
```diff
@@ -0,0 +1,287 @@
...
+                if self.local:          # display log file only when building locally
+                    os.startfile(os.path.join(root_dir, 'php-sdk', logfile))
+                os.chdir(work_dir)    
+                break
+
+            # Only ask when building locally
+            if self.local:               
+                choice = input("Rebuild using the same configuration(yes) or quit (no) [yes/no]: ")
+
+                if choice.lower() == 'yes' or choice.lower() == 'y' or choice.lower() == '':
```
lilgreenbird: do choice.lower() only once?

buildscripts/README.md
```diff
@@ -0,0 +1,97 @@
...
+    * `clean` to remove previous builds (binaries) 
+    * `superclean` to remove the entire `php-<version>-src` directory, which is often unnecessary
+
+6. If you choose not to download from a GitHub repository, you will be asked to provide the full path to your local Source folder.
+
+7. If the compilation is successful, you will be given the option to rebuild or quit. 
+
+#### Troubleshooting
+
+When something goes wrong during the build, the log file will be launched (you can find the log files in `C:\php-sdk`). Otherwise, the log file will not be shown, and they remain in `C:\php-sdk` until you remove them manually.
```
lilgreenbird: something went wrong
actually how about just "if the build failed"?

yitam: Not necessarily only when the build failed. It can be somewhere in the scripts that went wrong.

buildscripts/README.md
```diff
@@ -43,9 +43,9 @@ The sample build scripts, `builddrivers.py` and `buildtools.py`, are expected to
...
+When asked to provide the PHP version, you should enter values like `7.1.7`. If it's alpha, beta, or RC version, make sure the name you provide matches the PHP tag name without the prefix `php-`. For example, for PHP 7.2 beta 2, the tag name is `php-7.2.0beta2`, so you should enter `7.2.0beta2`. Visit [PHP SRC]( https://github.com/php/php-src) to find the appropriate tag names.
 
-It's recommended that the PHP SDK is unzipped into the shortest possible path, preferrably somewhere near the root drive. Therefore, this script will create a `php-sdk` folder in the C:\ drive. This `php-sdk` folder will remain unless you remove it yourself. For ongoing development, it's suggested you keep it around. The build scripts will handle updating the PHP SDK if new version is available. 
+PHP recommendeds to unzip the PHP SDK into the shortest possible path, preferrably somewhere near the root drive. Therefore, this script will create a `php-sdk` folder in the C:\ drive. This `php-sdk` directory tree will remain unless you remove it yourself. For ongoing development, we suggest you keep it around. The build scripts will handle updating the PHP SDK if a new version is available.
```
yitam: I will fix the typo here

buildscripts/README.md
lilgreenbird: the script changes everything to lower, so it doesn't matter what the user enter?

buildscripts/README.md
```diff
@@ -80,15 +80,16 @@ It's recommended that the PHP SDK is unzipped into the shortest possible path, p
...
 
-#### Variation
+#### Setting alternative destination for drivers
 
-If you want the PHP drivers to be copied to somewhere else, you can invoke `py builddrivers.py` by providing the option `--DESTPATH=<some valid path>`. 
+If your main goal is to build the drivers, and/or there is no need to keep the `php-sdk` directory around, you can invoke `py builddrivers.py` with the necessary command-line arguments plus `--DESTPATH=<some valid path>`, which is **None** by default. Note that this option is not available in the interactive mode. 
+
+By setting an alternative destination automatically turns off the looping mechanism. When the build is finished, you will find a copy of the drivers (unless the build failed) and the `php-sdk` folder in the same directory of these Python scripts. This option is particularly useful in a test environment (or a virtual machine) in which these build scripts are copied to a temporary folder. After the drivers have been successfully compiled and copied to the designated location, the temporary folder can be safely removed afterwards.
```
david-puglielli: 'Setting an alternative destination automatically turns off ...' or 'By setting an alternative destination, the looping mechanism is turned off.'

Also last sentence, there are two 'afters', you can remove 'afterwards'.

PR #511, opened 2017-08-23
Added support to load a custom keystore provider for Column Encryption

yitam:


source/shared/core_conn.cpp
```diff
@@ -776,7 +778,68 @@ void determine_server_version( _Inout_ sqlsrv_conn* conn TSRMLS_DC )
...
+    
+    // Next, extract the character string from conn->ce_option.ksp_encrypt_key into encrypt_key
+    char* encrypt_key = Z_STRVAL_P( conn->ce_option.ksp_encrypt_key );    
+    memcpy_s( pKsd->data, key_size * sizeof( char ) , encrypt_key, key_size );
+
+    core::SQLSetConnectAttr( conn, SQL_COPT_SS_CEKEYSTOREPROVIDER, ksp_path, SQL_NTS );
+    core::SQLSetConnectAttr( conn, SQL_COPT_SS_CEKEYSTOREDATA, reinterpret_cast<SQLPOINTER>( pKsd ), SQL_IS_POINTER );
+}
+
+void common_conn_str_append_func( const char* odbc_name, const char* val, size_t val_len, std::string& conn_str TSRMLS_DC )
```
david-puglielli: Missing SAL annotations for this function - they were previously included here

source/shared/core_conn.cpp
```diff
@@ -776,7 +778,68 @@ void determine_server_version( _Inout_ sqlsrv_conn* conn TSRMLS_DC )
     conn->server_version = version_major;
 }
 
-void common_conn_str_append_func( _In_z_ const char* odbc_name, _In_reads_(val_len) const char* val, _Inout_ size_t val_len, _Inout_ std::string& conn_str TSRMLS_DC )
+// Column Encryption feature: if a custom keystore provider is specified, 
+// load and configure it when column encryption is enabled, but this step have
```
david-puglielli: this step *has*

source/shared/core_conn.cpp
```diff
@@ -776,7 +778,68 @@ void determine_server_version( _Inout_ sqlsrv_conn* conn TSRMLS_DC )
...
+    char* ksp_path = Z_STRVAL_P( conn->ce_option.ksp_path );
+    unsigned int name_len = Z_STRLEN_P( conn->ce_option.ksp_name );
+    unsigned int path_len = Z_STRLEN_P( conn->ce_option.ksp_path );
+    unsigned int key_size = conn->ce_option.key_size;    
+
+    sqlsrv_malloc_auto_ptr<unsigned char> ksp_data;
+
+    ksp_data = reinterpret_cast<unsigned char*>( sqlsrv_malloc( sizeof( CEKEYSTOREDATA ) + key_size ) );
+
+    CEKEYSTOREDATA *pKsd = (CEKEYSTOREDATA*) ksp_data.get();
```
david-puglielli: ksp_data appears to be some sort of intermediary - why is it needed? Why not just have `CEKEYSTOREDATA *pKsd;`?

.travis.yml
```diff
@@ -29,6 +32,12 @@ script:
   - docker ps -a 
   - docker exec client python ./test/functional/setup/setup_dbs.py -dbname $SQLSRV_DBNAME
   - docker exec client python ./test/functional/setup/setup_dbs.py -dbname $PDOSQLSRV_DBNAME
+  - docker exec client cp ./source/shared/msodbcsql.h ./test/functional/setup/
```
v-kaywon: does msodbcsql.h has to be in the setup folder the compiling and building msKSP.so?

source/shared/core_conn.cpp
```diff
@@ -776,7 +778,68 @@ void determine_server_version( _Inout_ sqlsrv_conn* conn TSRMLS_DC )
...
+    // First, convert conn->ce_option.ksp_name to a WCHAR version 
+    unsigned int wname_len = 0;
+    sqlsrv_malloc_auto_ptr<SQLWCHAR> wksp_name;
+    wksp_name = utf16_string_from_mbcs_string( SQLSRV_ENCODING_UTF8, ksp_name, name_len, &wname_len );
+
+    CHECK_CUSTOM_ERROR( wksp_name == 0, conn, SQLSRV_ERROR_CONNECT_STRING_ENCODING_TRANSLATE ) {
+	    throw core::CoreException();
+    }
+
+    pKsd->name = (wchar_t *) wksp_name.get();
```
david-puglielli: Is it necessary to cast to wchar_t? wchar_t is 16 bits on Windows and 32 on Linux, and wksp_name is defined with SQLWCHAR, which should resolve to wchar_t iirc. But if not, can this lead to cross-platform incompatibility?

source/shared/core_conn.cpp
```diff
@@ -776,7 +778,68 @@ void determine_server_version( _Inout_ sqlsrv_conn* conn TSRMLS_DC )
...
+    
+    // Next, extract the character string from conn->ce_option.ksp_encrypt_key into encrypt_key
+    char* encrypt_key = Z_STRVAL_P( conn->ce_option.ksp_encrypt_key );    
+    memcpy_s( pKsd->data, key_size * sizeof( char ) , encrypt_key, key_size );
```
david-puglielli: Does key_size account for the null terminator?

source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -221,15 +225,42 @@ const connection_option PDO_CONN_OPTS[] = {
...
+	},
+	{
+		PDOConnOptionNames::CEKeystoreEncryptKey,
+		sizeof(PDOConnOptionNames::CEKeystoreEncryptKey),
+		SQLSRV_CONN_OPTION_CEKEYSTORE_ENCRYPT_KEY,
+		ODBCConnOptions::CEKeystoreEncryptKey,
+		sizeof(ODBCConnOptions::CEKeystoreEncryptKey),
+		CONN_ATTR_STRING,
+		ce_ksp_provider_set_func::func
+	},
```
david-puglielli: The indentation is off...

source/shared/core_sqlsrv.h
```diff
@@ -2416,4 +2430,39 @@ struct column_encryption_set_func {
...
+                conn->ce_option.ksp_encrypt_key = value;
+                conn->ce_option.key_size = value_len;
+                conn->ce_option.ksp_required = true;
+                break;
+            default:
+                SQLSRV_ASSERT( false, "ce_ksp_provider_set_func: Invalid KSP option!" );
+                break;
+        }
+    }
+};
```
david-puglielli: Are all three keywords required if the user wants to use a custom keystore?

source/sqlsrv/conn.cpp
```diff
@@ -294,24 +298,51 @@ const connection_option SS_CONN_OPTS[] = {
...
+	},
+	{
+		SSConnOptionNames::CEKeystoreEncryptKey,
+		sizeof(SSConnOptionNames::CEKeystoreEncryptKey),
+		SQLSRV_CONN_OPTION_CEKEYSTORE_ENCRYPT_KEY,
+		ODBCConnOptions::CEKeystoreEncryptKey,
+		sizeof(ODBCConnOptions::CEKeystoreEncryptKey),
+		CONN_ATTR_STRING,
+		ce_ksp_provider_set_func::func
+	},
```
david-puglielli: Indentation is off

test/functional/setup/ksp_app.c
```diff
@@ -0,0 +1,300 @@
...
+    if (!(ppKsp = (CEKEYSTOREPROVIDER**)GetProcAddress(hProvLib, "CEKeystoreProvider"))) {
+#else
+    if (!(ppKsp = (CEKEYSTOREPROVIDER**)dlsym(hProvLib, "CEKeystoreProvider"))) {
+#endif
+        fprintf(stderr, "The export CEKeystoreProvider was not found in the KSP library\n");
+        return 3;
+    }
+    while (pKsp = *ppKsp++) {
+        if (!memcmp(KSPNAME, pKsp->Name, sizeof(KSPNAME)))
+            goto FoundProv;
```
david-puglielli: goto? really?

test/functional/setup/build_ksp.py
```diff
@@ -0,0 +1,122 @@
...
+
+# This compiles ksp app, which assumes the existence of the .dll or the .so file.
+#
+# In Windows, a batch file is created in order to compile the code. 
+def configure_KSP(app_name):
+    print('Compiling ', app_name)
+    if platform.system() == 'Windows':
+        command = 'cl /MD {0}.c /link odbc32.lib /out:ksp_app.exe'.format(app_name)
+        batchfile = 'build_app.bat'
+        create_batch_file('x86', batchfile, command)
```
v-kaywon: Why do you need to write the commands to a batch file and then executing the whole thing? Why can't you just run the command one by one in the script?

test/functional/setup/build_ksp.py
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        os.system('gcc -fshort-wchar -fPIC -o myKSP.so -shared {0}.c'.format(ksp_name))
+
+# This compiles ksp app, which assumes the existence of the .dll or the .so file.
+#
+# In Windows, a batch file is created in order to compile the code. 
+def configure_KSP(app_name):
+    print('Compiling ', app_name)
+    if platform.system() == 'Windows':
+        command = 'cl /MD {0}.c /link odbc32.lib /out:ksp_app.exe'.format(app_name)
```
v-kaywon: not a big deal, but would be nice if the order of these arguments are consistent with the ones in compile_KSP_windows

test/functional/setup/build_ksp.py
```diff
@@ -0,0 +1,122 @@
...
+        batchfile = 'build_app.bat'
+        create_batch_file('x86', batchfile, command)
+        os.system(batchfile)
+        os.remove(batchfile)        
+    else:
+        os.system('gcc -o ksp_app -fshort-wchar {0}.c -lodbc -ldl'.format(app_name))
```
v-kaywon: make the ordering of the arguments consistent with the ones in conpile_KSP

test/functional/setup/build_ksp.py
```diff
@@ -0,0 +1,122 @@
...
+
+################################### Main Function ###################################
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-ksp', '--KSPNAME', default='myKSP', help='The KSP (keystore provider)')
+    parser.add_argument('-app', '--APPNAME', default='ksp_app', help='The app that uses the KSP')
+    args = parser.parse_args()
+
+    ksp_name = args.KSPNAME
+    app_name = args.APPNAME
```
v-kaywon: these variable names are a little confusing. These variables correspond to the name of the C file that is used to compile, but the variable names sound like they are the names of the dlls or executables. Maybe rename it to something like ksp_src_name and app_src_name.

Another not major thing at all is you may want to consider making the names to `args.KSPNAME + '.c`' so you don't need to append the '.c' everywhere.

test/functional/setup/build_ksp.py
```diff
@@ -0,0 +1,122 @@
...
+#               - myKSP.dll (myKSPx64.dll) / myKSP.so 
+#               - ksp_app.exe / ksp_app
+#
+# Requirement:
+#               python 3.x
+#               myKSP.c (or any equivalent)
+#               ksp_app.c (or any equivalent)
+#               msodbcsql.h (odbc header file)
+#
+# Execution: Run with command line with required options
```
v-kaywon: *optional* options?

test/functional/setup/ksp_app.c
```diff
@@ -0,0 +1,300 @@
...
+    checkRC(rc, "allocating connection handle", 4, env, SQL_HANDLE_ENV);
+    rc = SQLDriverConnect(dbc, 0, connStr, strlen(connStr), NULL, 0, NULL, SQL_DRIVER_NOPROMPT);
+    checkRC(rc, "connecting to data source", 5, dbc, SQL_HANDLE_DBC);
+    rc = SQLAllocHandle(SQL_HANDLE_STMT, dbc, &stmt);
+    checkRC(rc, "allocating statement handle", 6, dbc, SQL_HANDLE_DBC);
+
+    if (task == set_up) {
+        printf("Setting up KSP...\n");       
+        setKSPLibrary(stmt);
+        populateTestTable(dbc, stmt);
```
v-kaywon: Is there a reason why we're creating and populating a table here? wouldn't it be better to make and populate the tables on fly in the tests?

test/functional/setup/ksp_app.c
```diff
@@ -0,0 +1,300 @@
...
+    }
+    else if (task == clean_up) {
+        printf("Cleaning up KSP...\n");
+
+        SQLExecDirect(stmt, "DROP TABLE CustomKSPTestTable", SQL_NTS);
+        SQLExecDirect(stmt, "DROP COLUMN ENCRYPTION KEY CustomCEK", SQL_NTS);
+        SQLExecDirect(stmt, "DROP COLUMN MASTER KEY CustomCMK", SQL_NTS);
+        printf("Removed table, CEK, and CMK\n");
```
v-kaywon: If the database is dropped at the end, don't need this part

source/shared/core_conn.cpp
yitam: yes I use sqlsrv_malloc_auto_ptr to take care of releasing the memory afterwards

source/shared/core_conn.cpp
yitam: Please check the keystore definitions in msodbcsql.h, and they used wchar_t in quite a few places. In Linux, when compiling the KSP source and the app, we use `-fshort-wchar`

source/shared/core_conn.cpp
yitam: Good question. The answer is no. Only the length of the key.

source/shared/core_sqlsrv.h
yitam: yes, as you can see in the helper function load_configure_ksp()

test/functional/pdo_sqlsrv/skipif_server_old.inc
```diff
@@ -0,0 +1,19 @@
...
+
+if (!extension_loaded("pdo") || !extension_loaded('pdo_sqlsrv'))
+    die("PDO driver cannot be loaded; skipping test.\n");
+
+require_once( "MsSetup.inc" );
+$conn = new PDO("sqlsrv:server = $server;", $uid, $pwd );
+if( ! $conn )
+{
+    die( "skip - could not connect during SKIPIF." );
+}
```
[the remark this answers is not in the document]
yitam: I got the idea from skipif_protocol_not_tcp.inc :)

test/functional/setup/ksp_app.c
yitam: I basically got this from the sample code, but of course I can avoid using `goto` in this case.

test/functional/setup/build_ksp.py
yitam: because I need to invoke vcvarsall.bat first with the appropriate option (for x64 or x86) in order to use the c compiler cl.exe

test/functional/setup/build_ksp.py
yitam: I followed the ODBC sample codes on how to compile the source code. Besides, one is to compile the source into a DLL and the other is to create an *.exe, so the switches are different.

test/functional/setup/ksp_app.c
yitam: yes I agree but I provided this as a handy way to remove the KSP related entities in case the user doesn't want to drop the entire database

test/functional/setup/ksp_app.c
yitam: I agree, but as of today I can only insert data into the test table with encryption using sqlsrv...
For now the KSP tests show that they can retrieve data from a pre-populated table with encryption

test/functional/setup/build_ksp.py
yitam: I changed the arguments to be `KSPSRC` and `APPSRC` respectively

test/functional/setup/build_ksp.py
yitam: again, I basically followed the examples given in the ODBC sample code

test/functional/pdo_sqlsrv/skipif_server_old.inc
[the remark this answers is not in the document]
v-kaywon: hmmmm I agree with David too. If somehow it can't connect, the test should fail instead of being skipped. Should fix this in skipif_protocol_not_tcp.inc too (but maybe in a separate PR to the dev branch)

test/functional/pdo_sqlsrv/pdo_connect_encrypted_ksp.phpt
```diff
@@ -0,0 +1,54 @@
...
+        echo "Failed to connect.\n";
+        print_r( $e->getMessage() );
+        echo "\n";
+    }
+   
+    $tsql = "SELECT * FROM CustomKSPTestTable";
+    $stmt = $conn->query($tsql);
+    while ($row = $stmt->fetch(PDO::FETCH_NUM))
+    {
+        echo "c1=" . $row[0] . "\tc2=" . $row[1] . "\tc3=" . $row[2] . "\tc4=" . $row[3] . "\n";
```
v-kaywon: the danger of tabs..

test/functional/pdo_sqlsrv/skipif_server_old.inc
```diff
@@ -0,0 +1,19 @@
...
+
+$attr = $conn->getAttribute(constant('PDO::ATTR_SERVER_VERSION'));
+$version = substr($attr, 0, 2);
+if ($version < 13)
+{
+    die( "skip - feature not supported in this version of SQL Server." );
+}
```
v-kaywon: Should we not check for ODBC version too? at least exclude 11 and lower.

test/functional/sqlsrv/skipif_server_old.inc
```diff
@@ -0,0 +1,25 @@
...
+$server_info = sqlsrv_server_info( $conn );  
+if( $server_info )  
+{  
+    // check SQL Server version
+    $version = substr($server_info['SQLServerVersion'], 0, 2);
+    if ($version < 13)
+    {
+        die( "skip - feature not supported in this version of SQL Server." );
+    }
+}
```
v-kaywon: same comments as pdo_sqlsrv/skipif_server_old.inc

test/functional/sqlsrv/sqlsrv_connect_encrypted_ksp_errors.phpt
```diff
@@ -0,0 +1,150 @@
...
+<?php
+
+    function connect( $server, $connectionInfo )
+    {
+        $conn = sqlsrv_connect( $server, $connectionInfo );
+        if( $conn === false )
+        {
+            echo "Failed to connect.\n";
+            $errors = sqlsrv_errors();
+            print_r( $errors[0] );
```
v-kaywon: this prints an array with numeric and associative indexes, which makes the expected output unnecessarily long

source/shared/core_sqlsrv.h
```diff
@@ -2416,4 +2430,39 @@ struct column_encryption_set_func {
     }
 };
 
+struct ce_ksp_provider_set_func {
+
+    static void func( _In_ connection_option const* option, _In_ zval* value, _Inout_ sqlsrv_conn* conn, _Inout_ std::string& conn_str TSRMLS_DC)
+    {
+        CHECK_CUSTOM_ERROR( Z_TYPE_P( value ) != IS_STRING, conn, SQLSRV_ERROR_INVALID_OPTION_TYPE_STRING, option->odbc_name ) {
```
v-kaywon: why odbc_name? Isn't this error raised by our drivers?

source/shared/core_conn.cpp
```diff
@@ -776,6 +778,67 @@ void determine_server_version( _Inout_ sqlsrv_conn* conn TSRMLS_DC )
...
+    }
+
+    CHECK_CUSTOM_ERROR( conn->ce_option.key_size == 0, conn, SQLSRV_ERROR_KEYSTORE_KEY_MISSING) {
+        throw core::CoreException();
+    }
+
+    char* ksp_name = Z_STRVAL_P( conn->ce_option.ksp_name );
+    char* ksp_path = Z_STRVAL_P( conn->ce_option.ksp_path );
+    unsigned int name_len = Z_STRLEN_P( conn->ce_option.ksp_name );
+    unsigned int path_len = Z_STRLEN_P( conn->ce_option.ksp_path );
```
v-kaywon: path_len is not used

test/functional/sqlsrv/sqlsrv_connect_encrypted_ksp_errors.phpt
yitam: how else would you suggest I check the error messages?

source/shared/core_conn.cpp
david-puglielli: OK. Why use a C-style cast though? Wouldn't a static_cast be preferable?

.travis.yml
```diff
@@ -15,7 +15,9 @@ env:
   - PHPSQLDIR=/REPO/msphpsql-dev
   - TEST_PHP_SQL_SERVER=sql
   - SQLSRV_DBNAME=msphpsql_sqlsrv
-  - PDOSQLSRV_DBNAME=msphpsql_pdosqlsrv
+  - PDOSQLSRV_DBNAME=msphpsql_pdosqlsrv  
+  - TEST_PHP_SQL_UID=sa
+  - TEST_PHP_SQL_PWD=Password12@
```
Hadis-Fard: yml file should be synced with dev branch

Dockerfile-msphpsql
```diff
@@ -74,9 +79,6 @@ RUN sed -i -e 's/TARGET_DATABASE/msphpsql_sqlsrv/g' MsSetup.inc
 RUN sed -i -e 's/TARGET_USERNAME/'"$TEST_PHP_SQL_UID"'/g' MsSetup.inc
 RUN sed -i -e 's/TARGET_PASSWORD/'"$TEST_PHP_SQL_PWD"'/g' MsSetup.inc
 
-WORKDIR $PHPSQLDIR
-RUN chmod +x ./entrypoint.sh
-CMD /bin/bash ./entrypoint.sh
```
Hadis-Fard: the deleted lines are actually the latest dev branch, Docker file should be the latest on dev

test/functional/setup/ksp_app.c
v-kaywon: hmmm this may change very soon though once the ODBC driver is ready. We shall see

test/functional/sqlsrv/sqlsrv_connect_encrypted_ksp_errors.phpt
v-kaywon: print_r( $errors[0]['SQLSTATE'] );
print_r( $errors[0]['code'] );
print_r( $errors[0]['message'] );
or you can make a fancy foreach loop that asks if the key is a string then print the value.

test/functional/setup/ksp_app.c
yitam: No we will keep using a test table because we want to show that we can retrieve data encrypted by a 3rd party (here in this case it's odbc )

PR #512, opened 2017-08-23
Added SUSE 12 and CRUD to performance tests

david-puglielli:
Section for SUSE 12 added. A couple other changes as well.

### DIFF
--- a/test/Performance/benchmark/pdo_sqlsrv/PDOCRUDBench.php
+++ b/test/Performance/benchmark/pdo_sqlsrv/PDOCRUDBench.php
@@ -1,0 +1,79 @@
+<?php
+
+use PDOSqlsrvPerfTest\PDOSqlsrvUtil;
+include_once __DIR__ . "/../../lib/CRUDBaseBenchmark.php";
+/**
+ * @BeforeMethods({"connect", "setTableName", "createTable", "generateInsertValues", "generateUpdateValues", "generateUpdateParams"})
+ * @AfterMethods({ "dropTable", "disconnect"})
+ */
+class PDOSqlsrvCRUDBench extends CRUDBaseBenchmark 
+{
+
+    private $conn;
+    private $tableName;
+    private $insertValues;
+    private $updateValues;
+    private $updateParams;
+
+    public function setTableName()
+    {
+        $this->tableName = "datatypes_".rand();
+    }
+
+    public function connect()
+    {
+        $this->conn = PDOSqlsrvUtil::connect();
+    }
+
+    public function createTable()
+    {
+        PDOSqlsrvUtil::createCRUDTable( $this->conn, $this->tableName );
+    }
+
+    public function generateInsertValues()
+    {
+        $this->insertValues = PDOSqlsrvUtil::generateInsertValues();
+    }
+    
+    public function generateUpdateValues()
+    {
+        $this->updateValues = PDOSqlsrvUtil::generateUpdateValues();
+    }
+
+    public function generateUpdateParams()
+    {
+        $this->updateParams = PDOSqlsrvUtil::generateUpdateParams();
+    }
+    /**
+     * Each iteration does the following $loopsPerCRUDIter times:
+     *    (i) insert a row into the table with insertWithPrepare
+     *   (ii) fetch the row with fetchWithPrepare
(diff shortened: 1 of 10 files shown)


test/Performance/setup_env_unix.sh
```diff
@@ -158,12 +180,12 @@ cp php.ini-production php.ini
 driverName=$(basename $SQLSRV_DRIVER)
 echo "extension=$driverName" >> php.ini
 sudo cp -r $SQLSRV_DRIVER $phpExtDir/$driverName
-sudo chmod a+r $SQLSRV_DRIVER $phpExtDir/$driverName
+sudo chmod a+r $phpExtDir/$driverName
```
david-puglielli: chmod won't work if SQLSRV_DRIVER is located on a read-only filesystem, and it isn't even needed here once the drivers are copied to the PHP extensions directory. Same for PDO below.

test/Performance/setup_env_unix.sh
```diff
@@ -83,6 +83,28 @@ elif [ $PLATFORM = "RedHat7" ]; then
...
+    zypper -n ar https://packages.microsoft.com/config/sles/12/prod.repo 2>&1
+    zypper --gpg-auto-import-keys refresh 2>&1
+    ACCEPT_EULA=Y zypper -n install msodbcsql >> env_setup.log 2>&1
+    ACCEPT_EULA=Y zypper -n install mssql-tools >> env_setup.log 2>&1
+    zypper -n install unixODBC-devel >> env_setup.log 2>&1
+    printf "done\n"
+    
+    printf "Installing pyodbc"
+    pip3 install --upgrade pip >> env_setup.log 2>&1
+    pip3 install pyodbc >> env_setup.log 2>&1
```
v-kaywon: do we really need the stderr for all installation steps?

david-puglielli: For full logging, I think it's reasonable. Do you mean for the stderr output to go directly to the console?

david-puglielli:
Already tried it on Suse 12 and it seems to work, except for the Large tests, which I will look into.

test/Performance/setup_env_unix.sh
v-kaywon: Just that we didn't do so for the other platforms. You can leave it if you think it's reasonable.

v-kaywon:
@david-puglielli , which database are you using for Large tests. The large test assumes a database setup, so if you're not using one of the performance SQL servers, it will return an error. Also, if you're using a Linux SQL Server, there seem to be a difference in encoding for PDO_SQLSRV large test and the database, which results in an error. If you find out what's wrong, let me know too.

test/Performance/setup_env_unix.sh
david-puglielli: Hmm, it's done in a few places kind of haphazardly. I'll make it all consistent.

test/Performance/benchmark/pdo_sqlsrv/PDOCRUDBench.php
```diff
@@ -0,0 +1,79 @@
...
+     * Every update calls prepare, bindParam and execute APIs.
+     * Every delete calls prepare and execute APIs.
+     */    
+    public function benchCRUDWithPrepare()
+    {
+        for( $i=0; $i<PDOSqlsrvUtil::$loopsPerCRUDIter; $i++ )
+        {
+            PDOSqlsrvUtil::insertWithPrepare( $this->conn, $this->tableName, $this->insertValues );
+            PDOSqlsrvUtil::fetchWithPrepare( $this->conn, $this->tableName );
+            $stmt = PDOSqlsrvUtil::updateWithPrepare( $this->conn, $this->tableName, $this->updateValues, $this->updateParams );
```
v-kaywon: why need $stmt here? same with the SQLSRVCRUDBench test. Otherwise, this PR is approved.

david-puglielli: PDOUpdateBench and SqlsrvUpdateBench do the same, but updateWithPrepare() does not return anything. I will fix the UpdateBench tests too.

v-kaywon:
@david-puglielli merge it?

PR #514, opened 2017-09-05
add code for AE insertion

v-kaywon:


source/shared/core_sqlsrv.h
```diff
@@ -172,6 +172,7 @@ const int SQL_SERVER_MAX_FIELD_SIZE = 8000;
 const int SQL_SERVER_MAX_PRECISION = 38;
 const int SQL_SERVER_MAX_TYPE_SIZE = 0;
 const int SQL_SERVER_MAX_PARAMS = 2100;
+const int SQL_MAX_ERROR_MESSAGE_LENGTH = SQL_MAX_MESSAGE_LENGTH * 4;
```
yitam: What's SQL_MAX_MESSAGE_LENGTH? Why multiply it by 4? Perhaps add a simple comment to explain why this new constant is necessary

source/shared/core_sqlsrv.h
```diff
@@ -2184,6 +2196,17 @@ namespace core {
...
+    {
+        SQLRETURN r;
+        SQLHDESC hIpd = NULL;
+        core::SQLGetStmtAttr(stmt, SQL_ATTR_IMP_PARAM_DESC, &hIpd, 0, 0);
+        r = ::SQLSetDescField(hIpd, rec_num, fld_id, value_ptr, str_len);
+
+        CHECK_SQL_ERROR_OR_WARNING(r, stmt) {
+            throw CoreException();
+        }
+    }
```
yitam: It would be nice to add a simple comment what this function is for. Also, please adhere to the coding standard re the spacing

source/shared/core_stmt.cpp
```diff
@@ -436,16 +438,21 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
                   ( encoding == SQLSRV_ENCODING_SYSTEM || encoding == SQLSRV_ENCODING_UTF8 || 
                     encoding == SQLSRV_ENCODING_BINARY ), "core_sqlsrv_bind_param: invalid encoding" );
 
-    // if the sql type is unknown, then set the default based on the PHP type passed in
-    if( sql_type == SQL_UNKNOWN_TYPE ) {
-		default_sql_type( stmt, param_num, param_z, encoding, sql_type TSRMLS_CC );
+    if ( stmt->conn->ce_option.enabled && ( sql_type == SQL_UNKNOWN_TYPE || column_size == SQLSRV_UNKNOWN_SIZE ) &&
+         Z_TYPE_P( param_z ) != IS_NULL && Z_TYPE_P( param_z ) != IS_FALSE && Z_TYPE_P( param_z) != IS_TRUE ) {
+        ae_get_sql_type_info( stmt, param_num, direction, param_z, encoding, sql_type, column_size, decimal_digits TSRMLS_CC );
     }
```
yitam: Such a long list of conditions before calling ae_get_sql_type_info! So if param is null or boolean (true or false), it won't invoke ae_get_sql_type_info? Does it matter whether this is input / output / inout parameter?

source/shared/core_stmt.cpp
```diff
@@ -552,7 +559,7 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
                 // avoid this silent truncation, we set the column_size to be "MAX" size for 
                 // string types. This will guarantee that there is no silent truncation for 
                 // output parameters.
-                if( direction == SQL_PARAM_OUTPUT ) {
+                if( direction == SQL_PARAM_OUTPUT && !stmt->conn->ce_option.enabled ) {
```
yitam: please add a comment to explain why it's not necessary to set column_size here when column encryption is enabled

source/shared/core_stmt.cpp
```diff
@@ -2501,6 +2528,12 @@ void resize_output_buffer_if_necessary( _Inout_ sqlsrv_stmt* stmt, _Inout_ zval*
 
     // account for the NULL terminator returned by ODBC and needed by Zend to avoid a "String not null terminated" debug warning
     SQLULEN field_size = column_size;
+    // with AE on, when column_size is retrieved from SQLDescribeParam, column_size does not include the decimal place
+    // include the decimal for output params by adding elem_size
+    if ( stmt->conn->ce_option.enabled && decimal_digits != 0 )
+    {
```
yitam: is decimal_digits normally >= 0? so should it be > 0 here instead? Or even better, make this a boolean flag instead to add extra elem_size or not

source/shared/core_stmt.cpp
v-kaywon: Yes, if param is null, don't have to all SQLDescribedParam. It doesn't matter what direction it is.

source/shared/core_stmt.cpp
```diff
@@ -552,7 +559,8 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
                 // avoid this silent truncation, we set the column_size to be "MAX" size for 
                 // string types. This will guarantee that there is no silent truncation for 
                 // output parameters.
-                if( direction == SQL_PARAM_OUTPUT ) {
+                // if column encryption is enabled, do not set the length to unlimited since SQLDescribe paramter already derive the length for us
+                if( direction == SQL_PARAM_OUTPUT && !stmt->conn->ce_option.enabled ) {
```
yitam: Suggestion: modify the comment to "if column encryption is enabled, at this point the correct column size has been set by SQLDescribeParam()"

source/shared/core_stmt.cpp
```diff
@@ -2004,6 +2019,19 @@ void default_sql_size_and_scale( _Inout_ sqlsrv_stmt* stmt, _In_opt_ unsigned in
...
 }
 
+void ae_get_sql_type_info( _Inout_ sqlsrv_stmt* stmt, _In_opt_ SQLULEN paramno, _In_ SQLSMALLINT direction, _In_ zval* param_z, _In_ SQLSRV_ENCODING encoding,
+                           _Out_ SQLSMALLINT& sql_type, _Out_ SQLULEN& column_size, _Out_ SQLSMALLINT& decimal_digits TSRMLS_DC )
+{
+    SQLSMALLINT Nullable;
+    core::SQLDescribeParam( stmt, paramno + 1, &sql_type, &column_size, &decimal_digits, &Nullable );
+    // without AE, binding an INPUT bigint with sql type SQL_WVARCHAR returns a conversion error between nvarchar and bigint
+    // similate without AE by setting the sql type back to SQL_WVARCHAR using default_sql_type
+    if ( sql_type == SQL_BIGINT && Z_TYPE_P(param_z) == IS_STRING && direction == SQL_PARAM_INPUT ) {
```
yitam: I don't know how I missed this before. Anyway, the comment here is confusing... isn't the prerequisite of this method that AE is enabled? 
Also, why is `default_sql_size_and_scale` called to reset the column size? Besides, what if type of param_z is not a string?
Does this problem happen only with input param? Why not output or in/out param? 
Have you tested when encoding is SQLSRV_ENCODING_CHAR?

v-kaywon: Yes this function is only called when AE is enabled.

default_sql_size_and_scale is called to reset the column size because we want to mimic what happens without AE (i.e., as explained in the comment, without AE, there is an error in the conversion between SQL_WVARCHAR and SQL_BIGINT). If this block of code is not here, the will no longer be converting between bigint to wvarchar, instead it'll be converting between bigint and bigint (which is set by SQLDescribeParam). Everything would be fine at this point, except it will no longer mimic the behavior of without AE.

If the type of param_z is not a string, then we use the sql_type returned by SQLDescribeParam.

This is only for input params. I've tested without AE with output and in/out bigint, the conversion problem does not exist.

Yes, I've tested with SQLSRV_ENCODING_CHAR

Issue #507, opened 2017-08-18
SQLSTATE[IMSSP]: The active result for the query contains no fields.


dipens:
I have this very simple stored procedure execution on my server setup (SQL Server 2012 on a Windows Server connected to Ubuntu 16.04 running Symfony 3.3 on PHP 7.1 using these drivers ofcourse)

```
$connection = $em->getConnection()->getWrappedConnection();
$stmt = $connection->prepare('StoredProcedureName @a= :a, @b= :b, @c= :c');
$stmt->execute(array(
    'a'=> 'a',
    'b'=> 'b',
    'c'=> 'c',
));
$stmt->nextRowset();
$result = $stmt->fetchAll();
```
This works fine when there are records being returned. but sometimes depending on the parameters, the stored procedure returns no records. That's when I get this error:

> SQLSTATE[IMSSP]: The active result for the query contains no fields.

I have already looked into [here](https://stackoverflow.com/questions/16735898/the-active-result-contains-no-fields-using-pdo-with-ms-sql), [here](https://stackoverflow.com/questions/17626897/the-active-result-for-the-query-contains-no-fields), [here](https://stackoverflow.com/questions/37457065/not-receiving-result-from-stored-procedure-mssql-db-using-php), and in issues here too but couldn't find a solution. I have already set `SET NOCOUNT ON` in the beginning of the stored procedure.

Is there anything I am missing here or is this a bug?

david-puglielli:
Hi @dipens, could you provide a complete repro script for this error, the stored procedure you are using, and parameters that trigger the error? Thanks

dipens:
Sure @david-puglielli and I apologize for the delay in my reply. Here is the complete repro.


```
CREATE  PROCEDURE [dbo].StoredProcedureName @a int, @b nvarchar(50),@c nvarchar(50)
AS
SET NOCOUNT ON

DECLARE @d nvarchar(30), @f int, @e INT, @h INT
SELECT @d=d,@e=e, @f=f FROM tblA WHERE a=@a

If @d='d'
BEGIN

	DECLARE @g INT

	UPDATE tblA set 
	AStatus = 1,
	Where a = @a
		
	exec [dbo].AnotherStoredProcedure @a=@a

	exec [dbo].YetAnotherStoredProcedure @a=@a OUTPUT,@j='j'

	Update tblb set bStatus='b' where b=@b

	UpDate tblc set cStatus='c' WHERE c=@c

[...]

PR #520, opened 2017-09-07
Added empty result set check to nextRowset()

david-puglielli:
This is a fix for the bug discussed in issue #507.


source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -1067,6 +1067,14 @@ int pdo_sqlsrv_stmt_next_rowset( _Inout_ pdo_stmt_t *stmt TSRMLS_DC )
...
         SQLSRV_ASSERT( driver_stmt != NULL, "pdo_sqlsrv_stmt_next_rowset: driver_data object was null" );
 
+        // If SQLNumResultCols returns 0, the result set is empty. Normally this error is 
+        // handled in core_sqlsrv_fetch, but if the user calls nextRowset() before fetch()
+        // the error message won't show up unless we handle it here.
+        SQLSMALLINT has_fields = core::SQLNumResultCols( driver_stmt TSRMLS_CC );
+        CHECK_CUSTOM_ERROR( has_fields == 0, driver_stmt, SQLSRV_ERROR_NO_FIELDS ) {
+            throw core::CoreException();
+        }
+
```
yitam: From this [page](https://docs.microsoft.com/en-us/sql/relational-databases/native-client-odbc-api/sqlnumresultcols) it seems valid and accurate to call this without making another trip to the server. Please confirm.

david-puglielli: Since a statement has to be executed for the user to call nextRowset, there is no trip to the server made here.

david-puglielli: Yes, I've verified the same problem exists on the sqlsrv side. But I think core_sqlsrv_fetch is never called if nextRowset is called before fetch, which is why we're catching the error here. I'll add a fix and test for sqlsrv.

david-puglielli:
Moved the error check to core_sqlsrv_next_result in core_stmt.cpp.

source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -1067,6 +1067,30 @@ int pdo_sqlsrv_stmt_next_rowset( _Inout_ pdo_stmt_t *stmt TSRMLS_DC )
...
+        }
+        
+        // Now make sure the result set is not null. 
+        bool has_result = core_sqlsrv_has_any_result( driver_stmt );
+
+        if(!driver_stmt->fetch_called){
+            CHECK_CUSTOM_ERROR( !has_result, driver_stmt, SQLSRV_ERROR_NO_FIELDS ) {
+                throw core::CoreException();
+            }
+        }
```
v-kaywon: I don't quite understand. I thought PSOStatement::nextRowset should only be called on a fetched statement. The condition for if says fetched is NOT called, and throws an exception if there is no result. How can there any result if the statement was never fetched?

test/functional/pdo_sqlsrv/pdo_empty_result_error.phpt
```diff
@@ -0,0 +1,96 @@
...
+--SKIPIF--
+<?php require('skipif.inc'); ?>
+--FILE--
+<?php
+require_once("MsSetup.inc");
+
+$conn = new PDO( "sqlsrv:Server = $server; Database = $databaseName; ", $uid, $pwd );
+$conn->setAttribute( PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION );
+
+$stmt = $conn->query("IF OBJECT_ID('TestEmptySetTable', 'U') IS NOT NULL DROP TABLE TestEmptySetTable");
```
v-kaywon: There a DropTable and DropProc function you can use in MsCommon

source/pdo_sqlsrv/pdo_stmt.cpp
david-puglielli: nextRowset() *should* only be called on a fetched statement, but there is nothing stopping the user from calling nextRowset immediately after executing, and we need to handle that case. In issue 507 it occurs when going into the 'else' part of stored procedure that returns a result set in the 'if' but not the 'else' branch, so it can happen that user does call nextRowset first accidentally via a stored procedure. 

This code catches the case where nextRowset() is called before fetch on a null result set, giving the appropriate error message; before, there would be no error given. In the case where nextRowset is called before fetch on non-null result set, core_sqlsrv_has_any_result will return true and the error is not triggered. In the event that nextRowset is called after fetch() and the result set is null, that will be handled in core_sqlsrv_next_result() (or by the check for past_next_result_end above). I think that covers all the relevant cases...

v-kaywon: yea I was confused what happens when fetch isn't called and has_result is true, but now it's more clear. Can you put a comment saying if fetch_called is false and has_result is true, core_sqlsrv_next_result will handle it.

test/functional/pdo_sqlsrv/pdo_empty_result_error.phpt
```diff
@@ -0,0 +1,98 @@
+--TEST--
+Test that calling nextRowset() on an empty result set produces the correct error message. Fix for Github 507.
+--SKIPIF--
+<?php require('skipif.inc'); ?>
```
yitam: Please try to limit the length of test title. If you have more please put it in --DESCRIPTION-- instead.
For tests, it's a good practice to drop test tables / procedures at the end to leave no trace.

test/functional/sqlsrv/sqlsrv_empty_result_error.phpt
```diff
@@ -0,0 +1,121 @@
...
+
+$stmt = sqlsrv_query($conn, "TestEmptySetProc @a='a', @b='c'");
+$result = sqlsrv_fetch_array($stmt);
+print_r($result);
+print_r(sqlsrv_errors());
+
+$stmt = sqlsrv_query($conn, "DROP TABLE TestEmptySetTable");
+$stmt = sqlsrv_query($conn, "DROP PROCEDURE TestEmptySetProc");
+sqlsrv_close($conn);
+?>
```
yitam: just for practice, probably free the statement object too before closing the connection?

test/functional/sqlsrv/sqlsrv_empty_result_error.phpt
```diff
@@ -0,0 +1,121 @@
...
+--EXPECTF--
+Return a nonempty result set:
```
yitam: why EXPECTF?

test/functional/pdo_sqlsrv/pdo_empty_result_error.phpt
david-puglielli: I do drop them at the end, but I kept the drop functions at the beginning too just in case...

test/functional/sqlsrv/sqlsrv_empty_result_error.phpt
david-puglielli: %s in the output may be either 'Microsoft' on Windows or 'unixODBC' on unix systems, and EXPECTREGEX seemed like overkill for one string in a long output like this.

david-puglielli:
I'm not sure how useful testing with a regular SELECT statement would be, because I don't think the result set could ever be null in that case. At most you can have an empty result set with no rows but at least one column, or you select from a nonexistent column, which yields a different error message. These are not the same as a null result set, in which there are no rows and no columns, and it's why I needed to use the stored procedure in the tests to trigger a null result.

test/functional/pdo_sqlsrv/pdo_empty_result_error.phpt
yitam: ahhh... I see! In the beginning you use DropTable and DropProc but later you simply use queries.

PR #524, opened 2017-09-13
Bvt test fix

david-puglielli:


test/bvt/sqlsrv/msdn_sqlsrv_cancel.phpt
```diff
@@ -45,5 +45,5 @@ echo "$count sales accounted for the first $$salesTotal in revenue.\n";
 /* Cancel the pending results. The statement can be reused. */
 sqlsrv_cancel( $stmt);
 ?>
---EXPECT--
-57 sales accounted for the first $104171.7607 in revenue.
+--EXPECTREGEX--
+[1-9][0-9] sales accounted for the first $10[0-9]{4}\.[0-9]{2,4} in revenue.
```
yitam: Perhaps restrict the first regex pattern to 5[0-9]? Often times I see 50+ (i.e. 57 or 58, something like that)

david-puglielli: I have seen it go down to 30 sales or so, so I think we should keep it flexible.

yitam: in this case I'd suggest shrinking the range to [3-5] instead of [1-9]

PR #530, opened 2017-09-14
add insertion tests for Always Encrypted; add common functions in MsC…

v-kaywon:
add insertion tests for Always Encrypted
add common functions in MsCommon.inc for refactoring in the future

### DIFF
--- a/test/functional/pdo_sqlsrv/MsCommon.inc
+++ b/test/functional/pdo_sqlsrv/MsCommon.inc
@@ -442,7 +442,7 @@ function is_col_enc()
 {
     require  'MsSetup.inc';
     if ( $keystore == "none" )
-        return false
+        return false;
     else
         return true;
 }
--- a/test/functional/pdo_sqlsrv/MsCommon.inc
+++ b/test/functional/pdo_sqlsrv/MsCommon.inc
@@ -20,11 +20,12 @@ function IsAEQualified($conn)
     $server_ver = $conn->getAttribute(PDO::ATTR_SERVER_VERSION);
     $msodbcsql_maj = explode(".", $msodbcsql_ver)[0];
     $msodbcsql_min = explode(".", $msodbcsql_ver)[1];
-    if ($msodbcsql_maj < 17 || explode('.', $server_ver)[0] < 13)
+    if ($msodbcsql_maj < 13 || explode('.', $server_ver)[0] < 13)
         return false;
     return true;
 }
 
+// TO BE DELETED
 function connect($options=array())
 {
   try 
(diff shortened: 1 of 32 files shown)


test/functional/pdo_sqlsrv/MsCommon.inc
```diff
@@ -52,6 +53,374 @@ function connect($options=array())
...
+}
+
+
+/**
+ * Create a row into a table
+ * @param object $conn : PDO connection object
+ * @param string $tbname : name of the table to be created
+ * @param array $inputs : an associative array column name and its value
+ * @param bool $r : if the row was successfully inserted
+ * $param string $api : PDO_SQLSRV API used for executing the insert query
```
yitam: Perhaps a more description variable for `$r`, and why is the default value for a boolean flag `null`?

Also, why return `$r` instead of throwing an error right when something wrong happens?

test/functional/pdo_sqlsrv/MsCommon.inc
```diff
@@ -52,6 +53,374 @@ function connect($options=array())
...
+}
+
+
+/**
+ * Fetch all rows and all columns given a table name, and print them
+ * @param object $conn : PDO connection object
+ * @param string $tbname : name of the table to fetch from
+ */
+function fetch_all( $conn, $tbname ) {
+    try
```
yitam: this function and others should check if `$conn` is null...

test/functional/pdo_sqlsrv/MsCommon.inc
```diff
@@ -52,6 +53,374 @@ function connect($options=array())
...
+    {
+        $sql = "SELECT * FROM $tbname";
+        $stmt = $conn->query( $sql );
+        while ( $row = $stmt->fetch( PDO::FETCH_ASSOC ))
+        {
```
yitam: perhaps the fetch type is not hardcoded but defaulted to FETCH_ASSOC?

test/functional/pdo_sqlsrv/MsCommon.inc
```diff
@@ -52,6 +53,374 @@ function connect($options=array())
...
+ */
+function get_insertSql_complete( $tbname, $inputs )
+{
+    $colStr = "INSERT INTO $tbname (";
+    $valStr = "VALUES (";
+    foreach( $inputs as $key => $value )
+    {
+        $colStr .= $key . ", ";
+        if ( is_string( $value ))
+            $valStr .= "'" . $value . "', ";
```
yitam: what is `$value` is null?

test/functional/pdo_sqlsrv/MsCommon.inc
```diff
@@ -52,6 +53,374 @@ function connect($options=array())
...
+ * @param string $tbname : name of the table to be created
+ * @param array $inputs : an associative array column name and its value
+ * @param bool $r : if the row was successfully inserted
+ * $param string $api : PDO_SQLSRV API used for executing the insert query
+ *                      accepted values: "exec", "query", "prepare", "prepareBindParam"
+ * @param array $paramOption : an array of bindParamOptions. Should only be provided if $api is "perpareWithParamOp", if provided, has to be in the same order as $inputs
+ * @return object PDOStatement object of the insert statement
+ */
+function insert_row( $conn, $tbname, $inputs, &$r = null, $api = "query", $paramOption = array() )
+{
```
yitam: This insert with params is somewhat error prone (say the order of $inputs is different from $paramOption, etc.) Why can't we add the logic to `bindParamOptions` such that it will simply invoke `bindParam()` when defaults are used?

yitam: Suggestion: split this `insert_row()` into two methods, one for _exec/query_ (using a boolean flag to distinguish the two, and this returns an error when `$keystore` is NOT "none") and the other for _prepare_, which takes the extra arguments plus checking if `$paramOption` is empty, etc.

test/functional/pdo_sqlsrv/MsCommon.inc
```diff
@@ -20,11 +20,12 @@ function IsAEQualified($conn)
     $server_ver = $conn->getAttribute(PDO::ATTR_SERVER_VERSION);
     $msodbcsql_maj = explode(".", $msodbcsql_ver)[0];
     $msodbcsql_min = explode(".", $msodbcsql_ver)[1];
-    if ($msodbcsql_maj < 17 || explode('.', $server_ver)[0] < 13)
+    if ($msodbcsql_maj < 13 || explode('.', $server_ver)[0] < 13)
         return false;
```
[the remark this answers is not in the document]
v-kaywon: I shall change it back

test/functional/pdo_sqlsrv/MsCommon.inc
```diff
@@ -52,6 +53,374 @@ function connect($options=array())
...
+            $encDet = "ENCRYPTED WITH (ENCRYPTION_TYPE = DETERMINISTIC, ALGORITHM = 'AEAD_AES_256_CBC_HMAC_SHA_256', COLUMN_ENCRYPTION_KEY = $cekName) ";
+            $encRand = "ENCRYPTED WITH (ENCRYPTION_TYPE = RANDOMIZED, ALGORITHM = 'AEAD_AES_256_CBC_HMAC_SHA_256', COLUMN_ENCRYPTION_KEY = $cekName) ";
+            if ( stripos( $this->dataType, "char" ) !== false )
+                $append .= "COLLATE Latin1_General_BIN2 ";
+            if ( $this->encType == "deterministic" )
+                $append .= $encDet;
+            else
+                $append .= $encRand;
+        }
+        $append .= $this->options;
```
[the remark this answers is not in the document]
v-kaywon: I just find making it deterministic and randomized a lot clearer than $randomized = true/false. Besides, there's already a default value so the user doesn't have to provide an encType when creating a columnMeta unless he/she really wants a randomized column.

I've changed getColDef to use sprintf

test/functional/pdo_sqlsrv/MsCommon.inc
v-kaywon: Return $r so user can handle the error / continue execution.

test/functional/sqlsrv/sqlsrv_ae_insert_datetime.phpt
```diff
@@ -0,0 +1,112 @@
...
+$dataTypes = array( "date", "datetime", "datetime2", "smalldatetime", "time", "datetimeoffset" );
+$conn = ae_connect();
+
+foreach ( $dataTypes as $dataType ) {
+    echo "\nTesting $dataType: \n";
+    
+    // create table
+    $tbname = GetTempTableName( "", false );
+    $colMetaArr = array( new columnMeta( $dataType, "c_det" ), new columnMeta( $dataType, "c_rand" ));
+    create_table( $conn, $tbname, $colMetaArr );
```
yitam: I believe `c_rand` is supposed to be randomized?

test/functional/pdo_sqlsrv/MsCommon.inc
v-kaywon: why do I need to? the function will throw an exception if $conn is not a PDO object

test/functional/pdo_sqlsrv/pdo_ae_insert_retrieve.phpt
```diff
@@ -0,0 +1,85 @@
...
+<?php
+include 'MsSetup.inc';
+include 'MsCommon.inc';
+include 'AEData.inc';
+
+try
+{
+    $conn = ae_connect();
+    $conn->setAttribute( PDO::ATTR_ERRMODE, PDO::ERRMODE_SILENT );
+
```
yitam: why errmode silent?

test/functional/pdo_sqlsrv/MsCommon.inc
v-kaywon: This function prints the result set and the user does not get the result set back. I don't think a user would want to see a numeric key being printed ( e.g., "1: blah" ) since it's very uninformative.

test/functional/pdo_sqlsrv/MsCommon.inc
v-kaywon: Can you elaborate on "Why can't we add the logic to bindParamOptions such that it will simply invoke bindParam() when defaults are used"? To prevent ordering problem between $inputs and $paramOptions, I've already added an $index member in bindParamOption. If the user does not put $inputs in order, the insert sql itself will adapt to that ordering; $index in bindParamOption matches with the $index'th "?" placeholder in the insertSql.

v-kaywon: For your suggestion, doesn't that mean the test will need to be different between CE enabled / not enabled? For example, for CE not enabled, the test wants to call insert_row( "query" ). If the same test is used for CE enabled, then the test will return an error, and for it to work it has to use insert_row( "prepareWithParamOp" ).

test/functional/sqlsrv/sqlsrv_ae_insert_datetime.phpt
v-kaywon: Yes my mistake, I'll change them

test/functional/pdo_sqlsrv/pdo_ae_insert_retrieve.phpt
v-kaywon: Some tests expects an error and I want those tests to continue executing. This test however shouldn't expect an error. I'll change it.

test/functional/pdo_sqlsrv/pdo_ae_insert_pdoparam_money.phpt
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    echo "$dataType is not compatible with any type.\n";
+                    $success = false;
+                }
+            }
+            $conn->query( "TRUNCATE TABLE $tbname" );
+        }
+        if ( $success )
+            echo "Test successfully.\n";
+        DropTable( $conn, $tbname );
```
yitam: Suggestion: either `Test is successful` or `Test successfully done`

test/functional/pdo_sqlsrv/pdo_ae_insert_pdoparam_numeric.phpt
```diff
@@ -0,0 +1,189 @@
+--TEST--
+Test for inserting and retrieving encrypted numeric types data
+Use PDOstatement::bindParam with all PDO::PARAM_ types
```
yitam: data of numeric types, likewise for the other tests

test/functional/pdo_sqlsrv/pdo_ae_insert_pdoparam_string.phpt
```diff
@@ -0,0 +1,121 @@
...
+Test for inserting and retrieving encrypted string types data
+Use PDOstatement::bindParam with all PDO::PARAM_ types
+--SKIPIF--
+<?php require('skipif_versions_old.inc'); ?>
+--FILE--
+<?php
+include 'MsCommon.inc';
+include 'AEData.inc';
+include 'MsSetup.inc';
+
```
yitam: Is `MsSetup.inc` really required in this test? Please check the other tests too

test/functional/pdo_sqlsrv/pdo_ae_output_param_all.phpt
```diff
@@ -0,0 +1,167 @@
+--TEST--
+Test for binding output params for encrypted data for all types
+--SKIPIF--
+<?php require('skipif_versions_old.inc'); ?>
+--FILE--
+<?php
+include 'MsCommon.inc';
+include 'MsSetup.inc';
+include 'AEData.inc';
+
```
yitam: Is `AEData.inc` required? Check the other tests too

test/functional/sqlsrv/MsCommon.inc
```diff
@@ -170,6 +170,381 @@ function Connect($options = array())
...
+    $colStr = rtrim( $colStr, ", " ) . ") ";
+    $valStr = rtrim( $valStr, ", " ) . ") ";
+    $insertSql = $colStr . $valStr;
+    return $insertSql;
+}
+
+
+/**
+ * @param string $tbname : name of the table for an insert sql
+ * @param array $input : associative array containing a key value pair of column name and data to put into an insert sql string
```
yitam: typo: should be `$inputs`, not `$input`

test/functional/pdo_sqlsrv/MsCommon.inc
```diff
@@ -52,6 +53,374 @@ function connect($options=array())
...
+        {
+            $this->colName = $colName;
+        }
+        $this->dataType = $dataType;
+        $this->encType = $encType;
+        $this->options = $options;
+    }
+    /**
+     * @return string column definition for creating a table
+     */
```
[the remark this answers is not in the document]
v-kaywon: already did

test/functional/sqlsrv/MsCommon.inc
```diff
@@ -170,6 +170,381 @@ function Connect($options = array())
...
+ * @return string an insert sql string with "?" placeholders for all values
+ */
+function get_insertSql_placeholders( $tbname, $inputs )
+{
```
yitam: should you check if `$inputs` is empty?

test/functional/sqlsrv/MsCommon.inc
```diff
@@ -170,6 +170,381 @@ function Connect($options = array())
...
+
+
+/**
+ * @param int $num : number of placeholders needed
+ * @return string a string containing $num number of repeated "?" placeholders delimited by ", "
+ */
+function get_seq_placeholders( $num )
+{
+    $placeholderStr = str_repeat( "?, ", $num );
+    $placeholderStr = rtrim( $placeholderStr, ", " );
```
yitam: check if `$num` is positive (> 0)

test/functional/sqlsrv/MsCommon.inc
```diff
@@ -170,6 +170,381 @@ function Connect($options = array())
...
+    while ( $row = sqlsrv_fetch_array( $stmt, SQLSRV_FETCH_ASSOC ))
+    {
+        foreach ( $row as $key => $value )
+        {
+            if ( is_object( $value ) )
+            {
+                print "$key:\n";
+                foreach ( $value as $k =>$value )
+                {
+                    print ( "  $k: $value\n" );
```
yitam: probably not a good idea to reuse `$value` here ?

test/functional/sqlsrv/sqlsrv_ae_insert_money.phpt
```diff
@@ -0,0 +1,77 @@
...
+    $colMetaArr = array( new columnMeta( $dataType, "c_det" ), new columnMeta( $dataType, "c_rand" ));
+    create_table( $conn, $tbname, $colMetaArr );
+    
+    // insert a row
+    $inputValues = array_slice( ${explode( "(", $dataType )[0] . "_params"}, 1, 2 );
+    $r;
+    $stmt = insert_row( $conn, $tbname, array( $colMetaArr[0]->colName => $inputValues[0], $colMetaArr[1]->colName => $inputValues[1] ), $r );
+    
+    if ( $keystore == "none" )
+    {
```
yitam: Perhaps add a function to MsCommon.inc to return a boolean that indicates if AE is on (i.e. if $keystore is "none") then in this case you don't really need to include 'MsSetup.inc' everywhere

test/functional/pdo_sqlsrv/MsCommon.inc
yitam: For your current test it makes sense to use FETCH_ASSOC but since this method is a COMMON method for other tests to use, it will be nice to make it more flexible

test/functional/pdo_sqlsrv/MsCommon.inc
yitam: ok... it's just a common practice to check the inputs

test/functional/pdo_sqlsrv/MsCommon.inc
yitam: Since you're creating a class bindParamOptions anyway, and it has one function that invokes bindParam() with extra options, it will be nice to add a simple method in the same class that calls bindParam() with defaults. As a result, you don't need two input arguments in which you need to compare its ordering, etc.

yitam: About insert_row(), based on what you said, why does the user need to specify if it's exec, query, or prepare, etc.?

v-kaywon: bindParamOption is only used when user wants to specify parameter options ( such as PDO::PARAM_ type, length and encoding ). If the user doesn't want to provide any of these options he/she can simply invoke PDOStatement;:bindParam. Thus I don't see why we need another function that called bindParam() with defaults. As for the "two input arguments", do you mean $stmt and $var? How else can I call bindParam() without these two inputs?

I made a parameter to contain the api being used to call the insert sql because I wanted more flexibility for this function (in fact I think we should include this parameter for all common functions that executes any query). A test can insert using exec, query, or prepare based on the user's preference and what he/she want to test. And as we both know, CE does not work without binding parameters, thus when CE is on, the function automatically chooses to bind param with prepare and execute. This way, the user can use this function to test exec, query or prepare on the non-CE side and the test won't fail when CE is being tested.

test/functional/pdo_sqlsrv/MsCommon.inc
v-kaywon: The function doesn't return the result set though, they just get printed.. most user prefer numeric keys for convenience, but in this case the user doesn't have access to the result set being fetched. I still don't see when a user would prefer to see numeric keys being printed.

A common function can be created in the future that returns the result set with optional fetch types.

test/functional/pdo_sqlsrv/MsCommon.inc
yitam: I meant the two input arguments to insert_row(), `$inputs` and `$paramOption`, which can be combined, changing the latter class to also take an input value of any type, thus eliminating the unnecessary logic to match each input value to its param options

We can therefore rename class bindParamOption as inputParam to make it more generic, having a method called bindParam() which takes care of invoking stmt->bindParam according to the param options

About splitting insert_row() into two methods, I suggested that mainly because "prepare" (with or without options) is quite different from exec/query. In the case when we want transparency, the user can still use insert_row_query but if AE is enabled, we can invoke insert_row_prepare behind the scene

Let's discuss

PR #533, opened 2017-09-15
ODBC 17 binaries

david-puglielli:
Added x86 and x64 Windows binaries.


Hadis-Fard:
Please add a file stating this is preview and should not be used in production

PR #534, opened 2017-09-15
add error handling for using sqlsrv_query without sqltype

v-kaywon:
add error handling for using sqlsrv_query without sqltype in a column encryption enabled connection


source/sqlsrv/stmt.cpp
```diff
@@ -1974,9 +1977,11 @@ void parse_param_array( _Inout_ ss_sqlsrv_stmt* stmt, _Inout_ zval* param_array,
 
         sql_type = sqlsrv_sql_type.typeinfo.type;
     }
-    // else the sql type and size are uknown, so tell the core layer to use its defaults
+    // else the sql type and size are unknown, so tell the core layer to use its defaults
     else {
-
+        CHECK_CUSTOM_ERROR( !stmt->prepared && stmt->conn->ce_option.enabled, stmt, SS_SQLSRV_ERROR_AE_QUERY_SQLTYPE_REQUIRED ) {
+            throw ss::SSException();
+        }
```
yitam: So only throw this error when the user is not using prepare() for AE?
What if the user uses the param array but does not specify SQL type?

test/functional/sqlsrv/sqlsrv_encrypted_query_nosqltype.phpt
```diff
@@ -0,0 +1,138 @@
+--TEST--
+Test using sqlserv_query for binding parameters with ColumnEncryption enabled and a custome keystore provider
+--SKIPIF--
+<?php require('skipif_server_old.inc'); ?>
```
yitam: if using a custom keystore provider, you will need to the other skipif specific for ksp
also, please fix the typo _custome_

source/sqlsrv/util.cpp
```diff
@@ -370,6 +370,10 @@ ss_error SS_ERRORS[] = {
 		SS_SQLSRV_ERROR_INVALID_AUTHENTICATION_OPTION,
 		{ IMSSP, (SQLCHAR*)"Invalid option for the Authentication keyword. Only SqlPassword or ActiveDirectoryPassword is supported.", -62, false }
 	},
+    {
+        SS_SQLSRV_ERROR_AE_QUERY_SQLTYPE_REQUIRED,
+        { IMSSP, (SQLCHAR*)"Using sqlsrv_query for binding parameters in a Column Encryption enabled connection requires the sqltype to be provided.", -63, false }
+    },
```
yitam: Suggestion: consider rephrasing the error message as "Must specify the SQL type for each parameter when using sqlsrv_query() with the Always Encrypted feature."

source/sqlsrv/stmt.cpp
v-kaywon: yes, when user uses prepare() there is no problem since SQLDescribeParam is called after SQLPrepare. When query is used however, SQLDescribeParam is call BEFORE SQLExecDirect, resulting a function sequence error.

The part that user uses a param array but does not specify SQL type is handled in line 1225 to 1227.

PR #544, opened 2017-09-19
fix handling decimal type with no decimal with always encrypted

v-kaywon:


test/functional/sqlsrv/sqlsrv_ae_insert_sqltype_datetime.phpt
```diff
@@ -48,7 +49,7 @@ foreach ( $dataTypes as $dataType ) {
                     if ( $compatType == $sqlType )
                         $isCompat = true;
                 }
-                if ( $isCompat )
+                if ( $isCompat && sqlsrv_errors()[0]['SQLSTATE'] == 22018 )
```
david-puglielli: Under what conditions would this error occur, if the data types are supposedly compatible?

v-kaywon: I've added a comment on this in the test for clarification. This error occurs when the datatypes between sqltype and column datatype is not compatible (i.e., implicit conversion is not allowed in SQL Server)

PR #546, opened 2017-09-20
fixed libc6 and libstdc++6 dependencies

lilgreenbird:


david-puglielli:
This takes care of the dependency issues, but when running sqlcmd I get 
`terminate called after throwing instance of 'std::runtime_eror'`
`  what(): locale::facet::_S_create_c_locale name not valid`

PR #549, opened 2017-09-21
Added install requirements to readme

david-puglielli:


ODBC 17 binaries preview/README.md
```diff
@@ -1,3 +1,6 @@
-The ODBC driver 17 preview binaries in this directory are required in order to use Always Encrypted (AE) functionality. Please note that these drivers should be considered to be preview versions -- they should not be used in production and are not supported by Microsoft. They will be replaced upon the official release of ODBC driver 17.
-
-To install on Windows, simply run the msi. For instructions on installing the binaries on Linux platforms, please see [this page](https://github.com/Microsoft/msphpsql/wiki/Install-and-configuration#odbc-17-linux-installation).
+
+The ODBC driver 17 preview binaries in this directory are required in order to use Always Encrypted (AE) functionality. Please note that these drivers should be considered to be preview versions -- they should not be used in production and are not supported by Microsoft. They will be replaced upon the official release of ODBC driver 17.
```
lilgreenbird: it is confusing but I don't think they refer to Always Encrypted as AE anymore. The feature is still Always Encrypted but it's it's CE (it's a product mgmt thing). We went through a whole round of changing all reference of AE in the code to CE (you see internally it's SQL_CE_ENABLED). I think it's best to remove the "(AE)" to avoid confusion.

david-puglielli: Ok I'll remove the AE. Though I was asked a few days ago to change column encryption to always encrypted on the wiki...

Issue #538, opened 2017-09-18
PDO::quote() for SQL Server mis-quotes strings containing ASCII NUL


MarkMaldaba:
I asked a question on StackOverflow and from the various responses it appears that perhaps this is a bug in the pdo_sqlsrv driver itself.  I am therefore re-posting it here so that it can be investigated properly.

Here is the same question as posted on StackOverflow: https://stackoverflow.com/questions/46100518

-----

I'm trying to insert ASCII NUL characters (`\0` aka `U+0000`) into an SQL Server database from PHP, using pdo_sqlsrv.  This is a requirement for handling PHP serialized strings, which contain NUL characters to represent private/protected variables.

However, there is something about PDO::quote() which is breaking the strings.

Code to reproduce (replace `DBNAME`, `USERNAME` and `PASSWORD` with appropriate values):

````php
<?php

try {
	$dsn = 'sqlsrv:Server=.\SQLEXPRESS;Database=DBNAME';
	$user = 'USERNAME';
	$pass = 'PASSWORD';

	$connection = new PDO($dsn, $user, $pass);
} catch (PDOException $e) {
	die("Connection error: " . $e->getMessage());
}

$str = "XX\0XX";

header("Content-Type: text/plain");

print("Original: " . str_replace("\0", "{NUL}", $str) . "\n");
$str = $connection->quote($str);
print("Quoted:   " . str_replace("\0", "{NUL}", $str) . "\n");

?>
````

**Expected Output:**

````
Original: XX{NUL}XX
Quoted:   'XX{NUL}XX'
````

**Actual Output:**

````
Original: XX{NUL}XX
Quoted:   'XX'{NUL}{NUL}a
````

Note that it appears that the final character is random, as on a subsequent run it was an `e`.  This implies some form of memory access error, e.g. reading past the end of the string.  Maybe a bug in the pdo_sqlsrv implementation?

**Comment by MarkMaldaba** [comment]
PS - The issue was experienced using ````php_pdo_sqlsrv_56_nts.dll```` (v3.2) on PHP 5.6.18.

**Comment by v-kaywon** [comment]
Hi @MarkMaldaba , we do not actively support PHP 5.6 anymore. Please get the latest PDO_SQLSRV driver from (here)[https://github.com/Microsoft/msphpsql/releases/tag/v5.0.0-preview]. From your repo script and from the latest driver, the output that I get is:
```
'XX'{NUL}{NUL}{NUL}
```
[...]

PR #550, opened 2017-09-22
Fix PDO::quote with string containing ASCII NUL character

v-kaywon:
Fix for https://github.com/Microsoft/msphpsql/issues/538


source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -1463,7 +1463,7 @@ int pdo_sqlsrv_dbh_quote( _Inout_ pdo_dbh_t* dbh, _In_reads_(unquoted_len) const
         if ( encoding == SQLSRV_ENCODING_UTF8 ) {
             quotes_needed = 3;
         }
-        for ( size_t index = 0; index < unquoted_len && unquoted[ index ] != '\0'; ++index ) {
+        for ( size_t index = 0; index < unquoted_len; ++index ) {
```
yitam: Is the argument `unquoted_len` always correct? What is `unquoted_len` if the string is empty? Or what if the input string is composed of only one null character, something like `'\0'`, or a series of null characters, something like `'\0\0\0\0'`?

test/functional/pdo_sqlsrv/pdo_538_quote_nul.phpt
```diff
@@ -0,0 +1,29 @@
...
+--FILE--
+<?php
+  
+require_once 'MsCommon.inc';
+
+try 
+{      
+    $connection = connect(); 
+    //$connection->setAttribute( PDO::SQLSRV_ATTR_DIRECT_QUERY, PDO::SQLSRV_ENCODING_SYSTEM );
+
```
yitam: Why did you comment this line? Have you tested this with PDO::SQLSRV_ENCODING_SYSTEM in Windows?

v-kaywon: I tested both ENCODING_UTF8 and ENCODING_SYSTEM and they are both OK no. I'll delete this line from the test.

source/pdo_sqlsrv/pdo_dbh.cpp
v-kaywon: From what I've tested, unquote_len seems to be always correct ( it was parsed out by the PDO extension using zend_parse_parameter. If the string is empty (""), unquoted_len is 0. If the input string is composed of only one null character, unquoted_len is 1, and for '\0\0\0\0' unquoted_len is 4.

test/functional/pdo_sqlsrv/pdo_538_quote_nul.phpt
```diff
@@ -0,0 +1,28 @@
...
+}
+
+catch( PDOException $e ) {
+    die("Connection error: " . $e->getMessage());
+}
+?> 
+
+--EXPECT--
+Original: XX{NUL}XX
+Quoted:   'XX{NUL}XX'
```
lilgreenbird: what is this character at the EOF?

v-kaywon: no newline

PR #552, opened 2017-09-22
Search for installed ODBC driver(s)

yitam:


source/shared/core_conn.cpp
```diff
@@ -150,21 +150,55 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+    bool is_missing_driver = true;
 
     if ( conn->is_driver_set ) {
         r = core_odbc_connect( conn, conn_str, is_missing_driver, is_pooled );
     }
     else if ( conn->ce_option.enabled ) {
+#ifndef _WIN32
+        // In non-Windows environment, because different unixODBC versions return different error states,
+        // if it fails to connect it is unreliable to check for a certain sql state error. Thus, 
+        // we will ignore the returned value of 'is_missing_driver' and simply throw an exception.
```
lilgreenbird: it would be cleaner to put this logic inside core_odbc_connect so it will set is_missing_driver correct then we shouldn't need to ignore it.

source/shared/core_conn.cpp
```diff
@@ -150,21 +150,55 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
     build_connection_string_and_set_conn_attr( conn, server, uid, pwd, options_ht, valid_conn_opts, driver, conn_str TSRMLS_CC );
     
-    bool is_missing_driver = false;
+    bool is_missing_driver = true;
 
     if ( conn->is_driver_set ) {
         r = core_odbc_connect( conn, conn_str, is_missing_driver, is_pooled );
     }
     else if ( conn->ce_option.enabled ) {
+#ifndef _WIN32
```
lilgreenbird: can we restructure this so it's a bit cleaner ie only ifdef _WIN32 once

source/shared/core_conn.cpp
```diff
@@ -150,21 +150,55 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
 
         CHECK_CUSTOM_ERROR( is_missing_driver, conn, SQLSRV_ERROR_AE_DRIVER_NOT_INSTALLED, get_processor_arch()) {
             throw core::CoreException();
         }
+#endif // !_WIN32 
     }
     else {
-
+#ifndef _WIN32
+        is_missing_driver = true;
```
lilgreenbird: already did this above before the ifndef

source/shared/core_conn.cpp
```diff
@@ -233,13 +268,58 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
 }
 
+// core_search_odbc_driver
+// This method is meant to be used in a non-Windows environment, 
+// searching for a particular ODBC driver name in the odbcinst.ini file
+// Parameters:
+// driver_version   - a valid value in enum DRIVER_VERSION
+// Return           - a boolean flag that indicates if the specified driver version is found or not
+
+bool core_search_odbc_driver( _In_ std::size_t driver_version )
```
lilgreenbird: can we rename this to something that indicates it's a unix_search_odbc_driver or something?

source/shared/core_conn.cpp
```diff
@@ -282,7 +362,8 @@ SQLRETURN core_odbc_connect( _Inout_ sqlsrv_conn* conn, _Inout_ std::string& con
         SQLCHAR state[ SQL_SQLSTATE_BUFSIZE ];
         SQLSMALLINT len;
         SQLRETURN sr = SQLGetDiagField( SQL_HANDLE_DBC, conn->handle(), 1, SQL_DIAG_SQLSTATE, state, SQL_SQLSTATE_BUFSIZE, &len );
-        // sql state IM002/IM003 in ODBC 17, means that the correct ODBC driver is not installed
+        // sql state IM002, means that the specified ODBC driver is not installed
+        // might return IM003 in ODBC 17 (an issue that will be fixed)
```
lilgreenbird: clarify that this is an issue in ODBC 17 and in the preview version only.
And do we have a backlog item to address this later?

source/shared/core_conn.cpp
```diff
@@ -282,7 +362,8 @@ SQLRETURN core_odbc_connect( _Inout_ sqlsrv_conn* conn, _Inout_ std::string& con
...
         is_missing_driver = ( SQL_SUCCEEDED(sr) && state[0] == 'I' && state[1] == 'M' && state[2] == '0' && state[3] == '0' && (state[4] == '2' || state[4] == '3'));
```
lilgreenbird: why do we need to compare character by character instead of the whole string?

source/shared/core_conn.cpp
```diff
@@ -233,13 +268,58 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+    return found_driver;
+}
 
 // core_odbc_connect
 // calls odbc connect API to establish the connection to server
 // Parameters:
 // conn                 - The connection structure on which we establish the connection 
 // conn_str             - Connection string 
-// missing_driver_error - indicates whether odbc driver is installed on client machine   
+// is_missing_driver    - indicates whether odbc driver is installed on client machine
```
lilgreenbird: see above on moving logic to determine whether drive is missing in unix here

source/shared/core_conn.cpp
yitam: Yes there's a backlog item for this and will be fixed.

source/shared/core_conn.cpp
yitam: good question...

source/shared/core_conn.cpp
```diff
@@ -150,21 +150,55 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
     
     build_connection_string_and_set_conn_attr( conn, server, uid, pwd, options_ht, valid_conn_opts, driver, conn_str TSRMLS_CC );
     
-    bool is_missing_driver = false;
+    bool is_missing_driver = true;
 
     if ( conn->is_driver_set ) {
         r = core_odbc_connect( conn, conn_str, is_missing_driver, is_pooled );
```
v-kaywon: What happens when user set driver to ODBC 13 and ce_option.enabled is true? Will there be an error? Maybe should check for if ( conn-> is_driver_set && conn->ce_option.enabled ) first.

source/shared/core_conn.cpp
```diff
@@ -150,21 +150,55 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+                std::string conn_str_driver = conn_str + CONNECTION_STRING_DRIVER_NAME[ DRIVER_VERSION( i ) ];
+                r = core_odbc_connect( conn, conn_str_driver, is_missing_driver, is_pooled );
+                // simply check the connection succeeded or not and ignore the returned value of 'is_missing_driver' 
+                if ( SQL_SUCCEEDED( r ) ) {
+                    is_missing_driver = false;
+                    break;
+                }
+            }
+        }
+        CHECK_CUSTOM_ERROR( is_missing_driver, conn, SQLSRV_ERROR_DRIVER_NOT_INSTALLED, get_processor_arch()) {
```
v-kaywon: this should be check in line 215 (after `if ( conn->is_driver_set )`, `else if ( conn->ce_option.enabled )` and `else `blocks )

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,134 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
-            CHECK_CUSTOM_ERROR( is_missing_driver && ( i == DRIVER_VERSION::LAST ), conn, SQLSRV_ERROR_DRIVER_NOT_INSTALLED, get_processor_arch()) {
+            r = core_odbc_connect( conn, conn_str, is_pooled );
+        }
+        else {
+            DRIVER_VERSION odbc_version = ODBC_DRIVER_UNKNOWN;
+            for( short i = DRIVER_VERSION::FIRST; i <= DRIVER_VERSION::LAST; ++i ) {
+                // skip ODBC 11 in a non-Windows environment -- only available in Red Hat / SUSE (preview)
+                // https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server#microsoft-odbc-driver-11-for-sql-server-on-linux
+                if (i == DRIVER_VERSION::ODBC_DRIVER_11)    
+                    continue;
```
lilgreenbird: I come from the camp that considers continue, break, goto as bad programming style.  In this case it can be easily avoided by adding this condition in the next if

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,134 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+            }
+        }
+        else {
+            for( short i = DRIVER_VERSION::FIRST; i <= DRIVER_VERSION::LAST; ++i ) {
+                std::string conn_str_driver = conn_str + CONNECTION_STRING_DRIVER_NAME[ DRIVER_VERSION(i) ];
+                r = core_odbc_connect( conn, conn_str_driver, is_pooled );
+
+                if( SQL_SUCCEEDED( r ) ) {
+                    // successfully connected!
+                    break;
```
lilgreenbird: add condition to exit loop criteria instead of breaking out in the middle of a loop.

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,134 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+                }
+                else if(! core_compare_error_state( conn, r, "IM002" ) ) {
+                    // sql state IM002 means that the specified ODBC driver is not installed
+                    // something went wrong other than missing the ODBC driver
+                    break;
```
lilgreenbird: see above

source/shared/core_conn.cpp
```diff
@@ -233,17 +333,84 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+bool core_compare_error_state( _In_ sqlsrv_conn* conn,  _In_ SQLRETURN rc, _In_ const char* error_state )
+{
+    if( SQL_SUCCEEDED( rc ) ) 
+        return false;
+
+    SQLCHAR state[ SQL_SQLSTATE_BUFSIZE ];
+    SQLSMALLINT len;
+    SQLRETURN sr = SQLGetDiagField( SQL_HANDLE_DBC, conn->handle(), 1, SQL_DIAG_SQLSTATE, state, SQL_SQLSTATE_BUFSIZE, &len );
+
+    bool same = false;
```
lilgreenbird: unnecessary local variable, just return (condition)

source/shared/core_conn.cpp
```diff
@@ -233,17 +333,84 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+    std::string driver_str = driver.substr( pos1 + 1, pos2 - pos1 - 1);
+
+    // search for the ODBC driver...
+    const char* driver_name = driver_str.c_str();
+    do
+    {
+        if( strstr( pszBuf, driver_name ) != 0  )
+        {
+            found_driver = true;
+            break;
```
lilgreenbird: do while (blah && !found_driver)

source/shared/core_conn.cpp
```diff
@@ -233,17 +333,84 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+// core_search_odbc_driver_unix
+// This method is meant to be used in a non-Windows environment, 
+// searching for a particular ODBC driver name in the odbcinst.ini file
+// Parameters:
+// driver_version   - a valid value in enum DRIVER_VERSION
+// Return           - a boolean flag that indicates if the specified driver version is found or not
+
+bool core_search_odbc_driver_unix( _In_ DRIVER_VERSION driver_version )
+{
+    char szBuf[DEFAULT_CONN_STR_LEN+1];     // use an arbitrary large enough buffer size
```
lilgreenbird: can we not use the term 'arbitrary' here? sounds like we're guessing and don't know what we're doing

source/shared/core_conn.cpp
```diff
@@ -278,13 +445,6 @@ SQLRETURN core_odbc_connect( _Inout_ sqlsrv_conn* conn, _Inout_ std::string& con
     memset( wconn_string, 0, wconn_len * sizeof( SQLWCHAR )); // wconn_len is the number of characters, not bytes
```
lilgreenbird: memset_s

source/shared/core_conn.cpp
```diff
@@ -908,10 +1068,20 @@ void driver_set_func::func( _In_ connection_option const* option, _In_ zval* val
...
     common_conn_str_append_func( option->odbc_name, val_str, val_len, driver_option TSRMLS_CC );
    
-    CHECK_CUSTOM_ERROR( std::find( CONNECTION_STRING_DRIVER_NAME.begin(), CONNECTION_STRING_DRIVER_NAME.end(), driver_option) == CONNECTION_STRING_DRIVER_NAME.end(), conn, SQLSRV_ERROR_CONNECT_INVALID_DRIVER, val_str){
+    conn->driver_version = ODBC_DRIVER_UNKNOWN;
+    for ( short i = DRIVER_VERSION::FIRST; i <= DRIVER_VERSION::LAST; ++i ) {
+        std::string driver_name = CONNECTION_STRING_DRIVER_NAME[ DRIVER_VERSION( i ) ];
+
+        if (! driver_name.compare( driver_option ) ) { 
+            conn->driver_version = DRIVER_VERSION( i );
+            break;
```
lilgreenbird: add condition to for loop exit condition?

source/shared/core_sqlsrv.h
```diff
@@ -1639,8 +1642,9 @@ enum SQLSRV_ERROR_CODES {
 
     SQLSRV_ERROR_ODBC,
     SQLSRV_ERROR_DRIVER_NOT_INSTALLED,
-    SQLSRV_ERROR_AE_DRIVER_NOT_INSTALLED,
+    SQLSRV_ERROR_AE_DRIVER_REQUIRED,
```
lilgreenbird: multiple occurrances:
I know it's confusing, but it's actually CE. The flags, and all Always Encrypted docs talk about CE. It's SQL_CE_ENABLED in msodbcsql.h too it would be confusing to call talk about an AE driver here

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,134 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
     
     build_connection_string_and_set_conn_attr( conn, server, uid, pwd, options_ht, valid_conn_opts, driver, conn_str TSRMLS_CC );
     
-    bool is_missing_driver = false;
+#ifndef _WIN32
+    if( conn->ce_option.enabled ) {
+        // when AE is enabled, must use ODBC driver 17
```
lilgreenbird: feature is Always Encrypted, option is CE (your comment in the close brace correctly refers to this

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,134 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
             }
-            if ( !is_missing_driver) {
-                break;
+            std::string conn_str_driver = conn_str + CONNECTION_STRING_DRIVER_NAME[ DRIVER_VERSION( odbc_version ) ];
+            r = core_odbc_connect( conn, conn_str_driver, is_pooled );
+        }  // else driver_version not unknown
+    } // else ce_option enabled
+#else
+    if( conn->ce_option.enabled ) {
+        // when AE is enabled, must use ODBC driver 17
```
lilgreenbird: CE

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,134 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+        if( conn->driver_version != ODBC_DRIVER_UNKNOWN ) {
```
lilgreenbird: this block is identical in both _WIN32 cases, move to before ifdef

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,134 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+
+            if( ! SQL_SUCCEEDED( r ) && core_compare_error_state( conn, r, "IM002" ) ) {
+                // sql state IM002 means that the specified ODBC driver is not installed
+                CHECK_CUSTOM_ERROR( true, conn, SQLSRV_ERROR_SPECIFIED_DRIVER_NOT_FOUND, CONNECTION_STRING_DRIVER_NAME[ conn->driver_version ].c_str() ) {
+                    throw core::CoreException();
+                }
+            }
+        }
+        else {
+            for( short i = DRIVER_VERSION::FIRST; i <= DRIVER_VERSION::LAST; ++i ) {
```
lilgreenbird: this looks very messy. There are multiple conditions conditions which break out of the loop. One of them is a successful condition, the others not - sometimes you throw an exception, sometimes you don't. When you don't there's a catchall at the end which checks and throws the exception anyway. 
This is better structured as a while with checks for conditions which will end the loop.

source/shared/core_conn.cpp
lilgreenbird: actually if there is a concern with sensitive data, then memset should definitely not be used as it's going to get optimized out.  I'm not sure if the compiler won't do the same with memset_s too? Depending on the platform I think you would need to cast it to a volatile pointer or pragma disable optimization here? hmm need to check tomorrow to see how ODBC handles this

test/functional/pdo_sqlsrv/pdo_connect_driver.phpt
```diff
@@ -0,0 +1,186 @@
...
+    // Test with {}
+    switch ( $msodbcsql_maj )
+    {
+        case 17:
+            $value = "{ODBC Driver 17 for SQL Server}";
+            break;
+        case 13:
+            $value = "{ODBC Driver 13 for SQL Server}";
+            break;
+        case 12:
```
lilgreenbird: there is no 12

test/functional/pdo_sqlsrv/pdo_connect_driver.phpt
```diff
@@ -0,0 +1,186 @@
...
+    // Test without {}
+    switch ( $msodbcsql_maj )
+    {
+        case 17:
+            $value = "ODBC Driver 17 for SQL Server";
+            break;
+        case 13:
+            $value = "ODBC Driver 13 for SQL Server";
+            break;
+        case 12:
```
lilgreenbird: there is no 12

test/functional/pdo_sqlsrv/pdo_connect_driver.phpt
```diff
@@ -0,0 +1,186 @@
+--TEST--
```
lilgreenbird: now this whole file isn't following the coding standard we just agreed to

source/shared/core_conn.cpp
```diff
@@ -908,10 +1068,20 @@ void driver_set_func::func( _In_ connection_option const* option, _In_ zval* val
     std::string driver_option( "" );
     common_conn_str_append_func( option->odbc_name, val_str, val_len, driver_option TSRMLS_CC );
    
-    CHECK_CUSTOM_ERROR( std::find( CONNECTION_STRING_DRIVER_NAME.begin(), CONNECTION_STRING_DRIVER_NAME.end(), driver_option) == CONNECTION_STRING_DRIVER_NAME.end(), conn, SQLSRV_ERROR_CONNECT_INVALID_DRIVER, val_str){
+    conn->driver_version = ODBC_DRIVER_UNKNOWN;
+    for ( short i = DRIVER_VERSION::FIRST; i <= DRIVER_VERSION::LAST; ++i ) {
+        std::string driver_name = CONNECTION_STRING_DRIVER_NAME[ DRIVER_VERSION( i ) ];
```
v-kaywon: same here, i should already be a DRIVER_VERSION

test/functional/pdo_sqlsrv/pdo_connect_driver.phpt
```diff
@@ -0,0 +1,186 @@
...
+    
+    $value = 123;
+    $connectionOptions = "Driver = $value";
+    $expected = "Invalid value $value was specified for Driver option.";
+    connect_verify_output( $connectionOptions, $expected );
+    
+    $value = false;
+    $connectionOptions = "Driver = $value";
+    $expected = "Invalid value $value was specified for Driver option.";
+    connect_verify_output( $connectionOptions, $expected );
```
v-kaywon: have an array of values and loop over it to avoid repeated code

test/functional/sqlsrv/sqlsrv_connect_driver.phpt
```diff
@@ -75,35 +78,92 @@ function test_valid_values( $msodbcsql_maj ,$server ,$connectionOptions){
...
     $connectionOptions['Driver']=$value;
     $expected = "Invalid value type for option Driver was specified.  String type was expected.";
-    connect_verify_output( $msodbcsql_maj ,$server ,$connectionOptions ,$expected );
+    connect_verify_output( $server, $connectionOptions, $expected );
     
     $value = false;
     $connectionOptions['Driver']=$value;
     $expected = "Invalid value type for option Driver was specified.  String type was expected.";
-    connect_verify_output( $msodbcsql_maj ,$server ,$connectionOptions ,$expected );
+    connect_verify_output( $server, $connectionOptions, $expected );
```
v-kaywon: same, please use array of values and iterate over it

source/shared/core_conn.cpp
yitam: I like the second suggestion better because I don't want it to proceed to get the next driver

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,134 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
 
-        CHECK_CUSTOM_ERROR( is_missing_driver, conn, SQLSRV_ERROR_AE_DRIVER_NOT_INSTALLED, get_processor_arch()) {
-            throw core::CoreException();
+            r = core_odbc_connect( conn, conn_str, is_pooled );
         }
+        else {  
+            // driver not specified, so connect using ODBC 17
+            // In non-Windows environment, unixODBC 2.3.4 and unixODBC 2.3.1 return different error states 
+            // If it fails to connect using ODBC 17, it is unreliable to check for a certain sql state error.  
+            // Thus, we will simply throw an exception.
```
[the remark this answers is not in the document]
yitam: no this can be any exception because while connecting there might be other reasons that it failed

source/shared/core_conn.cpp
```diff
@@ -233,17 +333,84 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
     return return_conn;
 }
 
+// core_compare_error_state
+// This method compares the error state to the one specified
+// Parameters:
+// conn         - the connection structure on which we establish the connection 
+// rc           - ODBC return code
+// Return       - a boolean flag that indicates if the error states are the same
+
```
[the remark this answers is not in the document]
yitam: The other comment blocks above each function in this file were like that. I was trying to be consistent...

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,134 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+        else {
+            for( short i = DRIVER_VERSION::FIRST; i <= DRIVER_VERSION::LAST; ++i ) {
+                std::string conn_str_driver = conn_str + CONNECTION_STRING_DRIVER_NAME[ DRIVER_VERSION(i) ];
+                r = core_odbc_connect( conn, conn_str_driver, is_pooled );
+
+                if( SQL_SUCCEEDED( r ) ) {
+                    // successfully connected!
+                    break;
+                }
+                else if(! core_compare_error_state( conn, r, "IM002" ) ) {
```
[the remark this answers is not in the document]
yitam: I did that before but this breaks one of the tests which specifically put in invalid option for a connection keyword

source/shared/core_conn.cpp
yitam: no it's not identical

test/functional/sqlsrv/sqlsrv_connect_driver.phpt
yitam: For pdo test we could do that but in sqlsrv the expected errors are not all the same.

source/shared/core_conn.cpp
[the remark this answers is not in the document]
lilgreenbird: hmm on 1st thought I wanted to say to fix the test...
but thinking more about this I think it's ok to stop there. While it's true that there can be other reasons for failing to connect to ODBC 13 I think it's fine to stop and not continue and try ODBC 11 because f you have ODBC 13 installed then I would think you do want to use ODBC 13. Why else would have have a newer version? If you wanted to use ODBC 11 you would've specified that explicitly. So if for some reason there was a failure connecting to ODBC 13 then I I think you would want to know and fix that instead of automatically going down to ODBC 11.

test/functional/pdo_sqlsrv/pdo_connect_driver.phpt
yitam: newer versions of ODBC 11 do return 12

lilgreenbird: you are right! pls add a comment about this

source/shared/core_conn.cpp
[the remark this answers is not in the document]
v-kaywon: Isn't this comment referring to the exception being thrown in line 177 (SQLSRV_ERROR_CE_DRIVER_REQUIRED)? This error is specific for the SQLSRV and PDO_SQLSRV driver right?

source/shared/core_conn.cpp
```diff
@@ -246,34 +235,36 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
-                if( SQL_SUCCEEDED( r ) ) {
-                    // successfully connected!
-                    break;
-                }
-                else if(! core_compare_error_state( conn, r, "IM002" ) ) {
-                    // sql state IM002 means that the specified ODBC driver is not installed
-                    // something went wrong other than missing the ODBC driver
-                    break;
+                if( SQL_SUCCEEDED( r ) ) { 
+                    done = true;
```
v-kaywon: Consider combining the first and last case together since they do the same thing?
`if( SQL_SUCCEEDED( r ) || !core_compare_error_state( conn, r, "IM002" ) )`

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,125 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
     build_connection_string_and_set_conn_attr( conn, server, uid, pwd, options_ht, valid_conn_opts, driver, conn_str TSRMLS_CC );
     
-    bool is_missing_driver = false;
+    if( conn->ce_option.enabled ) {
+        // when column encryption is enabled, must use ODBC driver 17
+        if( conn->driver_version != ODBC_DRIVER_UNKNOWN ) {
+            CHECK_CUSTOM_ERROR( conn->driver_version != ODBC_DRIVER_17, conn, SQLSRV_ERROR_CE_DRIVER_REQUIRED, get_processor_arch() ) {
+                throw core::CoreException();
+            }
+#ifndef _WIN32
```
lilgreenbird: eww this ifndef is inside a nested if and there are multiple conditions after this before the else then it gets back to this condition and indent level. Looks very messy hard to read and maintain.
The condition it's checking before this simply throws an except and exits anyway why even bother with the else and all the indentation?
ie if (ce enabled && (driver unknown || driver doesn't exist) die

source/shared/core_conn.cpp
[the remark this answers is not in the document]
yitam: No you have misunderstood... line 177 is invoked when ODBC 17 can be found
in the comment above the exception refers to the returned value of `core_odbc_connect()` which can be any exception from ODBC while in attempt to connect

source/shared/core_conn.cpp
yitam: sorry I'm not sure what you meant here??

source/shared/core_conn.cpp
yitam: the thing is the way to check if driver exists or not is very different in Windows and in Linux/mac...
it was you who insisted that I moved these few lines outside of ifndef... I already said it was harder to read

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,125 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+            }
+            std::string conn_str_driver = conn_str + CONNECTION_STRING_DRIVER_NAME[odbc_version];
+            r = core_odbc_connect( conn, conn_str_driver, is_pooled );
+        }  // else driver_version not unknown
+    } // else ce_option enabled
+#else
+            r = core_odbc_connect( conn, conn_str, is_pooled );
+
+            if( core_compare_error_state( conn, r, "IM002" ) ) {
+                // sql state IM002 means that the specified ODBC driver is not installed
```
lilgreenbird: IM002 means Data source not found and no default driver specified. You could get this if the driver was installed but the DSN specified isn't defined or the DSN specifies a path to the driver that's wrong.

source/shared/core_conn.cpp
lilgreenbird: the original comment I suggested was to refactor common c&p code outside the ifdef, not nest it to make it harder to read.  If there's any misunderstanding, don't make any changes that doesn't make any sense to you.

The block in green above is common in all platforms and results in exception being thrown.
Why not just:

if (ce enabled && !driver 17)
  throw

#ifndef WIN 
//do this
#else 
//do that

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,125 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
+        } 
+    }
+    else { 
+        // column encryption NOT enabled
+        if( conn->driver_version != ODBC_DRIVER_UNKNOWN ) {
+            r = core_odbc_connect( conn, conn_str, is_pooled );
+
+            if( core_compare_error_state( conn, r, "IM002" ) ) {
+                // sql state IM002 means that the specified ODBC driver is not installed
+                CHECK_CUSTOM_ERROR( true, conn, SQLSRV_ERROR_SPECIFIED_DRIVER_NOT_FOUND, CONNECTION_STRING_DRIVER_NAME[conn->driver_version].c_str() ) {
```
lilgreenbird: 1. please be more consistent -
  sometimes we do core_compare_error_state() then CHECK_CUSTOM_ERROR(true, blah)
  sometimes we do CHECK_CUSTOM_ERROR(core_compare_error_state(), blah)
  the latter is preferable IMHO
2. there are multiple places where this is done, can we refactor this so it's not in multiple places??

source/shared/core_conn.cpp
v-kaywon: both `if `and `else `make `done = true`, you can combine them

source/shared/core_conn.cpp
```diff
@@ -150,34 +150,125 @@ sqlsrv_conn* core_sqlsrv_connect( _In_ sqlsrv_context& henv_cp, _In_ sqlsrv_cont
...
-        } // for
+        }
+        else {
+            bool done = false;
+            for( short i = DRIVER_VERSION::FIRST; i <= DRIVER_VERSION::LAST && ! done; ++i ) {
+                std::string conn_str_driver = conn_str + CONNECTION_STRING_DRIVER_NAME[i];
+                r = core_odbc_connect( conn, conn_str_driver, is_pooled );
+
+                if( SQL_SUCCEEDED( r ) ) { 
+                    done = true;
```
lilgreenbird: done = SQL_SUCCEEDED(r)

PR #560, opened 2017-09-30
Skip fix connres aad

david-puglielli:


test/functional/sqlsrv/skipif_version_not_2k14.inc
```diff
@@ -0,0 +1,30 @@
+<?php
```
lilgreenbird: the name is misleading this will return true for 2014 and above.

test/functional/sqlsrv/skipif_version_not_2k16.inc
```diff
@@ -0,0 +1,30 @@
+<?php
```
lilgreenbird: see above comment on name

test/functional/sqlsrv/skipif_version_not_2k14.inc
david-puglielli: skipif_version_less_than_2k14.inc? Name starting to get a bit unwieldy...

PR #567, opened 2017-10-12
convert long output param to string, then convert back to long if less than max int

v-kaywon:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/567)

### DIFF
--- a/source/shared/core_stmt.cpp
+++ b/source/shared/core_stmt.cpp
@@ -386,8 +386,12 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
                 }
                 if( zval_was_long ){
                     convert_to_string( param_z );
-                    if ( encoding != SQLSRV_ENCODING_SYSTEM && encoding != SQLSRV_ENCODING_UTF8 && encoding != SQLSRV_ENCODING_BINARY )
-                        encoding = SQLSRV_ENCODING_UTF8;
+                    if (encoding != SQLSRV_ENCODING_SYSTEM && encoding != SQLSRV_ENCODING_UTF8 && encoding != SQLSRV_ENCODING_BINARY) {
+                        encoding = stmt->encoding();
+                        if (encoding == SQLSRV_ENCODING_DEFAULT) {
+                            encoding = stmt->conn->encoding();
+                        }
+                    }
                     match = Z_TYPE_P( param_z ) == IS_STRING;
                 }
                 else {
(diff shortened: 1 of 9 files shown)


source/shared/core_stmt.cpp
```diff
@@ -383,7 +384,14 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
                 if( zval_was_null || zval_was_bool ) {
                     convert_to_long( param_z );
                 }
-                match = Z_TYPE_P( param_z ) == IS_LONG;
+                if( zval_was_long ){
+                    convert_to_string( param_z );
+					encoding = SQLSRV_ENCODING_SYSTEM;
```
v-kaywon: will fix indentation

source/shared/core_stmt.cpp
```diff
@@ -551,7 +565,7 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
                                                    buffer, buffer_len TSRMLS_CC );
 
                 // save the parameter to be adjusted and/or converted after the results are processed
-				sqlsrv_output_param output_param( param_ref, encoding, param_num, static_cast<SQLUINTEGER>( buffer_len ));
+				sqlsrv_output_param output_param( param_ref, encoding, param_num, static_cast<SQLUINTEGER>( buffer_len ), zval_was_long );
```
v-kaywon: will fix indentation

source/shared/core_stmt.cpp
```diff
@@ -415,7 +423,13 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
     if( direction == SQL_PARAM_OUTPUT ) {
         switch( php_out_type ) {
             case SQLSRV_PHPTYPE_INT:
-                convert_to_long( param_z );
+                if( zval_was_long ){
+                    convert_to_string( param_z );
+                    encoding = SQLSRV_ENCODING_SYSTEM;
+                }
```
david-puglielli: Do any encoding issues relating to use of commas as separators crop up? I should think that convert_to_string just takes the int and converts to a string of digits, eg. 1,000,000 becomes '1000000'. But if not, do we run into any issues handling commas when calling convert_to_double or convert_to_long below?

test/functional/pdo_sqlsrv/pdo_bigint_outparam.phpt
```diff
@@ -0,0 +1,82 @@
...
+<?php
+require_once("MsCommon_mid-refactor.inc");
+
+$conn = connect();
+
+// Create the table
+$tbname = "bigint_table";
+createTable($conn, $tbname, array("c1_bigint" => "bigint"));
+
+// Create a Store Procedure
```
david-puglielli: *stored

test/functional/pdo_sqlsrv/pdo_bigint_outparam.phpt
```diff
@@ -0,0 +1,82 @@
...
+$stmt->bindParam(1, $bigintOut, PDO::PARAM_INT | PDO::PARAM_INPUT_OUTPUT, PDO::SQLSRV_PARAM_OUT_DEFAULT_SIZE);
+$stmt->execute();
+printf("Large bigint inout:\n" );
+var_dump($bigintOut);
+printf("\n");
+
+$conn->exec("TRUNCATE TABLE $tbname");
+
+// Insert a small bigint
+insertRow($conn, $tbname, array("c1_bigint" => 922337203));
```
david-puglielli: Is there any particular reason for choosing this value?

v-kaywon: this is just a random value truncated from the bigint input

source/shared/core_stmt.cpp
v-kaywon: I've tried testing with both encodings (utf-8 and char) and they both are fine. From wheat I researched, user can only retrieve a comma separated integer by explicitly casting an int to a string. If that is the case, the php_out_type would not be INT anymore, instead it would be a string, and the problem with bigint to int conversion would not exist.

As for the second question "do we run into any issues handling commas when calling convert_to_double or convert_to_long below", this will not be a problem since the string retrieved will not have any comma separators.

test/functional/pdo_sqlsrv/pdo_bigint_outparam.phpt
lilgreenbird: perhaps add that to description?

PR #575, opened 2017-10-24
Refactored SQLSRV tests to test AE with various data types

yitam:
Mainly the tests using a test table with 28 data types

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/575)


test/functional/sqlsrv/MsCommon.inc
```diff
@@ -436,56 +347,12 @@ function insertCheck($stmt)
...
-    return (trim($data));
-}
-
 function createProc($conn, $procName, $procArgs, $procCode)
 {
     dropProc($conn, $procName);
     $stmt = sqlsrv_query($conn, "CREATE PROC [$procName] ($procArgs) AS BEGIN $procCode END");
     if ($stmt === false) {
-        fatalError("Failed to create test procedure");
+        fatalError("Failed to create test procedure", true);
```
v-kaywon: default is already true, no need to put this in

test/functional/sqlsrv/MsCommon.inc
```diff
@@ -509,7 +376,7 @@ function callProcEx($conn, $procName, $procPrefix, $procArgs, $procValues)
 {
     $stmt = sqlsrv_query($conn, "{ $procPrefix CALL [$procName] ($procArgs)}", $procValues);
     if ($stmt === false) {
-        fatalError("Failed to call test procedure");
+        fatalError("Failed to call test procedure", true);
```
v-kaywon: same as line 355

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -28,22 +28,67 @@ const INSERT_PREPARE_PARAMS = 4;
...
+        } elseif (!strcasecmp($this->dataType, "xml") ||
+            !strcasecmp($this->dataType, "timestamp") ||
+            !strcasecmp($this->dataType, "image") ||
+            !strcasecmp($this->dataType, "ntext") ||
+            !strcasecmp($this->dataType, "text") ||
+            !strcasecmp($this->dataType, "sql_variant") ||
+            !strcasecmp($this->dataType, "hierarchyid") ||
+            !strcasecmp($this->dataType, "geography") ||
+            !strcasecmp($this->dataType, "geometry") ||
+            !strcasecmp($this->dataType, "alias")) {
```
v-kaywon: money and smallmoney are also not supported, why comment out?

v-kaywon: also it would look nicer to put all the unsupported types in an array and there an array_in function that you can use in php to find a match

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -28,22 +28,67 @@ const INSERT_PREPARE_PARAMS = 4;
...
+
         $this->encType = ($deterministic ? "deterministic" : "randomized");
+        if (empty($dataType)) {
+            echo "Data type can not be null or empty!\n";
+        }
         $this->dataType = $dataType;
         $this->options = $options;
+
+        // first asssumes the column is not encryptable
+        $this->encryptable = false;
```
v-kaywon: why not put this in an else block instead of assuming?

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -28,22 +28,67 @@ const INSERT_PREPARE_PARAMS = 4;
...
+        }
+    }
+
+    /**
+     * if Always Encrypted not supported for this data type set $noEncryption to false
+     * @return void
+     */
+    protected function checkIfUnsupported()
+    {
+        $this->encryptable = true;
```
v-kaywon: same here, you can put this into an else block

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -325,6 +377,7 @@ function connect($options = array(), $disableCE = false)
  * @param  object  $conn : sqlsrv connection object
  * @param  string  $tbname : name of the table to be created
  * @param  array  $columnMetaArr : array of ColumnMeta objects, which contain metadata for one column
+ * @return object sqlsrv statement
```
v-kaywon: I believe sqlsrv statement is a resource not an object, same with sqlsrv connection

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -361,7 +414,9 @@ function insertRow($conn, $tbname, $inputs, &$r = null, $api = INSERT_QUERY)
                 break;
             case INSERT_PREPARE:
                 $stmt = sqlsrv_prepare($conn, $insertSql);
-                $r = sqlsrv_execute($stmt);
+                if ($stmt) {
+                    $r = sqlsrv_execute($stmt);
+                }
```
v-kaywon: if you're going to check $stmt, maybe output the error as well

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -370,32 +425,106 @@ function insertRow($conn, $tbname, $inputs, &$r = null, $api = INSERT_QUERY)
         $params = array();
         foreach ($inputs as $key => $input) {
             if (is_object($input)) {
-                array_push($params, $input->bindParamArr($inputs[$key]));               
+                array_push($params, $input->bindParamArr($input->value));
```
v-kaywon: this is very strange looking. If you're calling a member function from $input, the function will have access to everything in $input. Passing in $input->value is not necessary. Please change bindParamArr function accordingly.

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -370,32 +425,106 @@ function insertRow($conn, $tbname, $inputs, &$r = null, $api = INSERT_QUERY)
...
+ * Perform a query from a table with or without where condition(s)
+ * @param  resource  $conn : connection resource
+ * @param  string  $sql : T-SQL query
+ * @param  string  $conds : string of condition(s) with placeholders, null by default
+ * @param  array  $values : array of parameters, null by default
+ * @return object sqlsrv statement upon success or false otherwise
+ */
+function executeQuery($conn, $sql, $conds = null, $values = null)
+{
+    // when AE is enabled, use sqlsrv_prepare() to handle fields with unlimited size
```
v-kaywon: how can you tell if a field has unlimited size from here? If AE is enabled, does it always use sqlsrv_prepare?

test/functional/sqlsrv/MsHelper.inc
yitam: when we do not encrypt money or smallmoney one of the AE tests with money types fail...
that's why I changed the ColumnMeta to type an optional parameter to not to encrypt a particular column instead

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -410,6 +539,201 @@ function fetchAll($conn, $tbname)
     }
 }
 
+/**
+ * Create a test table with columns of various types given a table name,
+ * all deterministic if AE is enabled
+ * @param  resource  $conn : connection resource
+ * @param  string  $tbname : name of the table to create
+ * @return object sqlsrv statement
```
v-kaywon: sqlsrv statement is a resoure

test/functional/sqlsrv/MsHelper.inc
yitam: yes, I'll modify the comments accordingly

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -410,6 +539,201 @@ function fetchAll($conn, $tbname)
...
+    }
+    return ($count);
+}
+
+/**
+ * Insert one or more specific rows of test data into the test table
+ * @param  resource  $conn : connection resource
+ * @param  string  $tbname : name of the table to create
+ * @param  int  $index : the index of a certain row of test data
+ * @return object sqlsrv statement upon success or false otherwise
```
v-kaywon: resource

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -410,6 +539,201 @@ function fetchAll($conn, $tbname)
...
+
+    // get array of input values
+    $inputArray = getInsertArray($index);
+    if (empty($inputArray)) {
+        fatalError("getInsertSqlComplete: inputs for inserting a row cannot be empty");
+    }
+    
+    $result = null;
+    if (isColEncrypted()) {
+        $stmt = insertRow($conn, $tbname, $inputArray, $result, INSERT_PREPARE_PARAMS);
```
v-kaywon: INSERT_PREPARE_PARAMS is redundant since in insertRow, isColEncrypted() and INSERT_PREPARE_PARAMS go to the same path (line 435)

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -336,18 +389,18 @@ function createTable($conn, $tbname, $columnMetaArr)
...
  * Insert a row into a table
  * @param  object  $conn : sqlsrv connection object
  * @param  string  $tbname : name of the table for the row to be inserted
- * @param  array  $inputs : an associative array column name and its value, which may be a 
+ * @param  array  $inputs : an associative array column name and its value, which may be a
  *                          literal or a BindParamOption object
  * @param  bool  $r : true if the row was successfully inserted, otherwise false. Default value is null to make this parameter optional.
  * $param  string  $api : SQLSRV API used for executing the insert query
- *                      accepted values: INSERT_QUERY, INSERT_PREPARE, INSERT_QUERY_PARAMS, INSERT_PREPARE_PARAMS
+ *                        accepted values: INSERT_QUERY, INSERT_PREPARE, INSERT_QUERY_PARAMS, INSERT_PREPARE_PARAMS
```
v-kaywon: since you've already defined these constants, they are no longer string, use int

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -410,6 +539,201 @@ function fetchAll($conn, $tbname)
...
+        $col = 1;
+        foreach ($inputArray as $key => $value) {
+            $colStr .= $key . ", ";
+            
+            if (is_array($value)) {
+                $value = $value[0];
+                // this might be an input to a decimal, a numeric or a binary field
+                if (isBinary($col)) {
+                    $value = "0x" . $value;    // annotate the input string as a hex string
+                }
```
v-kaywon: can't you call insertRow after you've modified the binary data?

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -410,6 +539,201 @@ function fetchAll($conn, $tbname)
...
+function getInsertArray($index)
+{
+    if (! UseUTF8data()) {
+        require_once('MsData.inc');
+    } else {
+        require_once('MsData_UTF8.inc');
+    }
+
+    // get array of input values
+    return getTestData($index);
```
v-kaywon: did you mean getInsertData?

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -410,6 +539,201 @@ function fetchAll($conn, $tbname)
...
+ * @param  int  $rowIndex : the row index
+ * @param  int  $colIndex : the column index
+ * @return data of mixed type or null if something went wrong
+ */
+function getInsertData($rowIndex, $colIndex)
+{
+    // get array of input values
+    $inputArray = getInsertArray($rowIndex);
+    if (empty($inputArray)) {
+        fatalError("getInsertData: failed to retrieve data at row $rowIndex and column $colIndex.\n");
```
v-kaywon: get rid of "and column $colIndex" since the error comes from only $rowIndex

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -449,7 +773,7 @@ function getSqlType($k)
     return ("udt");
 }
 
-function getDriverType($k, $dataSize)
+function getDriverType($k, $dataSize = 512)
```
v-kaywon: where does 512 come from?

test/functional/sqlsrv/TC24_Close.phpt
```diff
@@ -6,24 +6,22 @@ that resources are invalidated when connection is closed.
...
-    $conn1 = connect();
-    $tableName = 'TC24test';
+    $conn1 = AE\connect();
+    $tableName = 'TC24test';  
     
-    createTable($conn1, $tableName);
-    insertRows($conn1, $tableName, $noRows);
+    // Insert a couple of random rows
+    AE\createTestTable($conn1, $tableName);
+    AE\insertTestRows($conn1, $tableName, 2);
```
v-kaywon: why change $noRows from 5 to 2?

test/functional/sqlsrv/TC24_Close.phpt
```diff
@@ -40,43 +38,32 @@ function ConnectionClose()
     }
 
     // Invalid Statement
-    $conn2 = connect();
-    $stmt2 = selectFromTable($conn2, $tableName);
+    $conn2 = AE\connect();
+    $stmt2 = sqlsrv_query($conn2, "SELECT * FROM [$tableName]");
```
v-kaywon: why can't you call selectFromTable?

test/functional/sqlsrv/TC31_SimpleQuery.phpt
```diff
@@ -9,28 +9,28 @@ Basic verification of query statements (via "sqlsrv_query"):
...
-    $conn1 = connect();
+    $conn1 = AE\connect();
 
-    createTable($conn1, $tableName);
+    // just create an empty table
+    $stmt1 = sqlsrv_query($conn1, "CREATE TABLE $tableName (dummyColumn int)");
 
     trace("Executing SELECT query on $tableName ...");
-    $stmt1 = selectFromTable($conn1, $tableName);
+    $stmt1 = sqlsrv_query($conn1, "SELECT * FROM $tableName");
```
v-kaywon: why not use createTable and selectFromTable?

test/functional/sqlsrv/TC31_SimpleQuery.phpt
```diff
@@ -9,28 +9,28 @@ Basic verification of query statements (via "sqlsrv_query"):
...
     $tableName = 'TC31test';
 
-    $conn1 = connect();
+    $conn1 = AE\connect();
 
-    createTable($conn1, $tableName);
+    // just create an empty table
+    $stmt1 = sqlsrv_query($conn1, "CREATE TABLE $tableName (dummyColumn int)");
 
     trace("Executing SELECT query on $tableName ...");
```
v-kaywon: I thought we were going to get rid of trace

test/functional/sqlsrv/TC31_SimpleQuery.phpt
```diff
@@ -9,28 +9,28 @@ Basic verification of query statements (via "sqlsrv_query"):
...
-    $stmt1 = selectFromTable($conn1, $tableName);
+    $stmt1 = sqlsrv_query($conn1, "SELECT * FROM $tableName");
     $rows = rowCount($stmt1);
-    ;
     sqlsrv_free_stmt($stmt1);
```
v-kaywon: please dropTable at the end of the test

test/functional/sqlsrv/TC32_DeleteQuery.phpt
```diff
@@ -6,47 +6,49 @@ validates the outcome reported by "sqlsrv_rows_affected".
...
         $delRows = 1;
         if (strlen($keyValue) == 0) {
-            $stmt2 = executeQuery($conn1, "DELETE TOP(1) FROM [$tableName]");
+            $stmt2 = AE\executeQuery($conn1, "DELETE TOP(1) FROM [$tableName]");
             $cond = "(top row)";
         } else {
-            $cond = "(c1_int = $keyValue)";
-
-            $stmt3 = selectFromTableEx($conn1, $tableName, $cond);
+            $cond = "(c1_int = ?)";
```
v-kaywon: Is it a good idea to change the test to use a placeholder completely? Now even if isColEnc() is false, the test still uses sqlsrv_prepare with placeholder to run the query, whereas before is used to use sqlsrv_query on a direct query. When refactoring tests, we want to keep the behavior of the original test as intact as possible.

test/functional/sqlsrv/MsHelper.inc
yitam: I reverted the change now. Thanks!

test/functional/sqlsrv/MsHelper.inc
yitam: I changed it to `int`

test/functional/sqlsrv/MsHelper.inc
yitam: no I can't because binary data needs special handling which the generic function insertRow not supposed to know

test/functional/sqlsrv/MsHelper.inc
yitam: It was for convenience, but I only need it for one test, so I'll move this constant to that test instead.

test/functional/sqlsrv/TC24_Close.phpt
yitam: Just thought the essence of the test is not really the data. Anyway, I've reverted the change.

test/functional/sqlsrv/TC31_SimpleQuery.phpt
yitam: I don't mind keeping them around. Some tests have useful trace statements.

test/functional/sqlsrv/TC38_InvalidQuery.phpt
```diff
@@ -16,23 +16,23 @@ function invalidQuery()
...
     if ($stmt1) {
         die("Invalid query should have failed.");
     }
 
-    $dataType = "[c1] int, [c2] int";
     $tableName = 'TC38test';
-    createTableEx($conn1, $tableName, $dataType);
-
+    $columns = array(new AE\ColumnMeta('int', 'c1'), 
+                     new AE\ColumnMeta('int', 'c2'));
```
v-kaywon: where is createTable?

test/functional/sqlsrv/TC43_FetchData.phpt
```diff
@@ -45,11 +44,13 @@ function FetchFields()
                 fatalError("Field $j of Row $i is missing");
             }
             $col = $j + 1;
-            if (!IsUpdatable($col)) {
-                $skipCount++;
-            } else { // should check data even if $fld is null
-                $data = getInsertData($startRow + $i, $col, $skipCount);
-                if (!CheckData($col, $fld, $data)) {
+            $data = AE\getInsertData($startRow + $i, $col);
```
v-kaywon: how come you getInsertData before checking if $col isUpdatable?

Can you also explain what $skipCount does and why you got rid of it?

test/functional/sqlsrv/TC46_FetchNextResult.phpt
```diff
@@ -66,10 +65,10 @@ function fetchFields()
             }
         }
     }
-    if (sqlsrv_next_result($stmt1) ||
-        sqlsrv_next_result($stmt2)) {
+    if ($r1 = sqlsrv_next_result($stmt1) ||
+        $r2 = sqlsrv_next_result($stmt2)) {
```
v-kaywon: why do you need to introduce $r1 and $r2?

test/functional/sqlsrv/TC47_ParamQuery.phpt
```diff
@@ -11,27 +11,34 @@ PHPT_EXEC=true
...
+            $columns = array(new AE\ColumnMeta('int', 'c1'),
+                             new AE\ColumnMeta($sqlType, 'c2'));
+            traceData($sqlType, $data);
+
+            $res = null;
+            AE\createTable($conn1, $tableName, $columns);
+            AE\insertRow($conn1,
+                $tableName,
+                array("c1"=>$k, "c2"=>array($data, SQLSRV_PARAM_IN, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR), $phpDriverType)), 
+                $res,
```
v-kaywon: can you just put null here?

test/functional/sqlsrv/TC52_StreamSend.phpt
```diff
@@ -101,35 +101,58 @@ function SendStream($minType, $maxType, $atExec)
...
     dropTable($conn1, $tableName);
+    unlink($fileName);
 
     sqlsrv_close($conn1);
 
     endTest($testName);
 }
 
-function CreateQueryTable($conn, $table, $dataType, $dataCols, $dataValues, $dataOptions, $execMode)
+function insertQueryTable($conn, $tableName, $params, $execMode)
```
v-kaywon: can you not use insertRow for this function?

test/functional/sqlsrv/TC64_TransactionProc.phpt
```diff
@@ -31,79 +31,71 @@ function Transaction($minType, $maxType)
...
                 break;
             default:
-                $data = GetSampleData($k);
+                $data = getSampleData($k);
                 break;
         }
         if ($data != null) {
-            $sqlType = GetSqlType($k);
+            $sqlType = getSqlType($k);
+            $driverType = getDriverType($k);
```
v-kaywon: the name of this function is very misleading. I thought getDriverType gets the PHP zend type... can you change the name of this to getSqlsrvSqltype() or something more clear?

test/functional/sqlsrv/TC74_ProcQuery.phpt
```diff
@@ -74,40 +74,50 @@ function ProcQuery($minType, $maxType)
     endTest($testName);
 }
 
-function ExecProcQuery($conn, $procName, $dataType, $inData1, $inData2, $outData)
+function execprocQuery($conn, $procName, $type, $dataType, $inData1, $inData2, $outData)
```
v-kaywon: execProcQuery

test/functional/sqlsrv/TC75_ProcRoundtrip.phpt
```diff
@@ -11,22 +11,24 @@ PHPT_EXEC=true
 <?php
 require_once('MsCommon.inc');
 
-function StoredProcRoundtrip($minType, $maxType)
+function storedProcRoundtrip($minType, $maxType)
 {
-    include 'MsSetup.inc';
+    // include 'MsSetup.inc';
```
v-kaywon: delete this line

test/functional/sqlsrv/TC79_RegressionVSTS846501.phpt
```diff
@@ -11,57 +11,50 @@ PHPT_EXEC=true
...
     startTest($testName);
 
     setup();
-    $conn1 = connect();
+    $conn1 = AE\connect();
     // empty parameter array
-    $s = sqlsrv_query($conn1, "DROP TABLE test_bq");
-    $s = sqlsrv_query($conn1, "CREATE TABLE test_bq (id INT IDENTITY NOT NULL, test_varchar_max varchar(max))");
+    $tableName = "test_bq";
+    dropTable($conn1, $tableName);
```
v-kaywon: I think createTable already drop the table before creating the table, so this line is not necessary.

test/functional/sqlsrv/TC32_DeleteQuery.phpt
yitam: I decided to put the logic in AE\executeQuery() instead so that each placeholder will be replaced with literal value if not encrypted

test/functional/sqlsrv/TC46_FetchNextResult.phpt
yitam: It was for debugging when I thought I discovered a bug... forgot to remove them afterwards ;)

test/functional/sqlsrv/TC47_ParamQuery.phpt
yitam: I just tried to be faithful to the original test, in which it called a function insertRowEx() that does the following:
``` $stmt = sqlsrv_query($conn, "INSERT INTO [$tableName] ($dataCols) VALUES ($dataValues)", $dataOptions);```

File: test/functional/sqlsrv/TC43_FetchData.phpt (hunk 25 above)
Comment (yitam) [reply test/functional/sqlsrv/TC43_FetchData.phpt]: Only the timestamp column (column 27) is updatable, and the original way to look for insert data is to skip this column. I'll explain more if this is unclear.

File: test/functional/sqlsrv/TC52_StreamSend.phpt (hunk 28 above)
Comment (yitam) [reply test/functional/sqlsrv/TC52_StreamSend.phpt]: no because it's specific to this test, in which they set the option `SendStreamParamsAtExec` on and off for testing

File: test/functional/sqlsrv/TC64_TransactionProc.phpt (hunk 29 above)
Comment (yitam) [reply test/functional/sqlsrv/TC64_TransactionProc.phpt]: I agree that it's confusing, but that was one of the common functions originally in MsCommon... I'll see how many tests are affected by the renaming.

File: test/functional/sqlsrv/MsHelper.inc [hunk 33]
```diff
@@ -28,22 +28,59 @@ const INSERT_PREPARE_PARAMS = 4;
...
+
         $this->encType = ($deterministic ? "deterministic" : "randomized");
+        if (empty($dataType)) {
+            echo "Data type can not be null or empty!\n";
+        }
         $this->dataType = $dataType;
         $this->options = $options;
+
+        // first asssumes the column is not encryptable
+        // $this->encryptable = false;
```
v-kaywon: Please get rid of comments

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -462,22 +456,38 @@ function selectFromTable($conn, $tbname, $conds = null, $values = null)
...
  * @param  array  $values : array of parameters, null by default
- * @return object sqlsrv statement upon success or false otherwise
+ * @return resource sqlsrv statement upon success or false otherwise
  */
 function executeQuery($conn, $sql, $conds = null, $values = null)
 {
-    // when AE is enabled, use sqlsrv_prepare() to handle fields with unlimited size
-    if (!isColEncrypted() && (is_null($conds) || empty($values))) {
+    if (!isColEncrypted()) {
+        if (!empty($values)) {
```
v-kaywon: can you add a comment saying this if block is for replacing '?' placeholders with values

test/functional/sqlsrv/TC31_SimpleQuery.phpt
v-kaywon: But we never turn tracemode on.. If you think the output is useful, then use echo or print

test/functional/sqlsrv/TC43_FetchData.phpt
v-kaywon: OK I'm beginning to understand what $skipCount is for now. But I still don't understand why you have getInsertData in line 47. There is an identical line in line 50, and $data is only used inside the isUpdatable if block.

test/functional/sqlsrv/TC47_ParamQuery.phpt
v-kaywon: ?? I was talking about $res

test/functional/sqlsrv/TC31_SimpleQuery.phpt
yitam: I personally do when I need to debug what goes wrong. If I use echo/print if will affect the expected output, which is often unnecessary.

test/functional/sqlsrv/TC43_FetchData.phpt
yitam: Yes, it was an oversight. Thanks!

test/functional/sqlsrv/TC47_ParamQuery.phpt
yitam: nope... I will get the error `Fatal error: Uncaught Error: Cannot pass parameter 4 by reference `

PR #579, opened 2017-10-26
added error handling for using direct_query and emulate_prepare in a Column Encryption enabled connection

v-kaywon:
The PDO_SQLSRV driver do not support direct query and emulate prepare in a column encryption enabled connection.

The driver does not support direct query because the MSODBC API SQLDescribeParam must be called to get the sqltype for that parameter. However, SQLDescribeParam can only be used on a prepared statement.

The driver does not support emulate prepare because in a emulate prepare enabled statement, bind parameter is done on the client side and a complete query is sent to the server. Column encryption however require to use the parameterized query feature on the server side.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/579)


source/pdo_sqlsrv/pdo_util.cpp
```diff
@@ -409,6 +409,14 @@ pdo_error PDO_ERRORS[] = {
...
         { IMSSP, (SQLCHAR*) "The specified ODBC Driver is not found.", -80, false }
     },
+    {
+        PDO_SQLSRV_ERROR_CE_DIRECT_QUERY_UNSUPPORTED,
+        { IMSSP, (SQLCHAR*) "Connection with Column Encryption enabled do no support PDO::SQLSRV_ATTR_DIRECT_QUERY with binding parameters.", -81, false }
+    },
+    {
+        PDO_SQLSRV_ERROR_CE_EMULATE_PREPARE_UNSUPPORTED,
+        { IMSSP, (SQLCHAR*) "Connection with Column Encryption enabled do no support PDO::ATTR_EMULATE_PREPARES with binding parameters.", -82, false }
+    },
```
yitam: Please fix the two error messages: do _NOT_ support... 
Suggestion: Use English instead of the attribute names?
E.g. Emulate prepare is not supported when connecting with Column Encryption enabled. Something like that. I believe David might have a better idea.

test/functional/pdo_sqlsrv/pdo_ae_unsupported_stmt_attr.phpt
```diff
@@ -0,0 +1,66 @@
...
+    $st = $connection->prepare("INSERT INTO $tbname (name) VALUES (:p0)", array(PDO::ATTR_EMULATE_PREPARES => true));
+    $st->execute(array("p0" => $name));
+} catch (PDOException $e) {
+    $error = $e->errorInfo;
+    // expects an exception if Column Encryption is enabled
+    if (isColEncrypted()) {
+        if ($error[0] != "IMSSP" ||
+            $error[1] != -82 ||
+            $error[2] != "Connection with Column Encryption enabled do no support PDO::ATTR_EMULATE_PREPARES with binding parameters.") {
+            echo "An unexpected exception was caught.\n";
```
yitam: modify the error message :)

source/pdo_sqlsrv/pdo_util.cpp
v-kaywon: I will keep the attribute name since I think it's more clear for the user which attributes are not supported. But I'll change the error message according to your suggestions. :)

test/functional/pdo_sqlsrv/pdo_ae_unsupported_stmt_attr.phpt
```diff
@@ -0,0 +1,66 @@
+--TEST--
+Test emulate prepare utf8 encoding set at the statement level
```
v-kaywon: Just noticed this. Will change for next commit.

source/pdo_sqlsrv/pdo_util.cpp
david-puglielli: I agree, keep the name of the attribute in the error message.

source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -1146,9 +1146,19 @@ int pdo_sqlsrv_stmt_param_hook( _Inout_ pdo_stmt_t *stmt,
 
             // since the param isn't reliable, we don't do anything here
             case PDO_PARAM_EVT_ALLOC:
-                if ( stmt->supports_placeholders == PDO_PLACEHOLDER_NONE && (param->param_type & PDO_PARAM_INPUT_OUTPUT )) {
-                    sqlsrv_stmt* driver_stmt = reinterpret_cast<sqlsrv_stmt*>( stmt->driver_data );
-                    THROW_PDO_ERROR( driver_stmt, PDO_SQLSRV_ERROR_EMULATE_INOUT_UNSUPPORTED );
+                {
+                    pdo_sqlsrv_stmt* driver_stmt = reinterpret_cast<pdo_sqlsrv_stmt*>(stmt->driver_data);
```
david-puglielli: Is there any particular reason this is being cast to a pdo_sqlsrv_stmt instead of a sqlsrv_stmt? Is direct_query specific to pdo_sqlsrv_stmt?

test/functional/pdo_sqlsrv/pdo_ae_unsupported_stmt_attr.phpt
```diff
@@ -0,0 +1,66 @@
...
+    unset($st);
+    unset($connection);
+} catch (PDOException $e) {
+    var_dump($e->errorInfo);
+}
+
+echo "Done\n";
+?>
+--EXPECT--
+Done
```
david-puglielli: Aren't we expecting some error messages here?

v-kaywon: No, the actual error messages and the expected error messages are compared in the test itself. The test is designed this way because AE on and off behave differently and we want the test to work for both flavours.

david-puglielli: Ah I see how it works now.

PR #584, opened 2017-11-03
Added error handling for LOB types as output parameters

yitam:
#231

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/584)


source/shared/core_stmt.cpp
```diff
@@ -422,6 +422,12 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
 
     // if it's an output parameter and the user asks for a certain type, we have to convert the zval to that type so
     // when the buffer is filled, the type is correct
```
david-puglielli: I think you should amend this comment to explain the error check that you added below. "But first, check that we are not using a LOB type with an output parameter" or something like that.

test/functional/sqlsrv/srv_231_string_truncation_text.phpt
```diff
@@ -0,0 +1,133 @@
...
+function invokeProc($conn, $procName, $k, $direction, $data)
+{
+    $sqlsrvSQLType = sqlsrvSqlType($k, strlen($data));
+    $callArgs = "?, ?";
+
+    // Data to initialize $callResult variable
+    $initData = "ShortString";
+    $callResult = $initData;
+
+    // Make sure not to specify the PHP type
```
david-puglielli: Why must you not specify the php type here?

yitam: It's a copy of another test for issue #231. I can remove the comment if you like. It doesn't matter for this particular test

david-puglielli: Then may as well remove it...

source/shared/core_stmt.cpp
```diff
@@ -420,10 +420,11 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
         }
     }
 
-    // if it's an output parameter and the user asks for a certain type, we have to convert the zval to that type so
-    // when the buffer is filled, the type is correct
-    CHECK_CUSTOM_ERROR( direction != SQL_PARAM_INPUT && (sql_type == SQL_LONGVARCHAR 
-                        || sql_type == SQL_WLONGVARCHAR || sql_type == SQL_LONGVARBINARY), 
+    // If the user specifies a certain type for an output parameter, we have to convert the zval 
+    // to that type so that when the buffer is filled, the type is correct. But first,
+    // should check that if a LOB type is specified.
```
david-puglielli: I think you mean '...should check if a LOB type is specified.'

PR #589, opened 2017-11-06
Fixed crash caused by non-freed reference

david-puglielli:
Fix for #434.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/589)


v-kaywon:
You're not going to include a test for this PR? Perhaps the one that begins a transaction and execute an invalid query. Iterate 20 times.

Hadis-Fard:
should have test for both PDO and SQLSRV to confirm it is fixed in PDO and doesn't exist in SQLSRV

david-puglielli:
Tests that crash because of the bug would still output the expected results/errors because the crash occurs after script execution has finished. Also, a test that crashes can still pass when run under run-tests.php, and there is no crash then. So a test that iterates beginTransaction() and query() 20 times could output whatever we want it to 20 times and then crash at the end, and run-tests.php won't catch it. We may need to use manual verification for this fix in future.

v-kaywon:
Should still include the test for "manual verification" in the future

david-puglielli:
Tests for manual verification that are always skipped when testing automatically on Github? I'll upload tests for that.

Issue #569, opened 2017-10-17
Use sqlsrv_query on encrypted varchar max fields results in function sequence error


yitam:
1. Create a table with an encrypted varchar max column and then insert one row
2. Fetch data by using sqlsrv_query() and then sqlsrv_get_field()
3. Failed with Function sequence error

The problem disappears if using sqlsrv_prepare() in step 2 above.

Hadis-Fard:
Would you please add a repro script?

yitam:
Repro steps:

1. Create a table as follows:
```
CREATE TABLE aTest ( [c1] varchar(max) COLLATE Latin1_General_BIN2 ENCRYPTED WITH (ENCRYPTION_TYPE = deterministic, ALGORITHM = 'AEAD_AES_256_CBC_HMAC_SHA_256', COLUMN_ENCRYPTION_KEY = AEColumnKey) )
```
2. Then insert some large string into this column
```
$value = "some very large string";
$stmt = sqlsrv_prepare($conn, "INSERT INTO aTest (c1) VALUES (?)", array($value));
sqlsrv_execute($stmt);
```
3. Data has been successfully inserted, so fetch data as follows:
```
$stmt = sqlsrv_query($conn, "SELECT * FROM aTest ");
sqlsrv_fetch($stmt);
$fieldVal = sqlsrv_get_field($stmt, 0, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR));
```
4. sqlsrv_get_field() returns `false` so print sqlsrv_errors()

Hadis-Fard:
if it is going to be a limitation, there should be an informative error preventing the calls to ODBC

david-puglielli:
Seems to work if the table is not empty. If the table is empty, I get an error saying `A row must be retrieved with sqlsrv_fetch before retrieving data with sqlsrv_get_field.`, which might be confusing since sqlsrv_fetch *is* called before sqlsrv_get_field, it's just that there is no row to fetch. We might want to clarify the error.

yitam:
@david-puglielli upon more investigation, the error message is not entirely wrong. Based on [sqlsrv_fetch](https://docs.microsoft.com/sql/connect/php/sqlsrv-fetch?view=sql-server-2017), `If there are no more results in the result set, null is returned.`

[...]

PR #590, opened 2017-11-07
Refactor sqlsrv tests with buffer or cursor

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/590)


test/functional/sqlsrv/0052.phpt
```diff
@@ -1,31 +1,32 @@
...
+    $tableName = 'ScrollTest';
 
-    $stmt = sqlsrv_query($conn, "IF OBJECT_ID('ScrollTest', 'U') IS NOT NULL DROP TABLE ScrollTest");
-    if ($stmt !== false) {
-        sqlsrv_free_stmt($stmt);
-    }
+    $columns = array(new AE\ColumnMeta('int', 'id'),
+                     new AE\ColumnMeta('char(10)', 'value'));
+    $stmt = AE\createTable($conn, $tableName, $columns);
+    sqlsrv_free_stmt($stmt);
```
v-kaywon: not going to check if stmt is false?

test/functional/sqlsrv/0052.phpt
```diff
@@ -1,31 +1,32 @@
...
 
-    $stmt = sqlsrv_query($conn, "CREATE TABLE ScrollTest (id int, value char(10))");
-    if ($stmt === false) {
-        die(print_r(sqlsrv_errors(), true));
+    $query = "SELECT * FROM $tableName";
+    if (AE\isColEncrypted()) {
+        $options = array('Scrollable' => SQLSRV_CURSOR_FORWARD);
```
v-kaywon: please include a comment that says Column Encryption does not support scrollable cursor other than SQLSRV_CURSOR_FORWARD

test/functional/sqlsrv/0052.phpt
```diff
@@ -39,8 +40,8 @@ scrollable results with no rows.
     if ($row === false) {
         print_r(sqlsrv_errors(), true);
     }
-
-    $stmt = sqlsrv_query($conn, "SELECT * FROM ScrollTest");
+    
+    $stmt = AE\selectFromTable($conn, $tableName);
```
v-kaywon: Do we need to use a common function for a one liner? I think it's a lot more clear to use `$stmt = sqlsrv_query($conn, "SELECT * FROM ScrollTest");` instead of `$stmt = AE\selectFromTable($conn, $tableName);`. Same with line 29.

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -382,6 +382,36 @@ function createTable($conn, $tbname, $columnMetaArr)
...
 
 /**
+ * Create a table with options for sqlsrv_prepare()
+ * @param  resource  $conn : sqlsrv connection resource
+ * @param  string  $tbname : name of the table to be created
+ * @param  array  $columnMetaArr : array of ColumnMeta objects, which contain metadata for one column
+ * @param  array  $options : array of options for sqlsrv_prepare()
+ * @return resource sqlsrv statement resource
+ */
+function createTableEx($conn, $tbname, $columnMetaArr, $options)
```
v-kaywon: What does createTable do then?

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -475,16 +506,17 @@ function executeQuery($conn, $sql, $conds = null, $values = null)
...
+        }   
+
+        $stmt = sqlsrv_query($conn, $sql, null, $options);
     } else {
         // with AE enabled, use sqlsrv_prepare() in case there are 
         // fields with unlimited size
-        if (empty($conds) || empty($values)) {
-            $stmt = sqlsrv_prepare($conn, $sql);
+        if (empty($conds) || empty($values)) {            
+            $stmt = sqlsrv_prepare($conn, $sql, null, $options);
```
v-kaywon: What happens if conds is empty but values is not empty? Doesn't that mean the parameter's not bound? Also, what if conds is not empty but values is empty, then the WHERE clause cannot be executed.

test/functional/sqlsrv/0052.phpt
yitam: When AE is enabled we have to select using sqlsrv_prepare/execute()

test/functional/sqlsrv/sqlsrv_buffered.phpt
```diff
@@ -1,46 +1,82 @@
...
+                      "Field5" => array("040506", null, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR), SQLSRV_SQLTYPE_VARBINARY('max')),
+                      "Field6" => "This is field 6.",
+                      "Field7" => "This is field 7.");
+        $query = AE\getInsertSqlPlaceholders($tableName, $data);
+
+        // form the array of parameters
+        $params = array();
+        foreach ($data as $key => $input) {
+            array_push($params, $input);
+        }
```
v-kaywon: Tried to use php array function `array_values`

test/functional/sqlsrv/MsHelper.inc
yitam: I added this extra method that takes options plus using sqlsrv_prepare/execute()
But with hindsight I think it's really unnecessary (at least for the test sqlsrv_buffered.phpt)

test/functional/sqlsrv/sqlsrv_buffered.phpt
```diff
@@ -1,46 +1,82 @@
...
-
-sqlsrv_execute($stmt);
+function insertOneRow($conn, $tableName)
+{
+    $result = null;
+    if (AE\isColEncrypted()) {
+        $data = array("Field2" => "This is field 2.",
+                      "Field3" => array("010203", null, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR), SQLSRV_SQLTYPE_IMAGE),
+                      "Field4" => "This is field 4.",
+                      "Field5" => array("040506", null, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR), SQLSRV_SQLTYPE_VARBINARY('max')),
```
v-kaywon: why do we need to explicitly tell the driver what the PHPTYPE and SQLTYPE are?

test/functional/sqlsrv/MsHelper.inc
yitam: Since it's a helper method for the sqlsrv tests and if conds/values aren't right the affected tests will fail. It's up to the caller to ensure the combinations are correct

test/functional/sqlsrv/sqlsrv_fetch_object_unicode_col_name2.phpt
```diff
@@ -1,10 +1,11 @@
 --TEST--
 sqlsrv_fetch_object() into a class with Unicode column name
 --SKIPIF--
+<?php require('skipif_versions_old.inc'); ?>
 --FILE--
 <?php
 
-/* Define the Product class. */
+// Define the Product classes
```
v-kaywon: why classes?

test/functional/sqlsrv/sqlsrv_buffered.phpt
yitam: To bind binary types without explicitly defining both the PHPTYPE and SQLTYPE the binding will fail with different ODBC errors.

test/functional/sqlsrv/sqlsrv_fetch_object_unicode_col_name2.phpt
```diff
@@ -56,72 +57,119 @@ class Sample extends Product
...
+                 new AE\ColumnMeta('VARCHAR(20)', 'Color'));
+AE\createTable($conn, $tableName1, $columns);
 
 // Insert data
-$sql = "INSERT INTO $tableName1 VALUES
-	('P001','Pencil 2B','102','24','0.24','2016-02-01','Red'),
-	('P002','Notepad','102','12','3.87', '2016-02-21',Null),
-	('P001','Mirror 2\"','652','3','15.99', '2016-02-01',NULL),
-	('P003','USB connector','1652','31','9.99','2016-02-01',NULL)";
-sqlsrv_query($conn, $sql) ?: die(print_r(sqlsrv_errors(), true));
```
v-kaywon: what would happen if all insertions are done in one query?

test/functional/sqlsrv/sqlsrv_buffered.phpt
yitam: If the hex value is used, will get `Restricted data type attribute violation` error, even with the correct SQL type. The PHP type SQLSRV_PHPTYPE_INT didn't work either.

test/functional/sqlsrv/sqlsrv_fetch_object_unicode_col_name2.phpt
yitam: Is it possible for binding params?

test/functional/sqlsrv/srv_230_sqlsrv_buffered_numeric_types.phpt
```diff
@@ -3,51 +3,58 @@ Read numeric types from SQLSRV with buffered query.
...
-if ($stmt === false) {
-    die(print_r(sqlsrv_errors(), true));
-}
-
-$query = 'SELECT TOP 2 * FROM #TESTTABLE';
-$stmt = sqlsrv_query($conn, $query, array(), array("Scrollable"=>SQLSRV_CURSOR_CLIENT_BUFFERED));
-if (!$stmt) {
-    echo "Statement could not be prepared.\n";
-    die(print_r(sqlsrv_errors(), true));
-}
```
v-kaywon: why ignore the stmt checking?

test/functional/sqlsrv/sqlsrv_buffered.phpt
yitam: If I use only a character string as input, will get this ODBC error `Implicit conversion from data type varchar to varbinary(max) is not allowed. Use the CONVERT function to run this query.`

yitam: This is a valid alternative to using a character string as input: 

array(hex2bin(0x040506), null, SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY) ),

test/functional/sqlsrv/srv_230_sqlsrv_buffered_numeric_types.phpt
yitam: AE\insertRow already checked $stmt, but I'll double check

test/functional/sqlsrv/0052.phpt
v-kaywon: why is that? you're selecting and you're not selecting a max column

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -211,6 +214,27 @@ function getCekName()
...
 /**
+ * @return the path to the KSP dll/so file
+ */
+function getKSPpath()
+{
+    $name = 'myKSP';
+
+    $dir_name = realpath(dirname(__FILE__));
+    $ksp = $dir_name . DIRECTORY_SEPARATOR . $name;
+    if ( strtoupper( substr( php_uname( 's' ), 0, 3 ) ) == 'WIN' ) {
```
v-kaywon: be consistent with coding standard (i.e., no space after open parenthesis and before close parenthesis)

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -22,6 +22,9 @@ const INSERT_PREPARE = 2;
 const INSERT_QUERY_PARAMS = 3;
 const INSERT_PREPARE_PARAMS = 4;
 
+const KSP_NAME = 'MyCustomKSPName';
+const ENCRYPT_KEY = 'LPKCWVD07N3RG98J0MBLG4H2';
+const KSP_TEST_TABLE = 'CustomKSPTestTable';
```
v-kaywon: I'm not sure if these constants should be here. To me I think MsHelper should only contain helper functions and not any constants. Maybe define them in MsSetup?

test/functional/sqlsrv/sqlsrv_encrypted_patients_ksp.phpt
```diff
@@ -7,202 +7,165 @@ Test simple insert, fetch and update with ColumnEncryption enabled and a custome
...
-    $tablename = CreatePatientsTable();
-    
-    InsertData('748-68-0245', 'Jeannette', 'McDonald', '2002-11-28');
-    InsertData('795-73-9838', 'John', 'Doe', '2001-05-29');
-    InsertData('456-12-5486', 'Jonathan', 'Wong', '1999-12-20');
-    InsertData('156-45-5486', 'Marianne', 'Smith', '1997-03-04'); 
-    
-    SelectData();
+
+    $tableName = CreatePatientsTable();
```
v-kaywon: createPatientsTable

test/functional/sqlsrv/0052.phpt
yitam: I know, but I want all refactored tests to be consistently using AE helper methods to replace the existing sqlsrv_query()

test/functional/sqlsrv/sqlsrv_encrypted_patients_ksp.phpt
```diff
@@ -7,202 +7,165 @@ Test simple insert, fetch and update with ColumnEncryption enabled and a custome
...
 
     ///////////////////////////////////////////
     echo "Update Patient Jonathan Wong...\n";
-    $params = array(array('1999-12-31', null, null, SQLSRV_SQLTYPE_DATE), array('Chang', null, null, SQLSRV_SQLTYPE_NVARCHAR(50)), array('456-12-5486', null, null, SQLSRV_SQLTYPE_CHAR(11)));
-    
+    $params = array(array('1999-12-31', null, null, SQLSRV_SQLTYPE_DATE), 
+                    array('Chang', null, null, SQLSRV_SQLTYPE_NVARCHAR(50)), 
+                    array('456-12-5486', null, null, SQLSRV_SQLTYPE_CHAR(11)));
+
     $tsql = "UPDATE Patients SET BirthDate = ?, LastName = ? WHERE SSN = ?";
```
v-kaywon: since $tableName is already defined, perhaps use it in the query.

test/functional/sqlsrv/MsHelper.inc
yitam: I know I must have missed some... too bad the editor no longer works for me :(

test/functional/sqlsrv/MsHelper.inc
yitam: But they are AE related constants... the variables in MsSetup apply to all tests, right?

test/functional/sqlsrv/0052.phpt
v-kaywon: I'm not too sure about this. That means you'll always be using sqlsrv_prepare and sqlsrv_execute for all queries when isColEncrypted is true even when prepare/execute isn't really required, and sqlsrv_query is never tested (or rarely tested) when column encryption is enabled.

test/functional/sqlsrv/MsHelper.inc
v-kaywon: these constants apply to all tests eventually since all tests are supposed to run in the ksp flavor at the end. This is just a suggestion, you may leave it here if you want.

test/functional/sqlsrv/MsHelper.inc
v-kaywon: :( hope you get it working soon.

test/functional/sqlsrv/MsHelper.inc
v-kaywon: Should still cover those cases (at least with messages saying the user isn't supposed to do this?). Also, put this into the description of this function so the user knows he/she isn't supposed to call this function with this combination of parameters.

test/functional/sqlsrv/sqlsrv_fetch_object_unicode_col_name2.phpt
v-kaywon: I haven't tried, but I would think something like this `$sql = 'INSERT INTO $tableName1 VALUES (?, ?), (?, ?)` should work.

test/functional/sqlsrv/MsHelper.inc
yitam: As I said before, these are AE related constants, so I will keep them here. Those in MsSetup are variables, meant to be changed based on test settings.

test/functional/sqlsrv/sqlsrv_fetch_object_unicode_col_name2.phpt
yitam: But that means I can't reuse the existing helper methods? Doing this or not doesn't change the essence of this particular test, right?

test/functional/sqlsrv/0052.phpt
yitam: Based on issue #569 we are supposed to handle the combination of AE and sqlsrv_query(). Do you want sqlsrv_query() to check if max column is in the query? Or do we simply discourage using sqlsrv_query() when AE is enabled?

test/functional/sqlsrv/MsHelper.inc
yitam: I agree, but there's a valid case when `$conds` is not empty but `$values` is null. For example, the condition is `aCol IS NULL` or `aCol IS NOT NULL`

test/functional/sqlsrv/0052.phpt
yitam: Plus, some existing KSP tests do use sqlsrv_query() to fetch data. I think we need to reconsider how to tackle issue 569.

test/functional/sqlsrv/MsHelper.inc
v-kaywon: Just one last suggestion regarding this. Seems like `$values` is only for placeholders within `$cond`. Can you specify this in the description of the function? e.g., `@param  array  $values : array of parameters for placeholders in $cond, null by default`

test/functional/sqlsrv/sqlsrv_fetch_object_unicode_col_name2.phpt
v-kaywon: I know it's a pain, but this seems to be the only test that does this (I haven't encounter another test that inserts multiple rows in one query). That's why I think it would be valuable to keep it.

test/functional/sqlsrv/sqlsrv_buffered_result_set_extended_ascii.phpt
```diff
@@ -1,40 +1,33 @@
...
+$res = null;
+$stmt = AE\insertRow($conn, $tableName, array('ID' => 'Aå_Ð×Æ×Ø_B'));
 
 // Fetch data
 $query = "SELECT * FROM $tableName";
-$stmt = sqlsrv_query($conn, $query, [], array("Scrollable"=>"buffered"));
-if (! $stmt) {
-    die(print_r(sqlsrv_errors(), true));
-}
+$stmt = AE\executeQueryEx($conn, $query, array("Scrollable"=>"buffered"));
```
v-kaywon: please use sqlsrv_query

test/functional/sqlsrv/sqlsrv_num_rows.phpt
```diff
@@ -1,30 +1,30 @@
...
+    
+    // Always Encrypted feature only supports SQLSRV_CURSOR_FORWARD or 
+    // SQLSRV_CURSOR_CLIENT_BUFFERED
+    if (AE\isColEncrypted()) {
+        $options = array('Scrollable' => SQLSRV_CURSOR_CLIENT_BUFFERED);
+    } else {
+        $options = array('Scrollable' => SQLSRV_CURSOR_KEYSET);
     }
-    $stmt = sqlsrv_query($conn, "SELECT * FROM utf16invalid", array(), array("Scrollable" => SQLSRV_CURSOR_KEYSET ));
+    $stmt = AE\executeQueryEx($conn, "SELECT * FROM $tableName", $options);
```
v-kaywon: please use sqlsrv_query

test/functional/sqlsrv/srv_012_sqlsrv_fetch_array.phpt
```diff
@@ -1,30 +1,27 @@
...
+AE\insertRow($conn, $tableName, array("ID" => '1998.1'));
+AE\insertRow($conn, $tableName, array("ID" => '-2004'));
+AE\insertRow($conn, $tableName, array("ID" => '2016'));
+AE\insertRow($conn, $tableName, array("ID" => '4.2EUR'));
 
 // Fetch data
 $query = "SELECT ID FROM $tableName";
-$stmt = sqlsrv_query($conn, $query, [], array("Scrollable"=>"buffered"))
-    ?: die(print_r(sqlsrv_errors(), true));
+$stmt = AE\executeQueryEx($conn, $query, array("Scrollable"=>"buffered"));
```
v-kaywon: please use sqlsrv_query

test/functional/sqlsrv/srv_037_sqlsrv_has_rows.phpt
```diff
@@ -43,8 +40,7 @@ if (sqlsrv_has_rows($stmt)) {
 }
 
 // Fetch data using a scrollable cursor
-$stmt = sqlsrv_query($conn, $query, [], array("Scrollable"=>"buffered"))
-        ?: die(print_r(sqlsrv_errors(), true));
+$stmt = AE\executeQueryEx($conn, $query, array("Scrollable"=>"buffered"));
```
v-kaywon: please use sqlsrv_query

test/functional/sqlsrv/srv_230_sqlsrv_buffered_numeric_types.phpt
```diff
@@ -3,51 +3,64 @@ Read numeric types from SQLSRV with buffered query.
...
+$stmt = AE\insertRow($conn, $tableName, $data, $res, AE\INSERT_QUERY_PARAMS);
 if (!$stmt) {
-    echo "Statement could not be prepared.\n";
-    die(print_r(sqlsrv_errors(), true));
+    fatalError("Failed to insert into $tableName!");
 }
-sqlsrv_execute($stmt);
+
+$query = "SELECT TOP 2 * FROM $tableName";
+$stmt = AE\executeQueryEx($conn, $query, array("Scrollable"=>SQLSRV_CURSOR_CLIENT_BUFFERED));
```
v-kaywon: please use sqlsrv_query

test/functional/sqlsrv/test_scrollable.phpt
```diff
@@ -1,249 +1,210 @@
...
-$stmt = sqlsrv_query( $conn, "SELECT * FROM ScrollTest", array(), array( "Scrollable" => 'dynamic' ));
 
-sqlsrv_fetch( $stmt );
-sqlsrv_fetch( $stmt );
+// Always Encrypted feature only supports SQLSRV_CURSOR_FORWARD, so skip the rest of the test
+// when AE is enabled
+// https://github.com/Microsoft/msphpsql/wiki/Features#aelimitation
+$query = "SELECT * FROM $tableName";
+$options = array('Scrollable' => SQLSRV_CURSOR_FORWARD);
+$stmt = AE\executeQueryEx($conn, $query, $options);
```
v-kaywon: please use sqlsrv_query

Issue #581, opened 2017-10-29
sqlsrv_next_result() causes "The active result for the query contains no fields" error in v5.1.1


theodorejb:
I have a generic class for executing queries and returning selected results. It worked fine with v5.1.0 and previous SQLSRV driver releases, but with v5.1.1 I am getting an error saying "The active result for the query contains no fields."

Before running the test script, create a `Users` table like this:

``` sql
CREATE TABLE Users (
    user_id int IDENTITY PRIMARY KEY,
    name varchar(50) NOT NULL
);
```

Here's the code that demonstrates the problem. In v5.1.0 and prior, the call to `sqlsrv_next_result()` for the second query returned `null`, but in v5.1.1 it returns `false`.

``` php
$connection = sqlsrv_connect('Computer-Name\SQLEXPRESS', ['Database' => 'MyDatabase']);

// Still works
$result = query("INSERT INTO Users (name) VALUES (?); SELECT SCOPE_IDENTITY() AS RowID;", ['Michael']);
$affected = $result->getAffected(); // 1
$id = $result->getFirst()['RowID']; // e.g. 1

// Worked in v5.1.0 and previous releases, but in v5.1.1 causes the following error:
// "Failed to get next result: The active result for the query contains no fields."
$result2 = query("UPDATE Users SET name = ? WHERE user_id = ?", ['Christopher', 0]);
$affected2 = $result2->getAffected(); // should be 0

function query(string $sql, array $params): Statement
{
    global $connection;
    $stmt = sqlsrv_query($connection, $sql, $params);

    if ($stmt === false) {
        throw new Exception('Query failed: ' . sqlsrv_errors()[0]['message']);
    }

    $result = new Statement($stmt);
    $result->execute();
    return $result;
}

class Statement
{
    /** @var resource */
    private $stmt;
    private $affected;

    public function __construct($stmt)
    {
        $this->stmt = $stmt;
    }

    public function execute()
    {
        $next = null;
        $this->affected = -1;

        // get affected row count from results prior to select result
        do {
            $affectedRows = sqlsrv_rows_affected($this->stmt);

            if ($affectedRows === false) {
                throw new Exception('Failed to get affected row count: '.  sqlsrv_errors()[0]['message']);
            }

            if ($affectedRows === -1) {
                // reached SELECT result
                break; // so that getIterator will be able to select the rows
            }

            if ($this->affected === -1) {
                $this->affected = 0;
            }

            $this->affected += $affectedRows;
[...]

Issue #580, opened 2017-10-27
Fetching from an encrypted max column after calling PDOstatement::nextRowset() returns some garage characters


v-kaywon:
After calling PDOstatement::nextRowset(), fetching from an encrypted max column returns some garbage characters. Following is a repro script:
```
<?php
try {
    $conn = new PDO("sqlsrv:server = $server; Database = $databaseName; ColumnEncryption = Enabled", $uid, $pwd);
    $tbname = "testNVarChar";
    $createSql = "CREATE TABLE $tbname
                  ([intCol] int ENCRYPTED WITH
                    (ENCRYPTION_TYPE = deterministic,
                     ALGORITHM = 'AEAD_AES_256_CBC_HMAC_SHA_256',
                     COLUMN_ENCRYPTION_KEY = AEColumnKey),
                   [nVarCharCol] nvarchar(max) COLLATE Latin1_General_BIN2 ENCRYPTED WITH
                    (ENCRYPTION_TYPE = deterministic,
                     ALGORITHM = 'AEAD_AES_256_CBC_HMAC_SHA_256',
                     COLUMN_ENCRYPTION_KEY = AEColumnKey))";
    $conn->exec($createSql);
    $intData = 1;
    $nvarcharData = "This is a really large string used to test certain large data types like xml data type. The length of this string is greater than 256 to correctly test a large data type. This is currently used by atleast varchar type and by xml type. The fetch tests are the primary consumer of this string to validate that fetch on large types work fine. The length of this string as counted in terms of number of characters is 417.";
    $insertSql = "INSERT INTO $tbname VALUES (?, ?)";
    $stmt = $conn->prepare($insertSql);
    $stmt->execute(array($intData, $nvarcharData));
    $selectSql = "SELECT intCol FROM $tbname; SELECT nVarCharCol FROM $tbname";
    $stmt = $conn->query($selectSql);

    $rowset = $stmt->fetch(PDO::FETCH_ASSOC);
    var_dump($rowset);
    $stmt->nextRowset();
    $rowset2 = $stmt->fetch(PDO::FETCH_ASSOC);
    var_dump($rowset2);

    $conn->exec("DROP TABLE $tbname");
    unset($stmt);
    unset($conn);
} catch (PDOException $e) {
    var_dump($e);
}
?>
```

Without column encryption, the proper data is returned. With column encryption however, the following is returned:
```
array(1) {
  [0]=>
  array(1) {
    ["intCol"]=>
    string(1) "1"
[...]

Issue #574, opened 2017-10-23
sqlsrv_next_result() returns incorrect result with varchar(max) field when Always Encrypted is enabled


yitam:
To reproduce this, connect with Always Encrypted enabled, create a table with only one `varchar(max)` column and then fetch the result by running two selects in a row. This is an example (the problem is gone when the column type is `varchar(1024)`):

```
$tableName = 'TextTable';
$sql = "create table $tableName (col1 varchar(max) COLLATE Latin1_General_BIN2 ENCRYPTED WITH (ENCRYPTION_TYPE = deterministic, ALGORITHM = 'AEAD_AES_256_CBC_HMAC_SHA_256', COLUMN_ENCRYPTION_KEY = AEColumnKey) )";

$stmt = sqlsrv_query($conn, $sql);

$sql = "insert into $tableName (col1) VALUES (?)";
$params = array("This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test This is a test ");

$stmt = sqlsrv_prepare($conn, $sql, $params);
if ($stmt) {
    $r = sqlsrv_execute($stmt);
    if (!$r) {
        print_r(sqlsrv_errors());
    }
}

$stmt = sqlsrv_prepare($conn, "SELECT * FROM [$tableName]; SELECT * FROM [$tableName]");
if ($stmt) {
    $r = sqlsrv_execute($stmt);
    if (!$r) {
        print_r(sqlsrv_errors());
    }
}

echo "1. next result? ";
$next = sqlsrv_next_result($stmt);
var_dump($next);

$row = sqlsrv_fetch($stmt);
if ($row) {
    $fld = sqlsrv_get_field($stmt, 0, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR));
    echo(substr($fld, 0, 15)) . PHP_EOL;
}

echo "2. next result? ";
$next = sqlsrv_next_result($stmt);
var_dump($next);

echo "3. next result? ";
$next = sqlsrv_next_result($stmt);
var_dump($next);
```

The expected output is 
```
1. next result? bool(true)
[...]

PR #592, opened 2017-11-09
Fix to GitHub issues 574 and 580

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/592)

### DIFF
--- a/source/shared/core_results.cpp
+++ b/source/shared/core_results.cpp
@@ -45,7 +45,7 @@ namespace {
 
 // *** internal constants ***
 
-const int INITIAL_FIELD_STRING_LEN = 1024;      // base allocation size when retrieving a string field
+const int INITIAL_FIELD_STRING_LEN = 8000;      // base allocation size when retrieving a string field
 
 // *** internal functions ***
 
--- a/source/shared/core_results.cpp
+++ b/source/shared/core_results.cpp
@@ -45,7 +45,7 @@ namespace {
 
 // *** internal constants ***
 
-const int INITIAL_FIELD_STRING_LEN = 8000;      // base allocation size when retrieving a string field
+const int INITIAL_LOB_FIELD_LEN = 2048;      // base allocation size when retrieving a LOB field
 
 // *** internal functions ***
 
@@ -1520,9 +1520,9 @@ SQLPOINTER read_lob_field( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT field_in
     }
 
     SQLLEN already_read = 0;
-    SQLLEN to_read = INITIAL_FIELD_STRING_LEN;
+    SQLLEN to_read = INITIAL_LOB_FIELD_LEN;
     sqlsrv_malloc_auto_ptr<char> buffer;
-    buffer = static_cast<char*>( sqlsrv_malloc( INITIAL_FIELD_STRING_LEN + extra + sizeof( SQLULEN )));
+    buffer = static_cast<char*>( sqlsrv_malloc( INITIAL_LOB_FIELD_LEN + extra + sizeof( SQLULEN )));
     SQLRETURN r = SQL_SUCCESS;
     SQLCHAR state[ SQL_SQLSTATE_BUFSIZE ];
     SQLLEN last_field_len = 0;
--- a/source/shared/core_results.cpp
+++ b/source/shared/core_results.cpp
@@ -45,7 +45,7 @@ namespace {
 
 // *** internal constants ***
 
-const int INITIAL_FIELD_STRING_LEN = 256;    // base allocation size when retrieving a string field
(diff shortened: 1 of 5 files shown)


source/shared/core_results.cpp
```diff
@@ -45,7 +45,7 @@ namespace {
 
 // *** internal constants ***
 
-const int INITIAL_FIELD_STRING_LEN = 256;    // base allocation size when retrieving a string field
+const int INITIAL_FIELD_STRING_LEN = 1024;      // base allocation size when retrieving a string field
```
v-kaywon: can you only increase the initial buffer size when column encryption is enabled? Also, if column encryption is enabled, I would love to blast the initial buffer size to 8000 (max size for varchar(n))

test/functional/pdo_sqlsrv/pdo_574_next_rowset.phpt
```diff
@@ -0,0 +1,154 @@
...
+    unset($stmt);
+
+    // run queries in a batch
+    $stmt = $conn->prepare("SELECT * FROM [$tableName]; SELECT artist FROM [cd_info]; SELECT * FROM [$tableName1]");
+    $stmt->execute();
+    
+    // fetch from $tableName
+    $row = $stmt->fetch(PDO::FETCH_NUM);
+    if ($row) {
+        echo(substr($row[0], 0, 15)) . PHP_EOL;
```
v-kaywon: this only echoes the first 10 characters, what if something has gone wrong in between? Would be great of you can compare it with `$phrase` instead

test/functional/pdo_sqlsrv/pdo_574_next_rowset.phpt
```diff
@@ -0,0 +1,154 @@
...
+
+    // create two tables with max fields
+    $columns = array(new ColumnMeta('varchar(max)', 'col1'));
+    createTable($conn, $tableName, $columns);
+
+    $columns = array(new ColumnMeta('varchar(max)', 'col1'));
+    createTable($conn, $tableName1, $columns);
+
+    // insert one row to each table
+    $phrase = str_repeat('This is a test ', 250);
```
v-kaywon: if you're going to change the initial buffer size, this number will have to change as well

test/functional/pdo_sqlsrv/pdo_574_next_rowset.phpt
```diff
@@ -0,0 +1,154 @@
...
+    $row = $stmt->fetch(PDO::FETCH_NUM);
+    if ($row) {
+        fatalError("This is unexpected!\n");
+    }
+
+    echo "4. next result? " . PHP_EOL;
+    try {
+        $next = $stmt->nextRowset();
+    } catch (PDOException $e) {
+        echo $e->getMessage() . PHP_EOL;
```
v-kaywon: This is a regression in 5.1.1-preview (see https://github.com/Microsoft/msphpsql/issues/581). newRowset() should not return any error. Please use 5.1.0-preview for testing and see what the behavior should be. Oh.. that means your test will fail against the dev branch for now... maybe talk to David to see when he'll revert the change and merge it with dev?

test/functional/sqlsrv/sqlsrv_574_next_result.phpt
```diff
@@ -0,0 +1,186 @@
+--TEST--
+GitHub issue 574 - Fetch Next Result Test
```
v-kaywon: please change this test as well according to the comments in the PDO_SQLSRV equivalent test

source/shared/core_results.cpp
yitam: I think it's reasonable to simply use 8000 regardless of AE setting. We use this buffer size for varchar / char parameters with unknown size.

yitam:
#574 and #580

source/shared/core_results.cpp
```diff
@@ -45,7 +45,7 @@ namespace {
 
 // *** internal constants ***
 
-const int INITIAL_FIELD_STRING_LEN = 1024;      // base allocation size when retrieving a string field
+const int INITIAL_FIELD_STRING_LEN = 8000;      // base allocation size when retrieving a string field
```
v-kaywon: Now that I think about it, changing this value for non-AE may cause a breaking change. Say a user is trying to fetch a string that is 5000 bytes long. Before this change, the user needs to call sqlsrv_next_result + sqlsrv_get_field many times to retrieve the complete string. With this change however, the user only needs to call sqlsrv_get_field once the retrieve the entire string. This seems like a good thing. But if the user has multiple sqlsrv_next_result + sqlsrv_get_field in their code, error will be thrown telling the user there's no more result left. Therefore, I think it's best to only change the buffer size for if column encryption is enabled. It's impossible to change the constant here though, you may have to go through the code, check where `INITIAL_FIELD_STRING_LEN` is used and change it there.

PR #594, opened 2017-11-16
Adjust precision for decimal input when column encryption is enabled

v-kaywon:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/594)


v-kaywon:
I'll add more tests for testing scientific notation later.

source/shared/core_stmt.cpp
```diff
@@ -2682,6 +2661,129 @@ bool reset_ae_stream_cursor( _Inout_ sqlsrv_stmt* stmt ) {
...
+        bool isNegExp = false;
+        if( *ptr == '+' || *ptr == '-' ){
+            if( *ptr == '-' ){
+                isNegExp = true;
+            }
+            ptr++;
+        }
+        while( isdigit( *ptr )){
+            scientificExp = scientificExp * 10 + *ptr - '0';
+            ptr++;
```
yitam: suggestion: put parentheses around *ptr - '0'

source/shared/core_stmt.cpp
```diff
@@ -2682,6 +2661,129 @@ bool reset_ae_stream_cursor( _Inout_ sqlsrv_stmt* stmt ) {
...
+        }
+        SQLSRV_ASSERT( scientificExp <=38, "Input decimal overflow: sql decimal type only supports up to a precision of 38." );
+        if( isNegExp ){
+            scientificExp = scientificExp * -1;
+        }
+    }
+    SQLSRV_ASSERT( *ptr == '\0', "Invalid input decimal: Invalid character found in the decimal string." );
+    // if number of decimal is less than the exponent, that means the number is a whole number, so no need to adjust the precision
+    if(( int )numDec > scientificExp ){
+        int decToRemove = numDec - scientificExp - decimal_digits;
```
yitam: why don't you simply declare numDec as `int`? Probably  `short`  is more than enough too

source/shared/core_stmt.cpp
```diff
@@ -2682,6 +2661,129 @@ bool reset_ae_stream_cursor( _Inout_ sqlsrv_stmt* stmt ) {
...
+        if( decToRemove > 0 ){
+            bool carryOver = false;
+            int backInd = 0;
+            // pop digits from the vector until there is only 1 more decimal place than required decimal_digits
+            while( decToRemove != 1 && !digits.empty() ){
+                digits.pop_back();
+                decToRemove--;
+            }
+            if( !digits.empty() ){
+                // check if the last digit to be popped is greater than 5, if so, the digit before is needs to round up
```
yitam: Why don't you use some variables up there to remember the exact positions where to pop or carry over?

source/shared/core_stmt.cpp
```diff
@@ -2682,6 +2661,129 @@ bool reset_ae_stream_cursor( _Inout_ sqlsrv_stmt* stmt ) {
...
+    size_t maxDecimalLen = 38 + 6;
+    SQLSRV_ASSERT( Z_STRLEN_P( param_z ) < maxDecimalLen, "Input decimal overflow: sql decimal type only supports up to a precision of 38." );
+    std::vector<size_t> digits;
+    char* ptr = ZSTR_VAL( Z_STR_P( param_z ));
+    bool isNeg = false;
+    char scientificChar = ' ';
+    int scientificExp = 0;
+    // parse digits in param_z into the vector digits
+    if( *ptr == '+' || *ptr == '-' ){
+        if( *ptr = '-' ){
```
david-puglielli: Should this be == ?

source/shared/core_stmt.cpp
v-kaywon: can you be a bit more specified? do you mean to have booleans to keep track of which number in digits are greater than 5? I don't want to do that since we'll have to keep track of another vector of booleans, and by then you still need to ask if the boolean at that position is true or not, which isn't that much faster than after if the digit itself is greater than or equals to 5.

yitam: Putting scientific notation aside, while parsing a number with decimals, you already know the precision you want, so you can keep track of how many decimals you have parsed so far and decided whether to continue parsing or not. If not, right then check if you need to round up the the digit that is last scanned.

david-puglielli:
*[source/shared/core_stmt.cpp, line 2689 at r6](https://reviewable.io:443/reviews/microsoft/msphpsql/594#-KzAyIGHKDUPI6FpMbhC:-KzAyIGIJpK8t_WFJpcG:bd5qsdj) ([raw file](https://github.com/microsoft/msphpsql/blob/fc7c4fb89bbf1dfeee18588085cef20b68f779df/source/shared/core_stmt.cpp#L2689)):*
> ```cpp
>     short numDec = 0;
>     while( isdigit( *ptr )){
>         digits.push_back( *ptr - '0' );
> ```

Is the purpose of subtracting '0' to convert the pointer to an int?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/msphpsql/594)*

david-puglielli:
*[test/functional/pdo_sqlsrv/pdo_ae_insert_decimal.phpt, line 11 at r3](https://reviewable.io:443/reviews/microsoft/msphpsql/594#-KzB-gryMXUAuxZYRB_9:-KzB-gryMXUAuxZYRB_A:b-nb8co9) ([raw file](https://github.com/microsoft/msphpsql/blob/4fb2c80e697a1eaafd362143fbbcab7cdac10601/test/functional/pdo_sqlsrv/pdo_ae_insert_decimal.phpt#L11)):*
> ```PHP
> $num = array("-10.0", "13.33", "-191.78464696202265", "833.33", "-850.00000000000006", "851.64835164835168", "-316053.16053160531", "505505.5055055055055055055", "-1535020.0615", "7501300.675013006750130067501300675", "-7540010.067540010067540010067540010067", "7540450.0675404500675404500675404500", "-820012820012820.01282001282001282001282", "1122551511225515.1122", "-1234567892112345678912.3456", "123456789012346261234567890123.4629", "-13775323913775323913775323913775323913");
> $frac = array("-0.000", "0.100", "-0.1333", "0.019178464696202265", "-0.083333", "0.085000000000000006", "-0.085164835164835168", "0.0000316", "-0.00005", "0.0000153502", "-0.0000075013", "0.00000754001", "-0.00000754045", "0.000000000008200", "-0.00000000000000112255", "0.00000000000000000000123456789", "-0.00000000000000000000000123456789012346", "0.00000000000000000000000000000001377532");
> $numSets = array("Testing numbers greater than 1:" => $num,
> ```

Strictly speaking, a negative number is less than 1

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/msphpsql/594)*

david-puglielli:
*[source/shared/core_stmt.cpp, line 2737 at r6](https://reviewable.io:443/reviews/microsoft/msphpsql/594#-KzB2uLx1LShLonF3ePO:-KzB2uLx1LShLonF3ePP:b-j4gaj7) ([raw file](https://github.com/microsoft/msphpsql/blob/fc7c4fb89bbf1dfeee18588085cef20b68f779df/source/shared/core_stmt.cpp#L2737)):*
> ```cpp
>             if( !digits.empty() ){
>                 // check if the last digit to be popped is greater than 5, if so, the digit before it needs to round up
>                 carryOver = digits.back() >= 5;
> ```

Is it correct to always round up when the last digit is 5? In some conventions we round down (only an issue if the very last nonzero digit is 5 though)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/msphpsql/594)*

v-kaywon:
*[source/shared/core_stmt.cpp, line 2689 at r6](https://reviewable.io:443/reviews/microsoft/msphpsql/594#-KzAyIGHKDUPI6FpMbhC:-KzBZIojtKcP2wCVyfuK:b2nign6) ([raw file](https://github.com/microsoft/msphpsql/blob/fc7c4fb89bbf1dfeee18588085cef20b68f779df/source/shared/core_stmt.cpp#L2689)):*
<details><summary><i>Previously, david-puglielli (David Puglielli) wrote…</i></summary><blockquote>

Is the purpose of subtracting '0' to convert the pointer to an int?
</blockquote></details>

yes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/msphpsql/594)*

v-kaywon:
*[source/shared/core_stmt.cpp, line 2731 at r3](https://reviewable.io:443/reviews/microsoft/msphpsql/594#-Kz1GEZmydmgOA9fpN8L-r3-2731:-KzBZU54tGzgX5VkLq6k:bz1uadt) ([raw file](https://github.com/microsoft/msphpsql/blob/4fb2c80e697a1eaafd362143fbbcab7cdac10601/source/shared/core_stmt.cpp#L2731)):*

Putting scientific notation aside, while parsing a number with decimals, you already know the precision you want, so you can keep track of how many decimals you have parsed so far and decided whether to continue parsing or not. If not, right then check if you need to round up the digit that is last scanned. 
</blockquote></details>

but we don't know if it's in scientific notation or not when we first start parsing

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/msphpsql/594)*

v-kaywon:
*[source/shared/core_stmt.cpp, line 2737 at r6](https://reviewable.io:443/reviews/microsoft/msphpsql/594#-KzB2uLx1LShLonF3ePO:-KzBb2FQ6gt9GQPeH4bj:bl93fnh) ([raw file](https://github.com/microsoft/msphpsql/blob/fc7c4fb89bbf1dfeee18588085cef20b68f779df/source/shared/core_stmt.cpp#L2737)):*
<details><summary><i>Previously, david-puglielli (David Puglielli) wrote…</i></summary><blockquote>

Is it correct to always round up when the last digit is 5? In some conventions we round down (only an issue if the very last nonzero digit is 5 though)
</blockquote></details>

After some research, SQL Server always rounds up unless the user calls the ROUND() function

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/msphpsql/594)*

source/shared/core_stmt.cpp
```diff
@@ -19,6 +19,10 @@
 
 #include "core_sqlsrv.h"
 
+#include <string>
```
david-puglielli: <string> is already #included in core_sqlsrv.h.

source/shared/core_stmt.cpp
```diff
@@ -2662,25 +2661,29 @@ bool reset_ae_stream_cursor( _Inout_ sqlsrv_stmt* stmt ) {
...
     char* ptr = ZSTR_VAL( Z_STR_P( param_z ));
     bool isNeg = false;
     char scientificChar = ' ';
-    int scientificExp = 0;
+    short scientificExp = 0;
     // parse digits in param_z into the vector digits
     if( *ptr == '+' || *ptr == '-' ){
-        if( *ptr = '-' ){
+        if( *ptr == '-' ){
             isNeg = true;
```
yitam: this seems to suggest you have assumed no space characters before the positive / negative sign?

source/shared/core_stmt.cpp
```diff
@@ -526,39 +530,13 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
             break;
         case IS_STRING:
             {
+                if ( sql_type == SQL_DECIMAL || sql_type == SQL_NUMERIC ) {
+                    adjustInputPrecision( param_z, decimal_digits );
```
yitam: Actually do we have to adjust input precision when AE is not enabled?
Besides, what's the likelihood the user will input a string of scientific notation for a decimal/numeric type?

source/shared/core_stmt.cpp
```diff
@@ -2682,6 +2660,134 @@ bool reset_ae_stream_cursor( _Inout_ sqlsrv_stmt* stmt ) {
...
+    // if number of decimal is less than the exponent, that means the number is a whole number, so no need to adjust the precision
+    if(*ptr == '\0' && ( int )numDec > scientificExp ){
+        int decToRemove = numDec - scientificExp - decimal_digits;
+        if( decToRemove > 0 ){
+            bool carryOver = false;
+            short backInd = 0;
+            // pop digits from the vector until there is only 1 more decimal place than required decimal_digits
+            while( decToRemove != 1 && !digits.empty() ){
+                digits.pop_back();
+                decToRemove--;
```
yitam: I still think we should not continue parsing the input string when some digits are meant to be dropped (popped) later.

source/shared/core_stmt.cpp
v-kaywon: yes, if the user specifies the sqltype (only the case for SQLSRV driver) as SQL_DECIMAL or SQL_NUMERIC, the input passed into SQLBindParam requires matching scale (i.e., number of decimal digits).

I don't know what is the likelihood of user inputting a stringified scientific notation, but it's better to have this case covered.

source/shared/core_stmt.cpp
v-kaywon: let me look up the performance of finding a character ('e') in a string and get back to you.

PR #595, opened 2017-11-16
Revert nextRowset changes

david-puglielli:
Reverting changes because several users have run into errors under the changes made to the way empty and null results are handled, and because the changes caused problems with multi-query statements. The original approach generally produced error messages only upon fetching, and next_result or nextRowset should simply move the cursor.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/595)


yitam:
#581

PR #596, opened 2017-11-16
Modified how to bind parameters when AE is enabled

yitam:
Should get meta data for all parameters before binding any of them.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/596)

### DIFF
--- a/source/shared/core_stmt.cpp
+++ b/source/shared/core_stmt.cpp
@@ -465,9 +465,10 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
 
     if( stmt->conn->ce_option.enabled && ( sql_type == SQL_UNKNOWN_TYPE || column_size == SQLSRV_UNKNOWN_SIZE )){
         // use the meta data only if the user has not specified the sql type or column size
-        sql_type = stmt->param_descriptions[param_num].sql_type;
-        column_size = stmt->param_descriptions[param_num].column_size;
-        decimal_digits = stmt->param_descriptions[param_num].decimal_digits;
+        SQLSRV_ASSERT( param_num < stmt->param_descriptions.size(), "Invalid param_num passed in core_sqlsrv_bind_param!" );
+        sql_type = stmt->param_descriptions[param_num].get_sql_type();
+        column_size = stmt->param_descriptions[param_num].get_column_size();
+        decimal_digits = stmt->param_descriptions[param_num].get_decimal_digits();
 
         // change long to double if the sql type is decimal
         if(( sql_type == SQL_DECIMAL || sql_type == SQL_NUMERIC ) && Z_TYPE_P(param_z) == IS_LONG )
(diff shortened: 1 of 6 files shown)


test/functional/sqlsrv/sqlsrv_buffered.phpt
```diff
@@ -12,9 +12,9 @@ function insertOneRow($conn, $tableName)
     $result = null;
     if (AE\isColEncrypted()) {
         $data = array("Field2" => "This is field 2.",
-                      "Field3" => array("010203", null, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR), SQLSRV_SQLTYPE_IMAGE),
+                      "Field3" => "010203",
                       "Field4" => "This is field 4.",
-                      "Field5" => array("040506", null, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR), SQLSRV_SQLTYPE_VARBINARY('max')),
+                      "Field5" => "040506",
```
david-puglielli: Why have these values been changed?

yitam: They were the workarounds added while refactoring the tests to bind input parameters. With this fix the workarounds are no longer needed.

source/shared/core_stmt.cpp
```diff
@@ -466,7 +464,11 @@ void core_sqlsrv_bind_param( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT param_
                     encoding == SQLSRV_ENCODING_BINARY ), "core_sqlsrv_bind_param: invalid encoding" );
 
     if( stmt->conn->ce_option.enabled && ( sql_type == SQL_UNKNOWN_TYPE || column_size == SQLSRV_UNKNOWN_SIZE )){
-        ae_get_sql_type_info( stmt, param_num, direction, param_z, encoding, sql_type, column_size, decimal_digits TSRMLS_CC );
+        // use the meta data only if the user has not specified the sql type or column size
+        sql_type = stmt->param_descriptions[param_num].sql_type;
+        column_size = stmt->param_descriptions[param_num].column_size;
+        decimal_digits = stmt->param_descriptions[param_num].decimal_digits;
+
```
v-kaywon: perhaps put this into a member function of param_meta_data?

v-kaywon: also, in the member function, check if any of this.sql_type, this.column_size, this.decimal_digits are null. Throw error if any of them is null.

yitam: I can put these into a member function but I don't see why these fields might be null. If anything goes wrong calling SQLDescribeParam would have thrown an exception already

v-kaywon: Putting null checks in is to make sure the order of execution is correct (i.e., make sure SQLDescribeParam is called before core_sqlsrv_bind_param).

yitam: I can check whether `stmt->param_descriptions[param_num]` exists. If it is, its members should be correct

v-kaywon: OK. Please check if stmt->param_descriptions vector size is less than param_num to check the existance of pram_descriptions[param_num]

PR #598, opened 2017-11-16
support for non-utf8 locales

lilgreenbird:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/598)

### DIFF
--- a/source/shared/core_init.cpp
+++ b/source/shared/core_init.cpp
@@ -39,7 +39,11 @@ void core_sqlsrv_minit( _Outptr_ sqlsrv_context** henv_cp, _Inout_ sqlsrv_contex
     SQLSRV_STATIC_ASSERT( sizeof( sqlsrv_sqltype ) == sizeof( zend_long ) );
     SQLSRV_STATIC_ASSERT( sizeof( sqlsrv_phptype ) == sizeof( zend_long ));
 
+#ifndef _WIN32
+    // set locale from environment
+    // this is necessary for ODBC and MUST be done before connection
     setlocale(LC_ALL, "");
+#endif
 
     *henv_cp = *henv_ncp = SQL_NULL_HANDLE; // initialize return values to NULL
 
--- a/source/shared/core_init.cpp
+++ b/source/shared/core_init.cpp
@@ -39,6 +39,8 @@ void core_sqlsrv_minit( _Outptr_ sqlsrv_context** henv_cp, _Inout_ sqlsrv_contex
     SQLSRV_STATIC_ASSERT( sizeof( sqlsrv_sqltype ) == sizeof( zend_long ) );
     SQLSRV_STATIC_ASSERT( sizeof( sqlsrv_phptype ) == sizeof( zend_long ));
 
+    setlocale(LC_ALL, "");
+
     *henv_cp = *henv_ncp = SQL_NULL_HANDLE; // initialize return values to NULL
 
     try {
--- a/source/shared/localizationimpl.cpp
+++ b/source/shared/localizationimpl.cpp
@@ -1,5 +1,5 @@
 //---------------------------------------------------------------------------------------------------------------------------------
-// File: LocalizationImpl.hpp
+// File: localizationimpl.cpp
 //
 // Contents: Contains non-inline code for the SystemLocale class
 //           Must be included in one c/cpp file per binary
@@ -71,6 +71,17 @@ const cp_iconv cp_iconv::g_cp_iconv[] = {
     {  1256, "CP1256//TRANSLIT" },
     {  1257, "CP1257//TRANSLIT" },
     {  1258, "CP1258//TRANSLIT" },
+    { CP_ISO8859_1, "ISO8859-1//TRANSLIT" },
+    { CP_ISO8859_2, "ISO8859-2//TRANSLIT" },
+    { CP_ISO8859_3, "ISO8859-3//TRANSLIT" },
+    { CP_ISO8859_4, "ISO8859-4//TRANSLIT" },
+    { CP_ISO8859_5, "ISO8859-5//TRANSLIT" },
+    { CP_ISO8859_6, "ISO8859-6//TRANSLIT" },
+    { CP_ISO8859_7, "ISO8859-7//TRANSLIT" },
+    { CP_ISO8859_8, "ISO8859-8//TRANSLIT" },
+    { CP_ISO8859_9, "ISO8859-9//TRANSLIT" },
+    { CP_ISO8859_13, "ISO8859-13//TRANSLIT" },
+    { CP_ISO8859_15, "ISO8859-15//TRANSLIT" },
     { 12000, "UTF-32LE" }
 };
 const size_t cp_iconv::g_cp_iconv_count = ARRAYSIZE(cp_iconv::g_cp_iconv);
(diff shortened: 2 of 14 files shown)


source/shared/core_init.cpp
```diff
@@ -39,6 +39,8 @@ void core_sqlsrv_minit( _Outptr_ sqlsrv_context** henv_cp, _Inout_ sqlsrv_contex
     SQLSRV_STATIC_ASSERT( sizeof( sqlsrv_sqltype ) == sizeof( zend_long ) );
     SQLSRV_STATIC_ASSERT( sizeof( sqlsrv_phptype ) == sizeof( zend_long ));
 
+    setlocale(LC_ALL, "");
+
```
yitam: Please add a comment here justifying this line 
Also, we only need this in non-Windows. Please use ifdef

source/shared/localizationimpl.cpp
```diff
@@ -270,7 +281,43 @@ using namespace std;
 
 SystemLocale::SystemLocale( const char * localeName )
     :   m_pLocale( new std::locale(localeName) )
+    , m_uAnsiCP(CP_UTF8)
 {
```
yitam: So this means m_uAnsiCP is defaulted to CP_UTF8 if none is matched?

source/shared/localizationimpl.cpp
```diff
@@ -270,7 +281,43 @@ using namespace std;
...
+        { "UTF-32LE", 12000 }
+    };
+    if (localeName)
+    {
+        const char *charsetName = strchr(localeName, '.');
+        charsetName = charsetName ? charsetName + 1 : localeName;
+        for (const LocaleCP& lcp : lcpTable)
+        {
+        //    if (!strcasecmp(lcp.localeName, charsetName))
+           if (!strncasecmp(lcp.localeName, charsetName, strlen(lcp.localeName)))
```
yitam: please remove the commented line if you're sure it's not necessary

source/shared/localizationimpl.cpp
```diff
@@ -285,7 +332,7 @@ const SystemLocale & SystemLocale::Singleton()
 #if !defined(__GNUC__) || defined(NO_THREADSAFE_STATICS)
     #error "Relying on GCC's threadsafe initialization of local statics."
 #endif
-    static const SystemLocale s_Default( "en_US.utf-8" );
+    static const SystemLocale s_Default(setlocale(LC_ALL, NULL));
     return s_Default;
```
yitam: again, please add a comment here

test/functional/sqlsrv/MsCommon.inc
```diff
@@ -32,6 +33,18 @@ function setUTF8Data($val)
...
+    return $testingAnsiData;
+}
+
+function setTestAnsiData($val)
+{
+    global $testingAnsiData;
+    $testingAnsiData = $val;
+}
+
 function testMode()
```
yitam: Is it possible to refactor the above four methods? Use constants (like enums) or boolean variables to distinguish the two test cases (either utf-8 or not)?

test/functional/sqlsrv/MsData_UTF8.inc
```diff
@@ -9,7 +9,7 @@
 */
 
 
-function getTestData($index)
+function getTestData_UTF8($index)
 {
     $inputs = null;
```
yitam: why do we need to change the method name?

test/functional/sqlsrv/TC44_FetchArray.phpt
```diff
@@ -6,7 +6,16 @@ by checking all fetch type modes.
...
-<?php require('skipif_versions_old.inc'); ?>
+<?
+// locale must be set before 1st connection
+if ( !isWindows() ) {
+    setlocale(LC_ALL, "en_US.ISO-8859-1");
+}
+
+// this skips for older ODBC versions in Linux which doesn't support non-UTF8
+setTestAnsiData(true)
+php require('skipif_versions_old.inc');
```
yitam: why do you call setTestAnsiData() in SKIPIF section?

test/functional/sqlsrv/TC43_FetchData.phpt
```diff
@@ -91,16 +105,27 @@ function checkData($col, $actual, $expected)
     return ($success);
 }
 
-if (! isWindows()) {
-    setUTF8Data(true);
+if ( !isWindows() ) {
+    setlocale(LC_ALL, "en_US.ISO-8859-1");
+}
```
v-kaywon: Why do you need to set it again? I thought it's already set before skipif

test/functional/sqlsrv/TC44_FetchArray.phpt
v-kaywon: it's needed by the skipif

test/functional/sqlsrv/TC43_FetchData.phpt
```diff
@@ -91,16 +105,27 @@ function checkData($col, $actual, $expected)
...
 }
+
+// test utf8
 try {
+    setUTF8Data(true);
     fetchFields();
 } catch (Exception $e) {
     echo $e->getMessage();
 }
-setUTF8Data(false);
```
v-kaywon: By default, utf8data is off right? If you don't set it back to false at the end of the test, when the next test runs, the global uft8data will be true.

test/functional/sqlsrv/TC46_FetchNextResult.phpt
```diff
@@ -5,7 +5,16 @@ Verifies the functionality of "sqlsrv_next_result"
...
 --SKIPIF--
-<?php require('skipif_versions_old.inc'); ?>
+<?
+// locale must be set before 1st connection
+if ( !isWindows() ) {
+    setlocale(LC_ALL, "en_US.ISO-8859-1");
+}
+
+// this skips for older ODBC versions in Linux which doesn't support non-UTF8
+setTestAnsiData(true)
```
v-kaywon: perhaps put this part into another require? looks like the same thing is being done for every test

lilgreenbird:
Review status: 0 of 14 files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[source/shared/localizationimpl.cpp, line 285 at r1](https://reviewable.io:443/reviews/microsoft/msphpsql/598#-Kz64rbS9gz100vPLacM-r1-285:-Kz6VKP2aWu-QU18rYSp:b-5fe5zx) ([raw file](https://github.com/microsoft/msphpsql/blob/5bbb3924cfcf7a094d8fa1b5939b3f525d7d6304/source/shared/localizationimpl.cpp#L285)):*

So this means m_uAnsiCP is defaulted to CP_UTF8 if none is matched?
</blockquote></details>

yes that's the default

---

*[test/functional/sqlsrv/MsData_UTF8.inc, line 14 at r1](https://reviewable.io:443/reviews/microsoft/msphpsql/598#-Kz64rbS9gz100vPLacO-r1-14:-Kz6W8G7W1aRRvITQURd:bjli6e2) ([raw file](https://github.com/microsoft/msphpsql/blob/5bbb3924cfcf7a094d8fa1b5939b3f525d7d6304/test/functional/sqlsrv/MsData_UTF8.inc#L14)):*

why do we need to change the method name?
</blockquote></details>

because I got an error from including that said it was already defined or something

---

*[test/functional/sqlsrv/TC44_FetchArray.phpt, line 17 at r1](https://reviewable.io:443/reviews/microsoft/msphpsql/598#-Kz64rbTZKLeNFfuwUL--r1-17:-Kz6WbvmBkyRK-eAkKXy:bfurict) ([raw file](https://github.com/microsoft/msphpsql/blob/5bbb3924cfcf7a094d8fa1b5939b3f525d7d6304/test/functional/sqlsrv/TC44_FetchArray.phpt#L17)):*

why do you call setTestAnsiData() in SKIPIF section?
</blockquote></details>

because skipif_versions_old checks for it to see if we should skip

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/msphpsql/598)*

test/functional/sqlsrv/MsCommon.inc
lilgreenbird: changes have been reverted

test/functional/sqlsrv/MsCommon.inc
```diff
@@ -467,4 +454,13 @@ function handleErrors()
     }
 }
 
+// non-UTF8 locale support in ODBC 17 and above only
+function isLocaleSupported()
+{
```
yitam: Please use a more generic name to check for ODBC 17... other tests that require this ODBC ver but have nothing to do with locale can use this method :)

test/functional/sqlsrv/TC43_FetchData.phpt
```diff
@@ -91,16 +101,29 @@ function checkData($col, $actual, $expected)
...
 
 ?>
 --EXPECT--
 Test "Fetch - Field Data" completed successfully.
+Test "Fetch - Field Data" completed successfully.
```
yitam: put endtest() at the end of this test so you won't see two of them in the exepcted output
for the case in which you need to skip ansi test you won't invoke this endtest() anyway

test/functional/sqlsrv/TC43_FetchData.phpt
lilgreenbird: I added this here as the tests needs it. Some of them have skipifs that connect some don't it's hard to tell jsut by the name. I wanted to be consistent.

test/functional/sqlsrv/MsCommon.inc
lilgreenbird: this is generic, checking for ODBC 17 is meaningless, why? it's because locale support is there.
the only other place we check for ODBC version is isQualified it's the same issue there we don't call the version out there either. (btw I think it should be isAEQualified but that's a diff topic...

PR #612, opened 2017-11-30
refactor PDO_SQLSRV bind param related tests

v-kaywon:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/612)

### DIFF
--- a/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
+++ b/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
@@ -153,7 +153,7 @@ class ColumnMeta
     public $dataType;   //a string that includes the size of the type if necessary (e.g., decimal(10,5))
     public $colName;    //column name
     public $options;    //a string that is null by default (e.g. NOT NULL Identity (1,1) )
-    public $encType;    //randomized, deterministic, or normal; default is null
+    public $encType;    //randomized, deterministic, or none; default is null
     public $forceEncrypt; //force encryption on a datatype not supported by Column Encrypton
 
     public function __construct($dataType, $colName = null, $options = null, $encType = null, $forceEncrypt = false)
@@ -169,7 +169,7 @@ class ColumnMeta
             if (isColEncrypted()) {
                 $this->encType = "deterministic";
             } else {
-                $this->encType = "normal";
+                $this->encType = "none";
             }
         } else {
             $this->encType = $encType;
--- a/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
+++ b/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
@@ -13,6 +13,10 @@
 // to be runnable from the MSSQL teams' internal proprietary test running system
 //
 
+const KSP_NAME = 'MyCustomKSPName';
+const ENCRYPT_KEY = 'LPKCWVD07N3RG98J0MBLG4H2';
+const KSP_TEST_TABLE = 'CustomKSPTestTable';
+
 function isAEQualified($conn)
 {
     $msodbcsql_ver = $conn->getAttribute(PDO::ATTR_CLIENT_VERSION)["DriverVer"];
@@ -81,8 +85,9 @@ function getDSN($sqlsrvserver, $database, $keywords = '', $disableCE = false)
         $dsn .= "ColumnEncryption=Enabled;";
     }
     if ($keystore == "ksp" && !$disableCE) {
-        require('AE_Ksp.inc');
         $ksp_path = getKSPPath();
+        $ksp_name = KSP_NAME;
+        $encrypt_key = ENCRYPT_KEY;
         $dsn .= "CEKeystoreProvider=$ksp_path;CEKeystoreName=$ksp_name;CEKeystoreEncryptKey=$encrypt_key;";
     }
     if ($keywords) {
@@ -117,6 +122,27 @@ function getCekName()
     }
     return $cekName;
 }
+
+/**
+ * @return the path to the KSP dll/so file
+ */
+function getKSPpath() 
+{
+    $name = 'myKSP';
+
+    $dir_name = realpath(dirname(__FILE__));
+    $ksp = $dir_name . DIRECTORY_SEPARATOR . $name;
+    if ( strtoupper( substr( php_uname( 's' ), 0, 3 ) ) == 'WIN' ) {
+        $arch = 'x64';
+        if ( PHP_INT_SIZE == 4 ) // running 32 bit
+            $arch = '';
+        $ksp .= $arch . '.dll';
+    }
+    else
+        $ksp .= '.so';
+        
+    return $ksp;
+}  
 
 
 /**
@@ -521,6 +547,13 @@ function isColEncrypted()
     } else {
         return true;
     }
+}
+
+
+function isAEConnected()
+{
+    require('MsSetup.inc');
+    return $keystore != "none";
 }
 
 function teardown()
--- a/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
+++ b/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
@@ -185,7 +185,7 @@ class ColumnMeta
         $append = " ";
 
         // an identity column is not encrypted because a select query with identity column as the where clause is often run and the user want to have to bind parameter every time
-        if (isColEncrypted() && ($this->encType == "deterministic" || $this->encType == "ramdomized") && $this->isEncryptableType() 
+        if (isColEncrypted() && ($this->encType == "deterministic" || $this->encType == "randomized") && $this->isEncryptableType() 
             && stripos($this->options, "identity") === false && stripos($this->options, "rowguidcol") === false) {
             $cekName = getCekName();
             if (stripos($this->dataType, "char") !== false) {
--- a/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
+++ b/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
@@ -151,7 +151,7 @@ class ColumnMeta
         $append = " ";
 
         // an identity column is not encrypted because a select query with identity column as the where clause is often run and the user want to have to bind parameter every time
-        if (isColEncrypted() && $this->isEncryptableType() && stripos($this->options, "identity") === false) {
+        if (isColEncrypted() && $this->isEncryptableType() && stripos($this->options, "identity") === false && stripos($this->options, "rowguidcol") === false) {
             $cekName = getCekName();
             if (stripos($this->dataType, "char") !== false) {
                 $append .= "COLLATE Latin1_General_BIN2 ";
@@ -168,7 +168,7 @@ class ColumnMeta
      */
     public function isEncryptableType()
     {
-        $unsupportedTypes = array("money", "smallmoney", "image", "ntext", "text", "xml", "sql_variant");
+        $unsupportedTypes = array("money", "smallmoney", "image", "ntext", "text", "xml", "sql_variant", "timestamp", "geography", "geometry", "hierarchyid");
         if (!$this->forceEncrypt && in_array(strtolower($this->dataType), $unsupportedTypes)) {
             return false;
         } else {
--- a/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
+++ b/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
@@ -153,10 +153,10 @@ class ColumnMeta
     public $dataType;   //a string that includes the size of the type if necessary (e.g., decimal(10,5))
     public $colName;    //column name
     public $options;    //a string that is null by default (e.g. NOT NULL Identity (1,1) )
-    public $encType;    //randomized or deterministic; default is deterministic
-    public $forceEncrypt; //force encryption on a datatype no supported by Column Encrypton
+    public $encType;    //randomized, deterministic, or normal; default is null
+    public $forceEncrypt; //force encryption on a datatype not supported by Column Encrypton
 
-    public function __construct($dataType, $colName = null, $options = null, $encType = "deterministic", $forceEncrypt = false)
+    public function __construct($dataType, $colName = null, $options = null, $encType = null, $forceEncrypt = false)
     {
         if (is_null($colName)) {
             $this->colName = getDefaultColName($dataType);
@@ -165,6 +165,15 @@ class ColumnMeta
         }
         $this->dataType = $dataType;
         $this->options = $options;
+        if (is_null($encType)) {
+            if (isColEncrypted()) {
+                $this->encType = "deterministic";
+            } else {
+                $this->encType = "normal";
+            }
+        } else {
+            $this->encType = $encType;
+        }
         $this->encType = $encType;
         $this->forceEncrypt = $forceEncrypt;
     }
@@ -173,11 +182,11 @@ class ColumnMeta
      */
     public function getColDef()
     {
-        //return getColDef($this->colName, $this->dataType, $this->options, $this->encType);
         $append = " ";
 
         // an identity column is not encrypted because a select query with identity column as the where clause is often run and the user want to have to bind parameter every time
-        if (isColEncrypted() && $this->isEncryptableType() && stripos($this->options, "identity") === false && stripos($this->options, "rowguidcol") === false) {
+        if (isColEncrypted() && ($this->encType == "deterministic" || $this->encType == "ramdomized") && $this->isEncryptableType() 
+            && stripos($this->options, "identity") === false && stripos($this->options, "rowguidcol") === false) {
             $cekName = getCekName();
             if (stripos($this->dataType, "char") !== false) {
                 $append .= "COLLATE Latin1_General_BIN2 ";
--- a/test/functional/pdo_sqlsrv/PDO100_InsertNulls.phpt
+++ b/test/functional/pdo_sqlsrv/PDO100_InsertNulls.phpt
@@ -14,10 +14,10 @@ function insertNullsTest($bindType)
 {
     $outvar = null;
     $failed = false;
-    $conn =connect();
+    $conn = connect();
 
     $tableName = "pdo_test_table";
-    $dataType = array("c1_int" => "int",
+    $dataTypes = array("c1_int" => "int",
                       "c2_tinyint" => "tinyint",
                       "c3_smallint" => "smallint",
                       "c4_bigint" => "bigint",
@@ -45,12 +45,12 @@ function insertNullsTest($bindType)
                       "c26_smalldatetime" => "smalldatetime",
                       "c27_timestamp" => "timestamp",
                       "c28_xml" => "xml");
-    createTable($conn, $tableName, $dataType);
+    createTable($conn, $tableName, $dataTypes);
 
     $stmt = $conn->query("SELECT [TABLE_NAME],[COLUMN_NAME],[IS_NULLABLE] FROM [INFORMATION_SCHEMA].[COLUMNS] WHERE [TABLE_NAME] = '$tableName'");
 
     if ($stmt === false) {
-        FatalError("Could not query for column information on table $tableName");
+        fatalError("Could not query for column information on table $tableName");
     }
 
     while ($row = $stmt->fetch(PDO::FETCH_ASSOC)) {
@@ -98,7 +98,7 @@ try {
     echo $e->getMessage();
 }
 if ($failed) {
-    FatalError("Possible Regression: Could not insert NULL");
+    fatalError("Possible Regression: Could not insert NULL");
 } else {
     echo "Test 'PDO - Insert Nulls' completed successfully.\n";
 }
--- a/test/functional/pdo_sqlsrv/PDO100_InsertNulls.phpt
+++ b/test/functional/pdo_sqlsrv/PDO100_InsertNulls.phpt
@@ -5,74 +5,82 @@ Test inserting nulls into nullable columns
 --ENV--
 PHPT_EXEC=true
 --SKIPIF--
-<?php require('skipif.inc'); ?>
+<?php require_once('skipif_mid-refactor.inc'); ?>
 --FILE--
 <?php
-include 'MsCommon.inc';
+require_once("MsCommon_mid-refactor.inc");
 
-function InsertNullsTest($bindType)
+function insertNullsTest($bindType)
 {
-    include 'MsSetup.inc';
-  
     $outvar = null;
-
     $failed = false;
+    $conn =connect();
 
-    Setup();
+    $tableName = "pdo_test_table";
+    $dataType = array("c1_int" => "int",
+                      "c2_tinyint" => "tinyint",
+                      "c3_smallint" => "smallint",
+                      "c4_bigint" => "bigint",
+                      "c5_bit" => "bit",
+                      "c6_float" => "float",
+                      "c7_real" => "real",
+                      "c8_decimal" => "decimal(28,4)",
+                      "c9_numeric" => "numeric(32,4)",
+                      "c10_money" => "money",
+                      "c11_smallmoney" => "smallmoney",
+                      "c12_char" => "char(512)",
+                      "c13_varchar" => "varchar(512)",
+                      "c14_varchar_max" => "varchar(max)",
+                      "c15_nchar" => "nchar(512)",
+                      "c16_nvarchar" => "nvarchar(512)",
+                      "c17_nvarchar_max" => "nvarchar(max)",
+                      "c18_text" => "text",
+                      "c19_ntext" => "ntext",
+                      "c20_binary" => "binary(512)",
+                      "c21_varbinary" => "varbinary(512)",
+                      "c22_varbinary_max" => "varbinary(max)",
+                      "c23_image" => "image",
+                      "c24_uniqueidentifier" => "uniqueidentifier",
+                      "c25_datetime" => "datetime",
+                      "c26_smalldatetime" => "smalldatetime",
+                      "c27_timestamp" => "timestamp",
+                      "c28_xml" => "xml");
+    createTable($conn, $tableName, $dataType);
 
-    $conn = Connect();
+    $stmt = $conn->query("SELECT [TABLE_NAME],[COLUMN_NAME],[IS_NULLABLE] FROM [INFORMATION_SCHEMA].[COLUMNS] WHERE [TABLE_NAME] = '$tableName'");
 
-    DropTable($conn, $tableName);
-
-    CreateTable($conn, $tableName);
-
-    $stmt = $conn->query(<<<SQL
-SELECT [TABLE_NAME],[COLUMN_NAME],[IS_NULLABLE] FROM [INFORMATION_SCHEMA].[COLUMNS] WHERE [TABLE_NAME] = '$tableName'
-SQL
-);
-
-    if ($stmt === false)
-    {
+    if ($stmt === false) {
         FatalError("Could not query for column information on table $tableName");
     }
 
-    while ($row = $stmt->fetch(PDO::FETCH_ASSOC))
-    {
-        Trace($row['COLUMN_NAME'] . ": " . $row['IS_NULLABLE'] . "\n");
-
+    while ($row = $stmt->fetch(PDO::FETCH_ASSOC)) {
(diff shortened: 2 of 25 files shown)


test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
```diff
@@ -168,7 +168,7 @@ class ColumnMeta
      */
     public function isEncryptableType()
     {
-        $unsupportedTypes = array("money", "smallmoney", "image", "ntext", "text", "xml", "sql_variant");
+        $unsupportedTypes = array("money", "smallmoney", "image", "ntext", "text", "xml", "sql_variant", "timestamp", "geography", "geometry", "hierarchyid");
         if (!$this->forceEncrypt && in_array(strtolower($this->dataType), $unsupportedTypes)) {
```
yitam: Didn't realize you use the flag _forceEncrypt_. Under what conditions will this flag be true? How could we force encrypt a column type when AE is not supported for that type?

test/functional/pdo_sqlsrv/PDO100_InsertNulls.phpt
```diff
@@ -5,74 +5,82 @@ Test inserting nulls into nullable columns
...
-SQL
-);
-
-    if ($stmt === false)
-    {
+    $conn =connect();
+
+    $tableName = "pdo_test_table";
+    $dataType = array("c1_int" => "int",
+                      "c2_tinyint" => "tinyint",
```
yitam: Please put a space character before `connect()`
Since this `$dataType` is an array, should probably use plurals or any name that implies it's not singular

test/functional/pdo_sqlsrv/PDO100_InsertNulls.phpt
```diff
@@ -5,74 +5,82 @@ Test inserting nulls into nullable columns
...
+                      "c26_smalldatetime" => "smalldatetime",
+                      "c27_timestamp" => "timestamp",
+                      "c28_xml" => "xml");
+    createTable($conn, $tableName, $dataType);
+
+    $stmt = $conn->query("SELECT [TABLE_NAME],[COLUMN_NAME],[IS_NULLABLE] FROM [INFORMATION_SCHEMA].[COLUMNS] WHERE [TABLE_NAME] = '$tableName'");
+
+    if ($stmt === false) {
         FatalError("Could not query for column information on table $tableName");
     }
```
yitam: _fatalError_

test/functional/pdo_sqlsrv/PDO100_InsertNulls.phpt
```diff
@@ -82,32 +90,19 @@ SQL
...
-        echo $e->getMessage();
-    }
-
-    if ($failed)
-        FatalError("Possible Regression: Could not insert NULL");
-    
-    EndTest($testName);
+try {
+    $failed = false;
+    $failed |= insertNullsTest(PDO::PARAM_LOB);
```
yitam: why changed _failed_ from null to false?

test/functional/pdo_sqlsrv/PDO102_MaxOutputParams.phpt
```diff
@@ -5,50 +5,34 @@ Fetch data as VARCHAR(MAX)
...
-    {
+    if ($outstr !== $expected) {
         print_r($stmt->errorInfo());
+        dropProc($conn, $procName);
         return(-1);
     }
-
+    dropProc($conn, $procName);
     return(0);
 }
```
yitam: Suggestion, set a flag to the right value, dropProc, then return that flag
Also, isn't returning true or false more intuitive than -1 or 0?

test/functional/pdo_sqlsrv/PDO38_FetchBound.phpt
```diff
@@ -5,33 +5,27 @@ Verification for "PDOStatenent::fetch(PDO::FETCH_BOUND)".
...
+                  array('30', 'Ghi', 'tsr'),
+                  array('40', 'Jkl', 'qpo'),
+                  array('50', 'Mno', 'nml'),
+                  array('60', 'Pqr', 'kji'));
 
     // Insert using question mark placeholders
     $stmt1 = PrepareQuery($conn1, "INSERT INTO [$tableName] VALUES(?, ?, ?)");
-    foreach ($data as $row)
-    {
+    foreach ($data as $row) {
```
yitam: _prepareQuery_

test/functional/pdo_sqlsrv/PDO38_FetchBound.phpt
```diff
@@ -40,72 +34,78 @@ function Fetch()
     $stmt2 = PrepareQuery($conn1, "SELECT COUNT(id) FROM [$tableName]");
     $stmt2->execute();
     $num = $stmt2->fetchColumn();
-    echo 'There are ' . $num . " rows in the table.\n";
+    echo "There are $num rows in the table.\n";
     $stmt2->closeCursor();
```
yitam: why closeCursor() here?

test/functional/pdo_sqlsrv/PDO38_FetchBound.phpt
```diff
@@ -40,72 +34,78 @@ function Fetch()
...
-    $stmt1 = PrepareQuery($conn1, "SELECT idx, txt FROM [$tableName] ORDER BY idx");
+    if (!isColEncrypted()) {
+        $stmt1 = PrepareQuery($conn1, "SELECT idx, txt FROM [$tableName] ORDER BY idx");
+    } else {
+        // ORDER BY does not work on encrypted columns
+        $stmt1 = PrepareQuery($conn1, "SELECT idx, txt FROM [$tableName]");
+    }
     $stmt1->execute();
     $data = $stmt1->fetchAll(PDO::FETCH_COLUMN|PDO::FETCH_UNIQUE);
+    if (isColEncrypted()) {
```
yitam: Do not use $data here because it's confusing when it's used in the beginning for data insertion.

test/functional/pdo_sqlsrv/PDO38_FetchBound.phpt
```diff
@@ -40,72 +34,78 @@ function Fetch()
...
         var_dump($stmt2->fetch(PDO::FETCH_BOUND));
         $stmt2->closeCursor();
 
         var_dump($stmt3->execute());
-        if ($idx == 0)
-        {   // bindColumn()s after execute() has been called at least once
+        if ($idx == 0) {   // bindColumn()s after execute() has been called at least once
             $stmt3->bindColumn('idx', $col2);
         }
         var_dump($stmt3->fetch(PDO::FETCH_BOUND));
```
yitam: please use more descriptive variables than $col1 or $col2 (because one is for txt column and the other for idx, which is actually the first column in the table)

test/functional/pdo_sqlsrv/PDO61_BindColumn1.phpt
```diff
@@ -5,124 +5,124 @@ Verification for "PDOStatement::bindColumn".
...
+    // ORDER BY doesn't work for encrypted columns
+    // need to fetch all rows first then sort and print
+    if (!isColEncrypted()) {
+        $stmt1 = $conn1->prepare("SELECT idx, txt FROM [$tableName] ORDER BY idx");
+    } else {
+        $stmt1 = $conn1->prepare("SELECT idx, txt FROM [$tableName]");
+    }
+    $stmt1->execute();
+    $data = $stmt1->fetchAll(PDO::FETCH_COLUMN|PDO::FETCH_UNIQUE);
+    if (isColEncrypted()) {
```
yitam: the fetch style PDO::FETCH_COLUMN returns only one column? With AE, which column does it return here?

test/functional/pdo_sqlsrv/pdoStatement_bindParam_inout_emulate_prepare.phpt
```diff
@@ -1,30 +1,36 @@
...
-}
-catch(PDOException $e) {
-    print("Error: " . $e->getMessage() . "\n");
+    dropProc($dbh, $procName);
+    unset($stmt);
+    unset($dbh);
+} catch (PDOException $e) {
+    $error = $e->getMessage();
+    $pass = !isColEncrypted() && $error === "SQLSTATE[IMSSP]: Statement with emulate prepare on does not support output or input_output parameters.";
+    $pass |= isColEncrypted() && ($error === "SQLSTATE[IMSSP]: Connection with Column Encryption enabled do no support PDO::ATTR_EMULATE_PREPARES with binding parameters.");
```
yitam: I didn't realize that the error message is not right... shouldn't it be "does not support"?

yitam: Checking error messages calls `isColEncrypted()` twice. Is it necessary to do it this way? 
Should this be isAEConnect() instead?

test/functional/pdo_sqlsrv/pdoStatement_bindParam_output_emulate_prepare.phpt
```diff
@@ -1,52 +1,48 @@
...
+    unset($stmt2);
+    unset($conn);
+} catch (PDOException $e) {
     print("Error: " . $e->getMessage() . "\n");
 }
 ?>
 --EXPECT--
-Error: SQLSTATE[IMSSP]: Statement with emulate prepare on does not support output or input_output parameters.
\ No newline at end of file
+Error: SQLSTATE[IMSSP]: Statement with emulate prepare on does not support output or input_output parameters.
```
yitam: So no AE specific error message?

test/functional/pdo_sqlsrv/pdo_fetch_variants_diff_styles.phpt
```diff
@@ -145,99 +158,88 @@ function GetValueFromObject($obj, $col)
...
-            {
+            for ($j = 0; $j < $numCols; $j++) {
                 $stmtOriginal->execute();
                 $value = $stmtOriginal->fetchColumn($j);
-                
+
                 // insert this value into the only column in the new table
-                $stmtTmp->bindParam($j + 1, $value, PDO::PARAM_STR);
+                $stmtTmp->bindParam(1, $value, PDO::PARAM_STR);
                 $res = $stmtTmp->execute();
```
yitam: thanks for fixing this bug in the test

test/functional/pdo_sqlsrv/pdo_katmai_special_types.phpt
```diff
@@ -1,150 +1,173 @@
...
+    $expOutput = array("c1_time" => '03:32:25.5643401',
+                  "c2_date" => '1439-01-10',
+                  "c3_datetimeoffset" => '0221-01-12 06:39:07.0620256 +00:00',
+                  "c4_geography" => 'e6100000010c8fc2f5285c2f53c0295c8fc2f5e83b40',
+                  "c5_geometry" => '0000000001140000000000803e401f85eb51b81ee5bf48e17a14ae073f4052b81e85eb51d8bf',
+                  "c6_hierarchyid" => '5bc0',
+                  "c7_uniqueidentifier" => '35413141383846372d333734392d343641332d384137412d454641453733454645383846');
+    if (isColEncrypted()) {
+        $toRemove = array("c4_geography", "c5_geometry", "c6_hierarchyid");
+        foreach ($toRemove as $key) {
```
yitam: why removed these columns? can we leave them not encrypted?

test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
v-kaywon: there are a few tests for testing error returned from creating table containing some of these types

test/functional/pdo_sqlsrv/PDO100_InsertNulls.phpt
v-kaywon: I just think it makes more sense to or a boolean

test/functional/pdo_sqlsrv/PDO38_FetchBound.phpt
v-kaywon: close the cursor so $stmt2 can be re-executed in line 47 (although not closing the cursor has no problem either, but it's a nice way to ensure the current result set is closed)

test/functional/pdo_sqlsrv/PDO61_BindColumn1.phpt
v-kaywon: PDO::FETCH_COLUMN | PDO::FETCH_UNIQUE returns idx as key and txt as value

test/functional/pdo_sqlsrv/pdoStatement_bindParam_inout_emulate_prepare.phpt
v-kaywon: isAEConnected is very quick (just checking the $keystore isn't null), so not going to change the code to only calling isAEConnected once.

test/functional/pdo_sqlsrv/pdoStatement_bindParam_output_emulate_prepare.phpt
v-kaywon: yes. With output parameter, the emulate prepare parser detects the output parameter (i.e., having '=' right after '?') and throws this error much earlier than detecting inout parameter

test/functional/pdo_sqlsrv/pdo_katmai_special_types.phpt
v-kaywon: I removed them because insertion of these special datatypes doesn't work with bind param. (i.e., cannot do `$stmt->bindparam(1, 'POINT(27.91 -76.74)')`) These datatypes need to be inserted via direct queries.

yitam: then please add a comment why you have to remove them

test/functional/pdo_sqlsrv/pdoStatement_bindParam_inout_emulate_prepare.phpt
yitam: so would you change to isAEConnected()?

test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
```diff
@@ -127,10 +153,10 @@ class ColumnMeta
     public $dataType;   //a string that includes the size of the type if necessary (e.g., decimal(10,5))
     public $colName;    //column name
     public $options;    //a string that is null by default (e.g. NOT NULL Identity (1,1) )
-    public $encType;    //randomized or deterministic; default is deterministic
-    public $forceEncrypt; //force encryption on a datatype no supported by Column Encrypton
+    public $encType;    //randomized, deterministic, or normal; default is null
+    public $forceEncrypt; //force encryption on a datatype not supported by Column Encrypton
```
yitam: From what I see below looks like it's either `deterministic` or `normal`?

test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
```diff
@@ -147,11 +182,11 @@ class ColumnMeta
...
     public function getColDef()
     {
-        //return getColDef($this->colName, $this->dataType, $this->options, $this->encType);
         $append = " ";
 
         // an identity column is not encrypted because a select query with identity column as the where clause is often run and the user want to have to bind parameter every time
-        if (isColEncrypted() && $this->isEncryptableType() && stripos($this->options, "identity") === false && stripos($this->options, "rowguidcol") === false) {
+        if (isColEncrypted() && ($this->encType == "deterministic" || $this->encType == "ramdomized") && $this->isEncryptableType() 
+            && stripos($this->options, "identity") === false && stripos($this->options, "rowguidcol") === false) {
             $cekName = getCekName();
```
yitam: why don't you add the check for identity and rowguidcol to isEncryptableType()? (might consider changing the name to isEncryptable()...)

test/functional/pdo_sqlsrv/PDO38_FetchBound.phpt
```diff
@@ -99,50 +91,50 @@ try {
...
     $stmt1->execute();
-    $col1s = array();
+    $cols = array();
     while ($stmt1->fetch(PDO::FETCH_BOUND)) {
-        array_push($col1s, $col1);
+        array_push($cols, $col);
     }
     if (isColEncrypted()) {
-        sort($col1s);
+        sort($cols);
```
yitam: why sorting required?

test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
```diff
@@ -139,6 +165,15 @@ class ColumnMeta
...
         $this->options = $options;
+        if (is_null($encType)) {
+            if (isColEncrypted()) {
+                $this->encType = "deterministic";
+            } else {
+                $this->encType = "normal";
+            }
+        } else {
+            $this->encType = $encType;
+        }
```
yitam: I'm not sure I understand the logic. $encType is null by default
Looks like it's either `deterministic` or `normal`?

test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
```diff
@@ -147,11 +182,11 @@ class ColumnMeta
      */
     public function getColDef()
     {
-        //return getColDef($this->colName, $this->dataType, $this->options, $this->encType);
         $append = " ";
 
         // an identity column is not encrypted because a select query with identity column as the where clause is often run and the user want to have to bind parameter every time
-        if (isColEncrypted() && $this->isEncryptableType() && stripos($this->options, "identity") === false && stripos($this->options, "rowguidcol") === false) {
+        if (isColEncrypted() && ($this->encType == "deterministic" || $this->encType == "ramdomized") && $this->isEncryptableType()
```
yitam: correct the typo `ramdomized`

test/functional/pdo_sqlsrv/pdo_katmai_special_types.phpt
```diff
@@ -1,150 +1,176 @@
...
+    if (isColEncrypted()) {
+        // remove these types from tests because these types require direct query for the data to be inserted
+        // and the insertRow common function uses bind parameters to insertion when column encryption is enabled
+        $toRemove = array("c4_geography", "c5_geometry", "c6_hierarchyid");
+        foreach ($toRemove as $key) {
+            unset($dataTypes[$key]);
+            unset($data[$key]);
+            unset($expOutput[$key]);
+        }
+    }
```
yitam: Just a suggestion, push these three columns to $data array only if _NOT_ AE enabled

test/functional/pdo_sqlsrv/pdo_katmai_special_types.phpt
```diff
@@ -1,150 +1,176 @@
...
         $value2 = $row[$fld2];
 
-        if ($input !== trim($value1))
-        {
+        if ($input !== trim($value1)) {
             echo "The value is unexpected!\n";
         }
-        if ($value1 !== $value2)
-        {
+        if (trim($value1) !== trim($value2)) {
```
yitam: why trimming $value2 (not in the original test)
please add a comment to explain

test/functional/pdo_sqlsrv/pdo_param_output_variants.phpt
```diff
@@ -1,75 +1,82 @@
...
         $stmt->execute();
-        echo "Does REVERSE work? $string \n";  
-    }
-    catch (Exception $e)
-    {
+        // Connection with Column Encryption enabled works for non encrypted SQL_VARIANT
+        // Since SQLDescribeParam is called
+        if (isColEncrypted() && $string === "987654321") {
+            echo "Test input output parameter with SQL_VARIANT successfully.\n";
+
```
yitam: successfully _done_ or _completed_

test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
v-kaywon: `$encType` is null by default so when the user does not provide an argument for this ("normal", "deterministic" or "randomized"), the `encType` is set based on whether or not isColEncrypted is true (i.e., when isColEncrypted is true, "deterministic" is the default encType; when isColEncrypted isn't true, "normal" is the encType). If the user provides an argument however, the `encType` is whatever that the user provides.

test/functional/pdo_sqlsrv/pdo_simple_update_variants.phpt
```diff
@@ -2,150 +2,134 @@
...
-    $stmt = $conn->query($query);  
+    $query = "SELECT * FROM $tableName ORDER BY id";
+    $stmt = $conn->query($query);
 
     $stmt->setFetchMode(PDO::FETCH_CLASS, 'Food');
-    while ($food = $stmt->fetch())
-    {   
-        echo "\nID: " . $food->id . " "; 
+    $foodArray = array();
+    while ($food = $stmt->fetch()) {
```
yitam: what's the use of $foodArray?

test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
```diff
@@ -1,167 +1,193 @@
 --TEST--
-PDO Fetch Mode Test with emulate prepare 
+PDO Fetch Mode Test with emulate prepare
 --DESCRIPTION--
-Basic verification for "PDOStatement::setFetchMode()�.
+Basic verification for "PDOStatement::setFetchMode()�.
 --ENV--
```
yitam: Please modify the funny character in the DESCRIPTION

test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
```diff
@@ -1,167 +1,193 @@
...
+    $tableName = "pdo_test_table";
+    createTable($conn1, $tableName, array(new ColumnMeta("int", "ID", "NOT NULL PRIMARY KEY"), "Policy" => "varchar(2)", "Label" => "varchar(10)", "Budget" => "money"));
 
-	// Prepare test table
-	CreateTableEx($conn1, $tableName, "ID int NOT NULL PRIMARY KEY, Policy VARCHAR(2), Label VARCHAR(10), Budget MONEY", null);
+    try {
+        $res = $conn1->setAttribute(PDO::ATTR_EMULATE_PREPARES, true);
+        if ($res) {
+            echo "setAttribute should have failed.\n\n";
+        }
```
yitam: No checking if connected with AE here? Because it should have failed anyway?

test/functional/pdo_sqlsrv/PDO38_FetchBound.phpt
v-kaywon: I think I just didn't see this one when I was refactoring. I'll fix it.

test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
```diff
@@ -1,167 +1,193 @@
...
+        array_push($resultset, $stmt1->fetch(PDO::FETCH_BOTH, PDO::FETCH_ORI_PRIOR));
+        // sort and print
+        sort($resultset);
+        $assocArr['ID'] = $resultset[0]['ID'];
+        $assocArr['Policy'] = $resultset[0]['Policy'];
+        $assocArr['Label'] = $resultset[0]['Label'];
+        $assocArr['Budget'] = $resultset[0]['Budget'];
+        var_dump($assocArr);
+        //print "$resultset[1][1]\n";
+        print($resultset[1][1] . "\n");
```
yitam: please remove the commented line

test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
```diff
@@ -1,167 +1,193 @@
...
+    if (isColEncrypted()) {
+        $stmtOptions[PDO::SQLSRV_ATTR_CURSOR_SCROLL_TYPE] = PDO::SQLSRV_CURSOR_BUFFERED;
+    }
+    $stmt1 = $conn1->prepare($tsql, $stmtOptions);
+    $stmt1->execute();
+    // The row order in the resultset when the column is encrypted (which is dependent on the encrytion key used)
+    // is different from the order when the column is not enabled
+    // To make this test work, if the column is encrypted, fetch all then find the corresponding row
+    if (!isColEncrypted()) {
+        var_dump($stmt1->fetch(PDO::FETCH_ASSOC));
```
yitam: again, can probably choose not to encrypt the ID column so that you can enforce ORDER BY?
would it work even with emulate prepare?

test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
```diff
@@ -1,167 +1,193 @@
...
-	$stmt4 = null;
-	$conn1 = null;
+    echo "\nBudget Metadata....\n";
+    $metadata = $stmt4->getColumnMeta(2);
+    var_dump($metadata);
 
-	EndTest($testName);
+    // Cleanup
+    DropTable($conn1, $tableName);
+    unset($stmt1);
```
yitam: dropTable

test/functional/pdo_sqlsrv/test_encoding_UTF8_emulate_prepare.phpt
```diff
@@ -1,113 +1,86 @@
...
+
+    $option;
+    if (!isAEConnected()) {
+        $option[PDO::ATTR_EMULATE_PREPARES] = true;
+    } else {
+        $option[PDO::ATTR_EMULATE_PREPARES] = false;
+    }
+    $stmt5 = $conn->prepare("SELECT ? = c2_char FROM $tbname", $option);
+    $stmt5->setAttribute(PDO::SQLSRV_ATTR_ENCODING, PDO::SQLSRV_ENCODING_UTF8);
+    $outValue1 = "hello";
```
yitam: setting encoding to UTF-8 is redundant... It's the default, isn't it?

test/functional/pdo_sqlsrv/pdo_katmai_special_types.phpt
v-kaywon: If I do it that way I need to push these variables separately

test/functional/pdo_sqlsrv/pdo_param_output_variants.phpt
v-kaywon: More code seems to be involved if I do it that way. I'll leave it the way it is now.

test/functional/pdo_sqlsrv/pdo_simple_update_variants.phpt
v-kaywon: it was there before to do the sorting. It's not needed anymore, thx.

test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
v-kaywon: yes

test/functional/pdo_sqlsrv/test_encoding_UTF8_emulate_prepare.phpt
v-kaywon: the connection attribute was set to ENCODING_SYSTEM, and this ones sets the statement attribute to UTF8

test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
yitam: exactly, so if `encType` is null, it's no encryption, correct? 
therefore, you can leave `encType` = null

test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
v-kaywon: I'll skip this since they original test did not have ORDER BY.

test/functional/pdo_sqlsrv/pdo_param_output_variants.phpt
yitam: Fine. Please just fix the grammar in the output message :)

test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
v-kaywon: if encType is null that means user is lazy and relies on MsCommon it pick the encType

PR #613, opened 2017-12-01
fix double reclaring KSP function; fixed tests with emulate prepare

v-kaywon:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/613)

### DIFF
--- a/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
+++ b/test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
@@ -81,7 +81,7 @@ function getDSN($sqlsrvserver, $database, $keywords = '', $disableCE = false)
         $dsn .= "ColumnEncryption=Enabled;";
     }
     if ($keystore == "ksp" && !$disableCE) {
-        require('AE_Ksp.inc');
+        require_once('AE_Ksp.inc');
         $ksp_path = getKSPPath();
         $dsn .= "CEKeystoreProvider=$ksp_path;CEKeystoreName=$ksp_name;CEKeystoreEncryptKey=$encrypt_key;";
     }
@@ -521,6 +521,13 @@ function isColEncrypted()
     } else {
         return true;
     }
+}
+
+
+function isAEConnected()
+{
+    require('MsSetup.inc');
+    return $keystore != "none";
 }
 
 function teardown()
--- a/test/functional/pdo_sqlsrv/pdo_138_unicode_column_name.phpt
+++ b/test/functional/pdo_sqlsrv/pdo_138_unicode_column_name.phpt
@@ -11,13 +11,13 @@ Github 138. Test for Unicode Column Metadata.
  * @param mixed $query
  * @return PDOStatement
  */
-function prepare($connection, $query) {
+function prepare($connection, $query)
+{
     $pdo_options = array();
     // emulate and binding parameter with direct query are not support in Always Encrypted
-    if ( !isColEncrypted() )
-    {
-        $pdo_options[PDO::ATTR_EMULATE_PREPARES] = TRUE;
-        $pdo_options[PDO::SQLSRV_ATTR_DIRECT_QUERY] = TRUE;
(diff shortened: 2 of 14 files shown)


test/functional/pdo_sqlsrv/pdo_138_unicode_column_name.phpt
```diff
@@ -11,13 +11,13 @@ Github 138. Test for Unicode Column Metadata.
...
-function prepare($connection, $query) {
+function prepare($connection, $query)
+{
     $pdo_options = array();
     // emulate and binding parameter with direct query are not support in Always Encrypted
-    if ( !isColEncrypted() )
-    {
-        $pdo_options[PDO::ATTR_EMULATE_PREPARES] = TRUE;
-        $pdo_options[PDO::SQLSRV_ATTR_DIRECT_QUERY] = TRUE;
+    if (!isAEConnected()) {
```
yitam: supported

test/functional/pdo_sqlsrv/pdo_138_unicode_column_name.phpt
```diff
@@ -35,7 +35,7 @@ try {
 
     // Create Table
     $tbname = "mytáble";
-    createTable( $connection, $tbname, array( new ColumnMeta( "nchar(10)", "id" ), new ColumnMeta( "nchar(10)", "väriable" ), new ColumnMeta( "nchar(10)", "tésting" )));
+    createTable($connection, $tbname, array( new ColumnMeta("nchar(10)", "id"), new ColumnMeta("nchar(10)", "väriable"), new ColumnMeta("nchar(10)", "tésting")));
```
yitam: please remove the space after `array(` too

test/functional/pdo_sqlsrv/pdo_140_emulate_prepare_mix_binary.phpt
```diff
@@ -4,22 +4,23 @@ Test emulate prepare with mix bound param encodings including binary data
...
-      if (isset($columnInformation['blobs'][$field_name])) {
-        $this->bindParam($placeholder, $field_value, PDO::PARAM_LOB, 0, PDO::SQLSRV_ENCODING_BINARY);
-      }
-      else {
-        // Even though not a blob, make sure we retain a copy of these values.
-        $this->bindParam($placeholder, $field_value, PDO::PARAM_STR);
-      }
+class MyStatement extends PDOStatement
+{
+    public function BindValues(array &$values, $placeholder_prefix, $columnInformation)
```
yitam: Looks like you only fix the formatting? If so, please rename BindValues as bindValues()

test/functional/pdo_sqlsrv/MsCommon_mid-refactor.inc
```diff
@@ -81,7 +81,7 @@ function getDSN($sqlsrvserver, $database, $keywords = '', $disableCE = false)
         $dsn .= "ColumnEncryption=Enabled;";
```
yitam: For sqlsrv tests, this part is like `if (isColEncrypted() && !$disableCE)`... should we change in sqlsrv too?

v-kaywon: isColEncrypted is true only if `$keystore != 'none'` and `$dataEncrypted == false`, but I think we want to connect with `ColumnEncryption` enabled as long as `keystore` is not none. `dataEncrypted` only applies to table creation and row insertion. So yes, please change this in the SQLSRV side. But be aware that some tests will begin to fail because a lot of AE features and limitations are base on ColumnEncrypted connection.

test/functional/pdo_sqlsrv/pdo_140_emulate_prepare_mix_binary.phpt
v-kaywon: I changed from isColEncrypted to IsAEConnect on line 38 :)
Yes, I'll change to bindValues()

PR #616, opened 2017-12-04
define getKSPPath function and ksp related variabled in MsCommon_mid-refactor.inc

v-kaywon:
---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/616)


v-kaywon:
No there are three ksp related tests that still use that

yitam:
those tests in sqlsrv simply connect using the common helper method, and skip if $keystore != ksp

Issue #615, opened 2017-12-01
Fetching varchar(max) as a stream results in Fatal Error with Always Encrypted 


yitam:
Tested in Windows with the latest sqlsrv 5.1.2-preview
1. Connect with ColumnEncryption enabled
2. Create a table with one column of type varchar(max) with encryption
3. Insert one row and retrieve data from this table as stream
4. Fatal Error (seemingly stuck in an infinite loop).
5. Same problem with UTF-8 option (connection or retrieval)

The repro script:

```
sqlsrv_query($conn, "CREATE TABLE [test_max_fields] ( [varchar_max_col] varchar(max) COLLATE Latin1_General_BIN2 ENCRYPTED WITH (ENCRYPTION_TYPE = deterministic, ALGORITHM = 'AEAD_AES_256_CBC_HMAC_SHA_256', COLUMN_ENCRYPTION_KEY = AEColumnKey) )");

$inValue = str_repeat("ÃÜðßZZýA©", 1000);
$insertSql = "INSERT INTO [test_max_fields] (varchar_max_col) VALUES (?)";
$params = array($inValue);

$stmt = sqlsrv_prepare($conn, $insertSql, $params);
if ($stmt) {
    sqlsrv_execute($stmt);
} 

$query = "SELECT * FROM [test_max_fields]";
$stmt = sqlsrv_prepare($conn, $query);
if ($stmt) {
    sqlsrv_execute($stmt);
} 

 if (!sqlsrv_fetch($stmt)) { 
     fatalError("Failed to fetch row ");
 }

$stream = sqlsrv_get_field($stmt, 0, SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_CHAR));

if ($stream === false) {
    fatalError("Failed to read field");
} else {
    $value = '';
    if ($stream) {
        $num = 0;
        while (!feof($stream)) {
            $value .= fread($stream, 8192); // infinite loop?! 
        }
        fclose($stream);
        
        if (!checkData($value, $inValue)) {  // compare the data to see if they match!
            fatalError("Data corruption!\n");
        }
    } else {
        fatalError("Error in getting stream!");
    }
}

```

yitam:
Resolved in [5.2.0-RC](https://github.com/Microsoft/msphpsql/releases/tag/v5.2.0-RC)

PR #621, opened 2017-12-08
add error handling for fetching stream with always encrypted

v-kaywon:
#615

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/621)


source/shared/core_stmt.cpp
```diff
@@ -1711,6 +1711,9 @@ void core_get_field_common( _Inout_ sqlsrv_stmt* stmt, _In_ SQLUSMALLINT field_i
 		// for how these fields are used.
 		case SQLSRV_PHPTYPE_STREAM:
 		{
+            CHECK_CUSTOM_ERROR(stmt->conn->ce_option.enabled, stmt, SQLSRV_ERROR_ENCRYPTED_STREAM_FETCH) {
+                throw core::CoreException();
+            }
```
yitam: please fix the spacing

source/pdo_sqlsrv/pdo_util.cpp
```diff
@@ -421,6 +421,10 @@ pdo_error PDO_ERRORS[] = {
         SQLSRV_ERROR_OUTPUT_PARAM_TYPES_NOT_SUPPORTED,
         { IMSSP, (SQLCHAR*) "Stored Procedures do not support text, ntext or image as OUTPUT parameters.", -83, false }
     },
+    {
+        SQLSRV_ERROR_ENCRYPTED_STREAM_FETCH,
+        { IMSSP, (SQLCHAR*) "Connection with Column Encryption enabled does not support fetching stream. Please fetch the data as a string.", -84, false }
+    },
     { UINT_MAX, {} }
```
yitam: Do we have to add this error in PDO when it doesn't support returning streams?

test/functional/sqlsrv/sqlsrv_fetch_large_stream.phpt
```diff
@@ -0,0 +1,86 @@
...
+<?php require('skipif_versions_old.inc'); ?>
+--FILE--
+<?php
+require_once('MsCommon.inc');
+
+$conn = AE\connect();
+
+$tableName = "test_max_fields";
+dropTable($conn, $tableName);
+AE\createTable($conn, $tableName, array(new AE\ColumnMeta("varchar(max)", "varchar_max_col")));
```
yitam: there's no need to drop table here because createTable will do it :)

test/functional/sqlsrv/sqlsrv_fetch_large_stream.phpt
```diff
@@ -0,0 +1,86 @@
...
+        fclose($stream);
+        if (checkData($value, $inValue)) {  // compare the data to see if they match!
+            $success = true;
+        } else {
+            fatalError("Data corruption!\n");
+        }
+    } else {
+        fatalError("Error in getting stream!");
+    }
+}
```
yitam: perhaps simplify this checking because $stream is expected to fail with AE?

lilgreenbird: use PHP_STREAM_BUFFER_SIZE?

source/pdo_sqlsrv/pdo_util.cpp
v-kaywon: Since the error is thrown in the core code, it needs to be here to compile.

test/functional/sqlsrv/sqlsrv_fetch_large_stream.phpt
v-kaywon: @yitam But the test should work without AE too right?
@lilgreenbird  where did you get that constant from? somehow when I put it in the test it returns undefined constant

yitam: I just thought you could check if AE is enabled first and expected it to fail with the correct message?
Otherwise, it should pass with the results matched?

PR #624, opened 2017-12-11
Check if the array is empty instead of using sizeof (issue with PHP 7.2)

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/624)


test/functional/sqlsrv/MsHelper.inc
```diff
@@ -364,7 +364,7 @@ function isQualified($conn)
 function connect($options = array(), $disableCE = false)
 {
     require('MsSetup.inc');
-    if (sizeof($options) > 0) {
+    if (! empty($options)) {
```
v-kaywon: I think the coding style is no space after !

PR #626, opened 2017-12-12
Modify sqlsrv tests to handle the new error message with stream data

yitam:
#621

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/626)

### DIFF
--- a/test/functional/sqlsrv/MsHelper.inc
+++ b/test/functional/sqlsrv/MsHelper.inc
@@ -90,7 +90,8 @@ class ColumnMeta
     {
         $append = " ";
 
-        if (isColEncrypted() && $this->encryptable) {
+        if ($this->encryptable && isDataEncrypted()) {
+
             $cekName = getCekName();
             if (stripos($this->dataType, "char") !== false) {
                 $append .= "COLLATE Latin1_General_BIN2 ";
@@ -325,13 +326,26 @@ function getSeqPlaceholders($num)
 }
 
 /**
- * @return bool false if $keystore specified in MsSetup.inc is none or data not encrypted,
- *              otherwise return true
+ * @return bool false if $keystore specified in MsSetup.inc is none; return true otherwise
  */
 function isColEncrypted()
 {
     global $keystore, $dataEncrypted;
-    if ($keystore == KEYSTORE_NONE || !$dataEncrypted) {
+    if ($keystore === KEYSTORE_NONE) {
+        return false;
+    } else {
+        return true;
+    }
+}
+
+/**
+ * @return bool false if $keystore specified in MsSetup.inc is none or data not encrypted;
+ *              return true otherwise 
+ */
+function isDataEncrypted()
+{
+    global $keystore, $dataEncrypted;
+    if ($keystore === KEYSTORE_NONE || !$dataEncrypted) {
         return false;
     } else {
         return true;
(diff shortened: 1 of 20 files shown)


test/functional/sqlsrv/MsHelper.inc
```diff
@@ -325,13 +326,26 @@ function getSeqPlaceholders($num)
...
 /**
- * @return bool false if $keystore specified in MsSetup.inc is none or data not encrypted,
- *              otherwise return true
+ * @return bool false if $keystore specified in MsSetup.inc is none; return true otherwise
  */
 function isColEncrypted()
 {
     global $keystore, $dataEncrypted;
-    if ($keystore == KEYSTORE_NONE || !$dataEncrypted) {
+    if ($keystore === KEYSTORE_NONE) {
```
v-kaywon: If you change this to just check if the connection is column encryption enabled, what happens to all the tests that uses this function?

yitam: Most tests only care to distinguish whether it's connected with ColumnEncryption enabled or not. Whether data is encrypted or not only affects whether the table columns are encrypted or not and some AE specific tests

test/functional/sqlsrv/sqlsrv_prepare.phpt
```diff
@@ -121,14 +129,14 @@ binding parameters, including output parameters, using the simplified syntax.
...
     sqlsrv_close($conn);
 ?>
---EXPECTF--
+--EXPECTREGEX--
 1
-12.0
+12\.0
 testtestte
-This is some text meant to test binding parameters to streams
+(This is some text meant to test binding parameters to streams)?
```
v-kaywon: should compare the results in the test instead (what if column encryption is enabled and it successfully fetches the stream, the test would still pass)

test/functional/sqlsrv/sqlsrv_query.phpt
```diff
@@ -61,5 +69,8 @@ sqlsrv_query test. Performs same tasks as 0006.phpt, using sqlsrv_query.
     sqlsrv_close($conn);
 
 ?>
---EXPECTF--
-1 <br/>12.0 <br/>testtestte <br/>This is some text meant to test binding parameters to streams<br/>%A
+--EXPECTREGEX--
+1
+12\.0
+testtestte
+(This is some text meant to test binding parameters to streams|Connection with Column Encryption enabled does not support fetching stream. Please fetch the data as a string.)
```
v-kaywon: Again, I prefer comparing the messages in the test

test/functional/sqlsrv/sqlsrv_send_stream_data.phpt
```diff
@@ -153,11 +161,11 @@ binding streams using full syntax.
...
-sqlsrv_query(3) failed.
+--EXPECTREGEX--
+sqlsrv_query\(2\) failed\.
+sqlsrv_query\(3\) failed\.
 1
-12.0
+12\.0
 testtestte
-This is some text meant to test binding parameters to streams
+(This is some text meant to test binding parameters to streams|Connection with Column Encryption enabled does not support fetching stream. Please fetch the data as a string.)
```
v-kaywon: please compare the message in the test

test/functional/sqlsrv/MsHelper.inc
v-kaywon: OK, as long as you've double check the usage of this function makes sense for each test

PR #628, opened 2017-12-13
fix warning according to prefast code analysis

v-kaywon:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/628)

### DIFF
--- a/source/shared/core_conn.cpp
+++ b/source/shared/core_conn.cpp
@@ -305,7 +305,7 @@ bool core_compare_error_state( _In_ sqlsrv_conn* conn,  _In_ SQLRETURN rc, _In_ 
     if( SQL_SUCCEEDED( rc ) ) 
         return false;
 
-    SQLCHAR state[ SQL_SQLSTATE_BUFSIZE ];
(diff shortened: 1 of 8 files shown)


source/shared/core_conn.cpp
```diff
@@ -321,13 +321,12 @@ bool core_compare_error_state( _In_ sqlsrv_conn* conn,  _In_ SQLRETURN rc, _In_
...
 {
+#ifndef _WIN32
     char szBuf[DEFAULT_CONN_STR_LEN+1];     // use a large enough buffer size
     WORD cbBufMax = DEFAULT_CONN_STR_LEN; 
     WORD cbBufOut;
     char *pszBuf = szBuf;
-    bool found_driver = false;
 
-#ifndef _WIN32
     // get all the names of the installed drivers delimited by null characters
```
yitam: why removed `#ifndef`?

v-kaywon: it's not removed. I moved it up to line 324. Those variables are not needed in Windows.

PR #629, opened 2017-12-14
Fix PDO67_BindTruncation test

yitam:
Doing `"SET IDENTITY_INSERT [$tableName] ON` or `OFF` with prepared statements has no effect. Only works with direct queries.
Reason:  A prepared statement runs in a separate context.
For more info, please click [here](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/b6d41dfc-ad83-4ed2-88be-6573baa040de/dynamically-set-identityinsert-onoff-in-stored-procedure?forum=transactsql)

This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/629)

### DIFF
--- a/test/functional/pdo_sqlsrv/PDO67_BindTruncation.phpt
+++ b/test/functional/pdo_sqlsrv/PDO67_BindTruncation.phpt
@@ -15,19 +15,16 @@ try {
 
     // Prepare test table
     $tableName = "pdo_test_table";
-    createTable($conn1, $tableName, array(new ColumnMeta("int", "id", "IDENTITY NOT NULL"), "class" => "int", "value" => "char(32)"));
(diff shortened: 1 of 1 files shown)


test/functional/pdo_sqlsrv/PDO67_BindTruncation.phpt
```diff
@@ -53,7 +50,8 @@ try {
     // Check data
     $id = 0;
     $value = '';
-    $stmt2 = $conn1->query($tsql4);
+    $tsql = "SELECT id, value FROM [$tableName]";
+    $stmt2 = $conn1->query($tsql);
```
v-kaywon: I think we should test batch query when AE is off

PR #636, opened 2017-12-19
fix tests related to having parameter markers in the select list

v-kaywon:
Added limitation "parameters in the SELECT list" to Wiki.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/636)

### DIFF
--- a/test/functional/pdo_sqlsrv/pdo_param_output_select_variant.phpt
+++ b/test/functional/pdo_sqlsrv/pdo_param_output_select_variant.phpt
@@ -81,6 +81,7 @@ try {
     dropTable($conn, $tableName);
     unset($conn);
 } catch (Exception $e) {
+    // binding parameters in the select list is not supported with Column Encryption
     $expMsg = "Implicit conversion from data type sql_variant to nvarchar(max) is not allowed. Use the CONVERT function to run this query.";
     $aeExpMsg = "Invalid Descriptor Index";
     checkError($e, $expMsg, $aeExpMsg);
--- a/test/functional/pdo_sqlsrv/pdo_param_output_select_variant.phpt
+++ b/test/functional/pdo_sqlsrv/pdo_param_output_select_variant.phpt
@@ -57,6 +57,16 @@ function createVariantTable($conn, $tableName)
     }
 }
 
+function checkError($e, $expMsg, $aeExpMsg)
+{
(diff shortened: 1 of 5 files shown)


test/functional/pdo_sqlsrv/pdo_param_output_select_variant.phpt
```diff
@@ -65,12 +71,15 @@ try {
     dropTable($conn, $tableName);
     unset($conn);
 } catch (Exception $e) {
-    echo $e->getMessage();
+    $error = $e->getMessage();
+    if (!(!isAEConnected() && strpos($error, "Implicit conversion from data type sql_variant to nvarchar(max) is not allowed. Use the CONVERT function to run this query.") !== false) &&
+        !(isAEConnected() && strpos($error, "Invalid Descriptor Index") !== false)) {
+        echo $error;
```
yitam: Is it possible to simplify the conditions and error checking here? Perhaps use a helper method with the expected message? The same applies to the other test in this PR that uses the very similar logic.

test/functional/pdo_sqlsrv/pdo_prepare_emulatePrepare_unicode.phpt
```diff
@@ -70,9 +70,21 @@ try {
     print_r("Prepare with emulate prepare and and SQLSRV_ENCODING_SYSTEM:\n");
     $stmt = prepareStmt($conn, $query, $options, PDO::PARAM_STR, 0, PDO::SQLSRV_ENCODING_SYSTEM);
     $row = $stmt->fetch(PDO::FETCH_ASSOC);
-    print_r($row);
-    if ($stmt->rowCount() == 0) {
-        print_r("No results for this query\n");
+    // Linux supports binding UTF8 data with SQLSRV_ENCODING_SYSTEM since the default encoding in Linux is UTF8
+    $success = true;
+    if (!(strtoupper( substr( php_uname( 's' ),0,3 ) ) === 'WIN') && isAEConnected()) {
+        if ($row['name'] != "가각" || $row['status'] != 1 || $row['age'] != 30) {
```
yitam: Not just Linux but also Mac? Perhaps also explains why AE is related to this condition?
in the else block please annotate that in Windows the default is not utf-8

PR #638, opened 2017-12-19
Testing locale issues

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/638)

### DIFF
--- a/test/functional/sqlsrv/MsCommon.inc
+++ b/test/functional/sqlsrv/MsCommon.inc
@@ -449,7 +449,7 @@ function isLocaleSupported()
     if (isWindows()) {
         return true;
     }
-    if (AE\isDataEncrypted()) {
+    if (AE\isColEncrypted()) {
         return false;
     }
     // now check ODBC version
--- a/test/functional/sqlsrv/MsCommon.inc
+++ b/test/functional/sqlsrv/MsCommon.inc
@@ -98,6 +98,8 @@ function startTest($testName)
  */
 function endTest($testName)
 {
+    // reset the utf-8 flag
+    setUTF8Data(false);
     echo "Test \"$testName\" completed successfully.\n";
 }
 
--- a/test/functional/sqlsrv/MsCommon.inc
+++ b/test/functional/sqlsrv/MsCommon.inc
@@ -98,8 +98,6 @@ function startTest($testName)
  */
 function endTest($testName)
 {
-    // reset the utf-8 flag
-    setUTF8Data(false);
     echo "Test \"$testName\" completed successfully.\n";
 }
 
--- a/test/functional/sqlsrv/MsCommon.inc
+++ b/test/functional/sqlsrv/MsCommon.inc
@@ -84,18 +84,6 @@ function isDaasMode()
     return ($daasMode ? true : false);
 }
 
-// function isAEQualified($conn)
-// {
-    // $msodbcsql_ver = sqlsrv_client_info($conn)['DriverVer'];
-    // $server_ver = sqlsrv_server_info($conn)['SQLServerVersion'];
-    // $msodbcsql_maj = explode(".", $msodbcsql_ver)[0];
-    // $msodbcsql_min = explode(".", $msodbcsql_ver)[1];
-    // if ($msodbcsql_maj < 17 || explode('.', $server_ver)[0] < 13) {
-        // return false;
-    // }
-    // return true;
-// }
-
 function startTest($testName)
 {
     if (traceMode()) {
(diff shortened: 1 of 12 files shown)


test/functional/sqlsrv/TC34_PrepAndExec.phpt
```diff
@@ -6,18 +6,25 @@ Validates that a prepared statement can be successfully executed more than once.
 --ENV--
 PHPT_EXEC=true
 --SKIPIF--
-<?php require('skipif_versions_old.inc'); ?>
+<?php 
+// locale must be set before 1st connection
+if (!isWindows()) {
+    setlocale(LC_ALL, "en_US.ISO-8859-1");
+}
+require('skipif_versions_old.inc');
```
lilgreenbird: We don't want to add this skipif because we do want to run this the UTF8 test for older versions. That was the whole point of isLocaleSupported, it does the checking there so we can do the skip inside each test. That was actually your idea, remember :)

test/functional/sqlsrv/MsCommon.inc
```diff
@@ -84,18 +84,6 @@ function isDaasMode()
     return ($daasMode ? true : false);
 }
 
-// function isAEQualified($conn)
```
lilgreenbird: don't we have tests which need this??

test/functional/sqlsrv/MsCommon.inc
```diff
@@ -110,6 +98,8 @@ function startTest($testName)
  */
 function endTest($testName)
 {
+    // reset the utf-8 flag
+    setUTF8Data(false);
```
lilgreenbird: I don't think we should do this here as this is a generic function it doesn't necessary have anything to do with UTF8 testing. The tests which care about this should set this themselves

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -330,7 +330,7 @@ function getSeqPlaceholders($num)
  */
 function isColEncrypted()
 {
-    global $keystore, $dataEncrypted;
+    global $keystore;
```
lilgreenbird: since you are changing this here can you fix - I really hate stupid unnecessary code like this. This is just
return $keystore != KEYSTORE_NONE
!!

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -330,7 +330,7 @@ function getSeqPlaceholders($num)
...
     if ($keystore === KEYSTORE_NONE) {
```
lilgreenbird: since you are changing this here can you fix - I really hate stupid unnecessary code like this. This is just
return $keystore != KEYSTORE_NONE
!!

lilgreenbird: similar issue in isDataEncrypted()

test/functional/sqlsrv/TC34_PrepAndExec.phpt
yitam: No worries. This skipif will only skip the test if AE is enabled. Otherwise, the test will still run

test/functional/sqlsrv/MsCommon.inc
yitam: the method has been moved to MsHelper in the AE namespace

test/functional/sqlsrv/MsCommon.inc
yitam: ok

test/functional/sqlsrv/MsHelper.inc
yitam: ok

yitam:
Review status: 0 of 12 files reviewed at latest revision, 3 unresolved discussions.

---

*[test/functional/sqlsrv/MsCommon.inc, line 102 at r4](https://reviewable.io:443/reviews/microsoft/msphpsql/638#-L0kuxklsLweUVxAS_MA-r4-102:-L0vV4i_BKIT237zr4An:b-896fix) ([raw file](https://github.com/microsoft/msphpsql/blob/6b209526cd80453defa6c7363ae8929268600727/test/functional/sqlsrv/MsCommon.inc#L102)):*

ok
</blockquote></details>

Done.

---

*[test/functional/sqlsrv/MsHelper.inc, line 334 at r4](https://reviewable.io:443/reviews/microsoft/msphpsql/638#-L0kuxklsLweUVxAS_MB-r4-334:-L0vV5K9Dhr6qZ6M86o2:b-896fix) ([raw file](https://github.com/microsoft/msphpsql/blob/6b209526cd80453defa6c7363ae8929268600727/test/functional/sqlsrv/MsHelper.inc#L334)):*

ok
</blockquote></details>

Done.

---

*[test/functional/sqlsrv/TC34_PrepAndExec.phpt, line 14 at r4](https://reviewable.io:443/reviews/microsoft/msphpsql/638#-L0p8aSClVOJqBR6tJbS-r4-14:-L0vV6FmU-4a2v9jymgF:b-896fix) ([raw file](https://github.com/microsoft/msphpsql/blob/6b209526cd80453defa6c7363ae8929268600727/test/functional/sqlsrv/TC34_PrepAndExec.phpt#L14)):*

No worries. This skipif will only skip the test if AE is enabled. Otherwise, the test will still run
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/msphpsql/638)*

PR #660, opened 2018-01-12
Update Known Issues

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/660)


v-kaywon:
I think we should remove the whole Known Issues entirely from README, or just point to the latest release.

yitam:
I've already made this change in the master branch, and this is to make sure the dev branch matches its contents. Or do you think we should simply remove this section?

PR #665, opened 2018-01-18
fix tests that uses scrollable cursor with AE

v-kaywon:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/665)

### DIFF
--- a/test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
+++ b/test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
@@ -61,9 +61,7 @@ try {
     echo "Now selecting....\n";
     $tsql = "SELECT * FROM [$tableName]";
     $stmtOptions[PDO::ATTR_CURSOR] = PDO::CURSOR_SCROLL;
-    if (isColEncrypted()) {
-        $stmtOptions[PDO::SQLSRV_ATTR_CURSOR_SCROLL_TYPE] = PDO::SQLSRV_CURSOR_BUFFERED;
-    }
+
     $stmt1 = $conn1->prepare($tsql, $stmtOptions);
     $stmt1->execute();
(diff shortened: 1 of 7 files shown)


test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
```diff
@@ -61,9 +61,7 @@ try {
     echo "Now selecting....\n";
     $tsql = "SELECT * FROM [$tableName]";
     $stmtOptions[PDO::ATTR_CURSOR] = PDO::CURSOR_SCROLL;
-    if (isColEncrypted()) {
-        $stmtOptions[PDO::SQLSRV_ATTR_CURSOR_SCROLL_TYPE] = PDO::SQLSRV_CURSOR_BUFFERED;
-    }
+
     $stmt1 = $conn1->prepare($tsql, $stmtOptions);
     $stmt1->execute();
     // The row order in the resultset when the column is encrypted (which is dependent on the encrytion key used)
```
yitam: Do we still need specific if...else block when AE is enabled?

test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
```diff
@@ -114,7 +112,6 @@ try {
     } else {
         // more and less than operators do not work for encrypted columns
         $tsql = "SELECT * FROM [$tableName] WHERE NOT ID = :id";
```
yitam: This else block is necessary, I understand, because comparisons are not supported

test/functional/pdo_sqlsrv/pdostatement_fetchmode_emulate_prepare.phpt
v-kaywon: yes, the row order of the result set between ae and off seems to be different

PR #671, opened 2018-01-24
Modified Readme and scripts to accept path to source in command line

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/671)

### DIFF
--- a/buildscripts/builddrivers.py
+++ b/buildscripts/builddrivers.py
@@ -271,7 +271,6 @@ if __name__ == '__main__':
         
         answer = input("Download source from a GitHub repo? [y/n]: ")
         if answer == 'yes' or answer == 'y' or answer == '':
-            # download = True
             repo = input("Name of the repo (hit enter for 'Microsoft'): ")
             branch = input("Name of the branch (hit enter for 'dev'): ")
             if repo == '':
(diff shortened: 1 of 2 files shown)


buildscripts/builddrivers.py
```diff
@@ -91,6 +90,28 @@ def clean_or_remove(self, root_dir, work_dir):
...
+        """
+        while True:
+            if source_path is None:
+                source = input('Enter the full path to the source folder: ')
+            else:
+                source = input("Hit ENTER to use '" + source_path + "' or provide another path to the source folder: ")
+                if len(source) == 0:
+                    source = source_path
+
+            valid = True
```
lilgreenbird: what is this for? it's not used anywhere?

buildscripts/builddrivers.py
```diff
@@ -259,9 +270,8 @@ def validate_input(question, values):
         debug_mode = input("Debug enabled? [y/n]: ")
         
         answer = input("Download source from a GitHub repo? [y/n]: ")
-        download = False
         if answer == 'yes' or answer == 'y' or answer == '':
-            download = True
+            # download = True
```
lilgreenbird: remove

buildscripts/builddrivers.py
yitam: only used in Interactive mode but not used in Testing mode

PR #697, opened 2018-02-26
Test ReturnDatesAsStrings true with AE

david-puglielli:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/697)

### DIFF
--- a/test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
+++ b/test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
@@ -136,7 +136,10 @@ function CompareDateTimeObject($dateTimeType, &$expectedDateTime, &$retrievedDat
                 $retrievedDateTime['time'] != $expectedDateTime['time'][4]) fatalError("Time tested, times do not match!");
             if (explode(" ", $retrievedDateTime['datetime'])[1] != explode(" ", $expectedDateTime['datetime'][1])[1] and
                 explode(" ", $retrievedDateTime['datetime'])[1] != explode(" ", $expectedDateTime['datetime'][2])[1] and
-                explode(" ", $retrievedDateTime['datetime'])[1] != explode(" ", $expectedDateTime['datetime'][3])[1]) fatalError("Time tested, datetimes do not match!");
+                explode(" ", $retrievedDateTime['datetime'])[1] != explode(" ", $expectedDateTime['datetime'][3])[1]) {
+                    print_r($retrievedDateTime);
+                    print_r($expectedDateTime);
+                fatalError("Time tested, datetimes do not match!");}
             if (explode(" ", $retrievedDateTime['datetime2'])[1] != explode(" ", $expectedDateTime['datetime2'][1])[1] and
                 explode(" ", $retrievedDateTime['datetime2'])[1] != explode(" ", $expectedDateTime['datetime2'][2])[1] and
                 explode(" ", $retrievedDateTime['datetime2'])[1] != explode(" ", $expectedDateTime['datetime2'][3])[1] and
--- a/test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
+++ b/test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
@@ -31,22 +31,17 @@ function ExecutePreparedStmt($stmt)
     }
 }
 
-// Compare dates retrieved from the database with the date used for testing.
+// Compare dates retrieved from the database with the date used for testing
+// when ReturnDatesAsStrings is true.
 // $expectedDateTime is an array of strings like '2002-01-31 23:59:59.049876'
 // $retrievedDateTime is date/time string of format either 'Y-m-d H:i:s.u' or
 // 'Y-m-d H:i:s.u P', which is the format returned when $returnDatesAsStrings 
-// is true, or by the date_format() calls in FetchDatesAndOrTimes when a PHP 
-// DateTime object is retrieved (i.e. when $returnDatesAsStrings is false),
-// unless $datetimetype is date or time. In those cases:
-// If $datetimetype is date and $returnDatesAsStrings is false:
-//   The date is as expected, the time defaults to 00:00:00.0000
-// If $datetimetype is time and $returnStrings is false:
-//   The date defaults to the current date, the time is as expected. 
-// If $datetimetype is date and $returnStrings is true:
+// is true, unless $datetimetype is date or time. In those cases:
+// If $datetimetype is 'date':
 //   $retrievedDateTime is only a date.
-// If $datetimetype is 'time' and $returnStrings is true:
+// If $datetimetype is 'time':
 //   $retrievedDateTime is only a time.
-function CompareDateTime($datetimetype, $returnStrings, &$expectedDateTime, $retrievedDateTime)
+function CompareDateTimeString($datetimetype, &$expectedDateTime, $retrievedDateTime)
 {
     $expected_date_time = array();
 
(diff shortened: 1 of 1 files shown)


test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,595 @@
...
+        fatalError("Failure in sqlsrv_prepare");
+    } else {
+        $r = sqlsrv_execute($stmt);
+        if ($r == false) {
+            $errors = sqlsrv_errors();
+            if ($errors[0]['SQLSTATE'] != '22018' and
+                $errors[0]['SQLSTATE'] != '22008' and
+                $errors[0]['SQLSTATE'] != '22007' and
+                $errors[0]['SQLSTATE'] != '07006') {
+                print_r($errors);
```
yitam: Please write a comment to explain what these SQLSTATES mean

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,595 @@
...
+// the date defaults to the current date. If however $returnStrings is true
+// and $datetimetype is 'date', $retrieved_date is only a date,
+// and if $datetimetype is 'time' and $returnStrings is true, $retrieved_date
+// only a time.
+// The concatenations involving zeroes below are to make direct string
+// comparisons feasible. Also, because PHP maxes out at microsecond precision
+// and SQL Server maxes out at 0.1 microsecond precision, the more precise 
+// types require an extra 0 for some comparisons.
+function CompareDateTime($datetimetype, $returnStrings, &$testingDate, $retrievedDate)
+{
```
yitam: Suggestion: refactor this method into one for date, one for time or one for combination?

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,595 @@
...
+    if (sizeof($ret_date_time)>1) {
+        $datetimeonly = $ret_date_time[0]." ".$ret_date_time[1];
+    }
+    if (sizeof($ret_date_time)>2) {
+        $timezone = $ret_date_time[2];
+        $dtoffset = $ret_date_time[0]." ".substr($ret_date_time[1], 0, -1)." ".$ret_date_time[2];
+    }
+   
+    if ($returnStrings == true) {
+        switch ($datetimetype) {
```
yitam: Suggestion: one comparison for datetime values as strings and one method for datetime objects?

Also, the variable names are too similar and hard to understand the logic. Is it possible to distinguish them by using adjectives like "actual" and "expected"?

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,595 @@
...
+    $stmt = AE\createTable($conn, $tableName, $columns);
+    if (!$stmt) {
+        fatalError("Failed to create table $tableName\n");
+    }
+
+    $insertSql = "INSERT INTO [$tableName] (id, [c1_$datetimetype]) VALUES (?, ?)";
+    for ($i=0; $i<$array_size; $i++)
+    {
+        $stmt = sqlsrv_prepare($conn, $insertSql, array($i, array($formats_array[$i], SQLSRV_PARAM_IN, SQLSRV_PHPTYPE_STRING('utf-8'), $SQLSRV_SQLTYPE_CONST)));
+        ExecutePreparedStmt($stmt);
```
yitam: why utf-8?

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,595 @@
...
+date_default_timezone_set('Canada/Pacific');
+sqlsrv_configure('WarningsReturnAsErrors', 0);
+sqlsrv_configure('LogSeverity', SQLSRV_LOG_SEVERITY_ALL);
+sqlsrv_configure('LogSubsystems', SQLSRV_LOG_SYSTEM_OFF);
+
+require_once('MsCommon.inc');
+
+$returnDatesAsStrings = true;
+
+$conn = AE\connect(array('ReturnDatesAsStrings' => $returnDatesAsStrings));
```
yitam: Suggestion: refactor this last section into one method that takes the boolean flag $returnDateAsStrings

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,595 @@
...
+                              $day."-".$month."-".$year,
+                              $month_name." ".$day.", ".$year,
+                              $day."-".$month_name."-".$year,
+                              $day."-".$month_abbr."-".$year
+                              );
+$time_formats_nonODBC = array($hour12.":".$minute." ".$meridian,
+                              $hour12.":".$minute.":".$second." ".$meridian,
+                              $hour12.":".$minute.":".$second.".".$frac." ".$meridian,
+                              $hour12.":".$minute.":".$second.".".$frac.$frac2." ".$meridian,
+                              $hour.":".$minute
```
yitam: Please enlighten me what these nonODBC formats are for.

david-puglielli: Some of them do work when passed in to sqlsrv_prepare as strings.

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
david-puglielli: Everything is compared as strings, because when a DateTime object is returned it is first formatted as a string using date_format() before being passed to this function.

david-puglielli:
I've added this file (and addressed some comments) in my new PR. I will close this one shortly.

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,627 @@
...
+            case 'date':
+                // Direct comparison of retrieved date and expected date
+                if ($retrievedDateTime != $expected_date_time[0][0]) {
+                    fatalError("Dates do not match!");
+                }                    
+                break;
+            case 'time': 
+                // Compare SQL time with expected time. The expected time was input
+                // with an accuracy of microseconds and the SQL Servertime type has
+                // accuracy to 100 ns, so times are returned with an extra zero. For
```
v-kaywon: change the comments to "The expected time was input with an accuracy of 0.000001s and the SQL Server time type has an accuracy to 0.0000001s, ". Compare the input and output is a lot easier when both are the same unit.

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,627 @@
...
+                    $retrievedDateTime != $expected_date_time[1][1]."0" and
+                    $retrievedDateTime != $expected_date_time[2][1]."0" and
+                    $retrievedDateTime != $expected_date_time[3][1]."0" ) {
+                        fatalError("Times do not match!");
+                }
+                break;
+            case 'datetime':
+                // Compare retrieved SQL datetime with expected date/time.
+                // SQL Server's datetime type is accurate to milliseconds and
+                // the expected time is accurate to microseconds, so append
```
v-kaywon: again it would be easier to understand  if the units are the same (i.e., 0.001s and 0.000001s)

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,627 @@
...
+                // three zeroes to the retrieved time for comparison.
+                if ($retrievedDateTime."000" != $expectedDateTime[0] and
```
v-kaywon: not sure why this passed. `$expectedDateTime[0]` is `$year."-".$month."-".$day." ".$hour.":".$minute.":".$second.".".$frac.$frac2`, and `$frac2` is `'9876'`.

When the input is first inserted, the input is `'2002-01-31 23:59:29.049876'`. The actual data inserted however should be `'2002-01-31 23:59:29.049'` since datetime only goes up to a precision of 0.001s.

When the data is retrieved, the `$retrievedDateTime` is `'2002-01-31 23:59:29.049'`, and you append `"000"` to it which makes it `'2002-01-31 23:59:29.049000'`, however, that is obviously not the same as `'2002-01-31 23:59:29.049876'`. So this condition should've evaluated to false.

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,627 @@
...
+                        fatalError("Datetimeoffsets do not match!");
+                }
+                break;
+            case 'smalldatetime':
+                // Compare retrieved SQL smalldatetime with expected date/time.
+                // SQL Server's smalldatetime type is accurate to seconds and
+                // the expected time is accurate to microseconds, so append
+                // '.000000' to the expected time for comparison.
+                if ($retrievedDateTime.".000000" != $expectedDateTime[0] and
+                    $retrievedDateTime.".000000" != $expectedDateTime[1] and
```
v-kaywon: same as the datetime case, not sure how these two conditions pass

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,627 @@
...
+// unless $datetimetype is date or time. In those cases:
+// If $datetimetype is date and $returnDatesAsStrings is false:
+//   The date is as expected, the time defaults to 00:00:00.0000
+// If $datetimetype is time and $returnStrings is false:
+//   The date defaults to the current date, the time is as expected. 
+// If $datetimetype is date and $returnStrings is true:
+//   $retrievedDateTime is only a date.
+// If $datetimetype is 'time' and $returnStrings is true:
+//   $retrievedDateTime is only a time.
+function CompareDateTime($datetimetype, $returnStrings, &$expectedDateTime, $retrievedDateTime)
```
v-kaywon: I agree with Jenny. Would be nice if this function is split into 2: one fore CompareDateTimeString, and one for CompareDateTimeObject

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,627 @@
...
+                break;
+            case 'time': 
+                // Compare SQL time with expected time. The expected time was input
+                // with an accuracy of microseconds and the SQL Servertime type has
+                // accuracy to 100 ns, so times are returned with an extra zero. For
+                // comparison the zero is appended to the times in expected_time_date.
+                if ($retrievedDateTime != $expected_date_time[0][1]."0" and
+                    $retrievedDateTime != $expected_date_time[1][1]."0" and
+                    $retrievedDateTime != $expected_date_time[2][1]."0" and
+                    $retrievedDateTime != $expected_date_time[3][1]."0" ) {
```
v-kaywon: can remove this lost condition since `$expected_date_time[3][1]` corresponds to format `$year."-".$month."-".$day." ".$hour.":".$minute.":00.000000"` and that should never be received with time type.

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
david-puglielli: I'll split it into two, but is it clear that when a DateTime object is returned, the test converts it to a string using date_format() before passing to CompareDateTime?

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,627 @@
...
+                break;
+            case 'smalldatetime':
+                // Compare retrieved SQL smalldatetime with expected date/time.
+                // SQL Server's smalldatetime type is accurate to seconds and
+                // the expected time is accurate to microseconds, so append
+                // '.000000' to the expected time for comparison.
+                if ($retrievedDateTime.".000000" != $expectedDateTime[0] and
+                    $retrievedDateTime.".000000" != $expectedDateTime[1] and
+                    $retrievedDateTime.".000000" != $expectedDateTime[2] and
+                    $retrievedDateTime.".000000" != $expectedDateTime[3] ) {
```
v-kaywon: I thought smalldatetime is only accurate to minute (see line 326). Why is it being compared to the expectedDateTime with seconds precision (elements 0, 1 and 2)

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,627 @@
...
+InsertDatesAndOrTimes($conn, 'datetime2', $datetime2_formats_all, $SZ_DATETIME_all, SQLSRV_SQLTYPE_DATETIME2);
+InsertDatesAndOrTimes($conn, 'datetimeoffset', $datetimeoffset_formats_all, $SZ_DATETIME_all, SQLSRV_SQLTYPE_DATETIMEOFFSET);
+InsertDatesAndOrTimes($conn, 'smalldatetime', $datetimesmall_formats_all, $SZ_DATETIME_all, SQLSRV_SQLTYPE_SMALLDATETIME);
+
+FetchDatesAndOrTimes($conn, 'date', $expectedDateTime, $returnDatesAsStrings);
+FetchDatesAndOrTimes($conn, 'time', $expectedDateTime, $returnDatesAsStrings);
+FetchDatesAndOrTimes($conn, 'datetime', $expectedDateTime, $returnDatesAsStrings);
+FetchDatesAndOrTimes($conn, 'datetime2', $expectedDateTime, $returnDatesAsStrings);
+FetchDatesAndOrTimes($conn, 'datetimeoffset', $expectedDateTime, $returnDatesAsStrings);
+FetchDatesAndOrTimes($conn, 'smalldatetime', $expectedDateTime, $returnDatesAsStrings);
```
v-kaywon: line 434-46 and line 454-466 are the same. Maybe you can refactor them a bit?

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
```diff
@@ -0,0 +1,627 @@
...
+// when the DateTime objects retrieved from a table are formatted as strings
+// with date_format().
+$expectedDateTime = array($year."-".$month."-".$day." ".$hour.":".$minute.":".$second.".".$frac.$frac2,
+                          $year."-".$month."-".$day." ".$hour.":".$minute.":".$second.".".$frac."0000",
+                          $year."-".$month."-".$day." ".$hour.":".$minute.":".$second.".000000",
+                          $year."-".$month."-".$day." ".$hour.":".$minute.":00.000000",
+                          $year."-".$month."-".$day." ".$hour.":".$minute.":".$second.".".$frac.$frac2." ".$tz_correction,
+                          $year."-".$month."-".$day." ".$hour.":".$minute.":".$second.".".$frac."0000 ".$tz_correction,
+                          $year."-".$month."-".$day." ".$hour.":".$minute.":".$second.".000000 ".$tz_correction,
+                          $year."-".$month."-".$day." ".$hour.":".$minute.":00.000000 ".$tz_correction,
```
v-kaywon: is this case ever going to occur? time only has minute precision when the type is smalldatetime, however then it's smalldatetime, there should not be offset right?

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
david-puglielli: Seconds are optional with the time type, so it's possible that time fields do not have seconds.

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
david-puglielli: It is accurate to minute but may include the seconds (as ':00').

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
david-puglielli: Yes, here I can remove this format.

david-puglielli: Actually it is needed here too, because we are inserting and retrieving all date/time types. If I remove this, a smalldatetime retrieved using date_format() gives something like 2002-01-31 23:59:00.000000, so this case can occur.

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
v-kaywon: Yes that's clear. Both compare with strings, but one from DatetimeAsString, and one is from a datetime object using date_format()

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
v-kaywon: then is should not be compared with $expectedDateTime[0], $expectedDateTime[1], and $expectedDateTime[2] right? Since they has seconds of 29.

test/functional/sqlsrv/sqlsrv_ae_datetimes_as_strings.phpt
v-kaywon: Then you can remove this comparison in CompareDateTimeString (line 110)? The retrieve value in this function doesn't use date_format().

david-puglielli:
Latest version requires some refactoring.

PR #708, opened 2018-03-07
sqlsrv tests for testing output params with sqltypes

lilgreenbird:
test output params with sqltypes

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/708)


test/functional/sqlsrv/AEData.inc
```diff
@@ -1,113 +1,206 @@
...
+{
+    switch ( $sqlType ) {
+		case 'SQLSRV_SQLTYPE_BIGINT':
+		    return SQLSRV_SQLTYPE_BIGINT;
+			break;
+		case 'SQLSRV_SQLTYPE_BINARY':
+		    return SQLSRV_SQLTYPE_BINARY;
+			break;
+		case 'SQLSRV_SQLTYPE_BIT':
+		    return SQLSRV_SQLTYPE_BIT;
```
v-kaywon: for getting the constant from a string, you can do `constant('SQLSRV_SQLTYPE_BIT')`
for getting the constant of sqltypes with parenthesis, you can do `call_user_func('SQLSRV_SQLTYPE_CHAR', 5)`

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_datetime.phpt
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+            // 22018 is the SQLSTATE for any incompatible conversion errors
+            if ($isCompatible && sqlsrv_errors()[0]['SQLSTATE'] == 22018) {
+                echo "$sqlType should be compatible with $dataType\n";
+                $success = false;
+            }
+        } else {
+            // always encrypted only allow sqlType that is identical to the encrypted column datatype
+            if (stripos("SQLSRV_SQLTYPE_" . $dataType, $sqlType) !== false) {
```
v-kaywon: this condition should be removed. The purpose of this test is to test the conversions between the datatype in the column and the SQLTYPE in binding output parameter. If this condition is here, only the matching SQLTYPE is being tested (e.g., only SQLSRV_SQLTYPE_DATE is being tested against a date column). Most conversions should not work, but some conversions are actually supported with AE (e.g., from a tinyint to a smallint if the data fits into a tinyint). For this test specifically, there is no supported conversions for any datetime types. But the test should check if the error returned is the operand clash error.

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_datetime.phpt
```diff
@@ -0,0 +1,115 @@
...
+            foreach ($compatList[$dataType] as $compatType) {
+                if (stripos($compatType, $sqlType) !== false) {
+                    $isCompatible = true;
+                }
+            }
+            // 22018 is the SQLSTATE for any incompatible conversion errors
+            if ($isCompatible && sqlsrv_errors()[0]['SQLSTATE'] == 22018) {
+                echo "$sqlType should be compatible with $dataType\n";
+                $success = false;
+            }
```
v-kaywon: this `if(!AE\isColEncrypted())` block is a bit confusing. The condition `$isCompatible && sqlsrv_errors()[0]['SQLSTATE'] == 22018` is checking if the error exists in the create stored procedure query (since that is the last operation performed and sqlsrv_errors get the error from the last operation), which makes this block of code unnecessary since rarely there will be something wrong with creating a stored procedure. This block of code should be moved down to after the store procedure is called to see if there is any error from calling the store procedure.

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_numeric.phpt
```diff
@@ -0,0 +1,165 @@
...
+                sqlsrv_execute($stmt);
+				$errors = sqlsrv_errors();
+                if ( !empty($errors) ) {
+				    var_dump( sqlsrv_errors() );
+					die("Error executing statement.");
+				}
+
+                print("c_det: " . $c_detOut . "\n");
+                print("c_rand: " . $c_randOut . "\n");
+                $inputValues = array_slice(${explode("(", $dataType)[0] . "_params"}, 1, 2);
```
v-kaywon: `$inputValues` is already defined in line 38; don't have to do it again here

PR #720, opened 2018-03-14
sqlsrv tests for output parms with sqltypes

lilgreenbird:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/720)


test/functional/sqlsrv/AEData.inc
```diff
@@ -110,4 +110,67 @@ function get_default_size_prec( $sqlType )
...
+			// our tests always use precision 5 for SQLSRV_SQLTYPE_NCHAR
+		    return SQLSRV_SQLTYPE_NCHAR(5);
+			break;		
+		case 'SQLSRV_SQLTYPE_NUMERIC':
+			// our tests always use precision 10 scale 5 for SQLSRV_SQLTYPE_NUMERIC
+		    return SQLSRV_SQLTYPE_NUMERIC(10, 5);
+			break;
+		default:
+		     echo "get_sqlType_constant: Invalid SQL Type $sqlType\n";
+			 break;
```
v-kaywon: Indents problem

test/functional/sqlsrv/sqlsrv_ae_output_param_all.phpt
```diff
@@ -51,9 +49,7 @@ $spSql = "CREATE PROCEDURE $spname (
                 @c13_datetimeoffset = c13_datetimeoffset, @c14_time = c14_time,
                 @c15_char = c15_char, @c16_varchar = c16_varchar,
                 @c17_nchar = c17_nchar, @c18_nvarchar = c18_nvarchar
-                FROM $tbname";
-sqlsrv_query($conn, $spSql);
-
+                FROM $tbname");
```
v-kaywon: excellent

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_datetime.phpt
```diff
@@ -0,0 +1,137 @@
...
+
+    // create table
+    $tbname = GetTempTableName("", false);
+    $colMetaArr = array(new AE\ColumnMeta($dataType, "c_det"), new AE\ColumnMeta($dataType, "c_rand", null, false));
+    AE\createTable($conn, $tbname, $colMetaArr);
+
+    if (AE\isColEncrypted()) {
+		// Create a Store Procedure
+		$spname = 'selectAllColumns';
+		createProc($conn, $spname, "@c_det $dataType OUTPUT, @c_rand $dataType OUTPUT", "SELECT @c_det = c_det, @c_rand = c_rand FROM $tbname");
```
v-kaywon: indents

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_datetime.phpt
```diff
@@ -0,0 +1,137 @@
...
+                        die(print_r(sqlsrv_errors(), true));
+                    }							
+                    sqlsrv_execute($stmt);
+                    $errors = sqlsrv_errors();
+                    if (empty($errors)) {
+                        // SQLSRV_PHPTYPE_DATETIME not supported
+                        echo "$dataType should not be compatible with any datetime type.\n";
+                        $success = false;
+                    }
+                }
```
v-kaywon: for this datetime test specifically, you can remove the `$compatList` and also the `if (!AE\isColEncrypted())` block since outputing datetime is not supported and the behaviour with and without AE should behave the same.

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_money.phpt
```diff
@@ -0,0 +1,113 @@
...
+		
+    // insert a row
+    $inputValues = array_slice(${explode("(", $dataType)[0] . "_params"}, 1, 2);
+    $r;
+    $stmt = AE\insertRow($conn, $tbname, array( $colMetaArr[0]->colName => $inputValues[0], $colMetaArr[1]->colName => $inputValues[1] ), $r);
+
+    if ($r != false) {
+		echo "$dataType should not be compatible with any money type.\n";
+        $success = false;
+    }
```
v-kaywon: actually~~~ sorry I didn't see this before, but I don't think this test is necessary since the money values weren't even inserted into the encrypted money columns (since binding parameter with money types are not supported with always encrypted). So calling a stored procedure that selects all the columns from an empty table probably doesn't tell us any more useful information.

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_numeric.phpt
```diff
@@ -0,0 +1,225 @@
...
+                    
+                    sqlsrv_free_stmt($stmt); 
+                }                
+            }            
+        }
+        
+        if (!AE\isColEncrypted()) {
+            AE\fetchAll($conn, $tbname);
+        }        
+	}
```
v-kaywon: indent

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_string.phpt
```diff
@@ -0,0 +1,158 @@
...
+                if (!isDateTimeType($sqlType)) {
+                    $sqlTypeConstant = get_sqlType_constant($sqlType);
+
+                    // Call store procedure
+                    $outSql = AE\getCallProcSqlPlaceholders($spname, 2);
+                    $c_detOut = '';
+                    $c_randOut = '';
+                    $stmt = sqlsrv_prepare($conn, $outSql, 
+                        //array(array(&$c_detOut, constant($direction), null, $sqlTypeConstant),
+                        //array(&$c_randOut, constant($direction), null, $sqlTypeConstant)));
```
v-kaywon: remove comments

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_numeric.phpt
```diff
@@ -0,0 +1,225 @@
...
+                    
+                    if (!empty($errors)) {
+                        if (stripos("SQLSRV_SQLTYPE_" . $dataType, $sqlType) !== false) {
+                            var_dump(sqlsrv_errors());
+                            $success = false;               
+                        }	
+                    }
+                    else {
+                        print("c_det: " . $c_detOut . "\n");
+                        print("c_rand: " . $c_randOut . "\n");
```
v-kaywon: If you're already comparing the output with the inputValues (line 110 and 115), I don't think printing the values out here is necessary. Especially for the float values, sometimes in some platform some flavor the float returned may be slightly different from the expected output. Comparing the difference with epsilon is good enough.

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_string.phpt
```diff
@@ -0,0 +1,158 @@
...
+                    if (!empty($errors) ) {
+                        if (stripos("SQLSRV_SQLTYPE_" . $dataType, $sqlType) !== false) {
+                            var_dump(sqlsrv_errors());
+                            $success = false;                    
+                        }                
+                    }
+                    else
+                    {
+                        print("c_det: " . $c_detOut . "\n");
+                        print("c_rand: " . $c_randOut . "\n");
```
v-kaywon: again I don't think printing them is necessary as they are already being compared against the inputValues

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_string.phpt
```diff
@@ -0,0 +1,158 @@
...
+                    if (stripos($compatType, $sqlType) !== false) {
+                        $isCompatible = true;
+                    }
+                }
+                // 22018 is the SQLSTATE for any incompatible conversion errors
+                if ($isCompatible && sqlsrv_errors()[0]['SQLSTATE'] == 22018) {
+                    echo "$sqlType should be compatible with $dataType\n";
+                    $success = false;
+                }
+            } else {
```
v-kaywon: so right now when AE is not on, the stored procedure isn't called. For the scope of AE testing, this is OK, but eventually we probably want to test the conversions between different datatypes when AE is off as well. Perhaps write a TODO in the description of these tests saying that these tests be modified to make them work for without AE as well.

PR #747, opened 2018-04-13
Replace most strlen with strnlen_s

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/747)

### DIFF
--- a/source/shared/core_sqlsrv.h
+++ b/source/shared/core_sqlsrv.h
@@ -56,10 +56,9 @@
 
 // #define MultiByteToWideChar SystemLocale::ToUtf16
 
-
-
 #define stricmp strcasecmp
 #define strnicmp strncasecmp
+#define strnlen_s(s) strnlen_s(s, INT_MAX)
 
 #ifndef _WIN32
 #define GetLastError() errno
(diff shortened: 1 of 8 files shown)


source/shared/core_sqlsrv.h
```diff
@@ -56,10 +56,9 @@
 
 // #define MultiByteToWideChar SystemLocale::ToUtf16
 
-
-
 #define stricmp strcasecmp
 #define strnicmp strncasecmp
+#define strnlen_s(s) strnlen_s(s, INT_MAX)
```
david-puglielli: There are two definitions for strnlen_s? Why not put this line in Stringfunctions.h? Then you should be able to remove the definition from StringFunctions.cpp.

yitam: No this is just a macro, mainly for WIN32, because strnlen_s() is builtin. Besides, StringFunctions.h is not included in Windows build

david-puglielli: Ok I see.

david-puglielli: btw have you checked that this compiles on all platforms? I'm just a bit concerned because strnlen_s is part of the c11 standard, not c++11.

yitam: yes, I've run customized in Test Lab last Friday

PR #758, opened 2018-04-24
Fix step 3 order (debian)

jlsjonas:
You're trying to load in modules that aren't installed yet; this causes confusing errors during installation

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/758)

### DIFF
--- a/Linux-mac-install.md
+++ b/Linux-mac-install.md
@@ -140,12 +140,12 @@ locale-gen
 
 ### Step 3. Install the PHP drivers for Microsoft SQL Server
 ```
+sudo pecl install sqlsrv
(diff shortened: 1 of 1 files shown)


david-puglielli:
The error only occurs because calling php --ini (inside the backticks) for sqlsrv.so after adding pdo_sqlsrv.so to an ini file causes php to try to load pdo_sqlsrv, which of course isn't yet installed. The error message is harmless and will vanish once it is installed in the next command.

david-puglielli:
This change has been incorporated into PR #770, so we will close this PR now. Thanks for bringing this to our attention.

PR #768, opened 2018-05-06
Azure key vault support

david-puglielli:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/768)


source/sqlsrv/util.cpp
```diff
@@ -400,6 +400,22 @@ ss_error SS_ERRORS[] = {
...
+        { IMSSP, (SQLCHAR*) "Authentication method for Azure Key Vault is missing. KeyStoreAuthentication must be set to KeyVaultPassword or KeyVaultClientSecret.", -111, false }
+    },
+    {
+        SQLSRV_ERROR_AKV_NAME_MISSING,
+        { IMSSP, (SQLCHAR*) "ID for Azure Key Vault is missing. A username or client Id is required.", -112, false }
+    },
+    {
+        SQLSRV_ERROR_AKV_SECRET_MISSING,
+        { IMSSP, (SQLCHAR*) "Secret for Azure Key Vault is missing. A password or client secret is required.", -113, false }
+    },
```
yitam: See my comment above about these error messages

test/functional/pdo_sqlsrv/MsSetup.inc
```diff
@@ -44,4 +44,11 @@ $traceEnabled = false;
 $keystore = "none";         // key store provider, acceptable values are none, win, ksp, akv
 $dataEncrypted = false;     // whether data is to be encrypted
 
+// for Azure Key Vault
+$keyStoreAuthentication = 'KeyVaultPassword';   // can be KeyVaultPassword or KeyVaultClientSecret
+$principalName = 'name';                        // for use with KeyVaultPassword
+$AKVPassword = 'password';                      // for use with KeyVaultPassword
+$clientID = 'clientid';                         // for use with KeyVaultClientSecret
+$AKVSecret = 'secret';                          // for use with KeyVaultClientSecret
+
```
yitam: The above are all AKV related, so please be consistent and use the same prefix

test/functional/sqlsrv/MsSetup.inc
```diff
@@ -44,4 +44,11 @@ if (isset($_ENV['MSSQL_SERVER']) || isset($_ENV['MSSQL_USER']) || isset($_ENV['M
 $keystore = "none";         // key store provider, acceptable values are none, win, ksp, akv
 $dataEncrypted = false;     // whether data is to be encrypted
 
+// for Azure Key Vault
+$keyStoreAuthentication = 'KeyVaultPassword';   // can be KeyVaultPassword or KeyVaultClientSecret
+$principalName = 'name';                        // for use with KeyVaultPassword
+$AKVPassword = 'password';                      // for use with KeyVaultPassword
+$clientID = 'clientid';                         // for use with KeyVaultClientSecret
+$AKVSecret = 'secret';                          // for use with KeyVaultClientSecret
+
```
yitam: Please be consistent with the other placeholders in this include file (same with the corresponding one for PDO)
So, use a format like `TARGET_***`

source/shared/core_conn.cpp
```diff
@@ -1005,6 +1074,41 @@ void column_encryption_set_func::func( _In_ connection_option const* option, _In
...
+void ce_akv_str_set_func::func( _In_ connection_option const* option, _In_ zval* value, _Inout_ sqlsrv_conn* conn, _Inout_ std::string& conn_str TSRMLS_DC )
+{
+    SQLSRV_ASSERT( Z_TYPE_P( value ) == IS_STRING, "Azure Key Vault keywords accept only strings." );
+        
+    switch( option->conn_option_key )
+    {
+        case SQLSRV_CONN_OPTION_KEYSTORE_AUTHENTICATION: 
+        {
+            char *value_str = Z_STRVAL_P( value );        
+            CHECK_CUSTOM_ERROR( stricmp( value_str, "KeyVaultPassword" ) && stricmp( value_str, "KeyVaultClientSecret" ), conn, SQLSRV_ERROR_INVALID_AKV_AUTHENTICATION_OPTION )
```
yitam: Should have checked the length of the string value here (instead of checking post connection), something like what we used to have for [KSP](https://github.com/Microsoft/msphpsql/blob/v5.2.0-RC/source/shared/core_conn.cpp?utf8=%E2%9C%93#L1077):

```
    size_t value_len = Z_STRLEN_P(value);

    CHECK_CUSTOM_ERROR(value_len == 0, conn, SQLSRV_ERROR_KEYSTORE_INVALID_VALUE) {
        throw core::CoreException();
    } 
```

yitam: In addition, consider adding a member to indicate the valid AKV auth mode to the structure col_encryption_option, such that it is initialized to some INVALID value. Right here in this method you can set auth mode to either AKVCFG_AUTHMODE_PASSWORD or AKVCFG_AUTHMODE_CLIENTKEY. 

Thus no need to compare the values again in load_azure_key_vault()

source/shared/core_conn.cpp
```diff
@@ -932,6 +937,70 @@ void determine_server_version( _Inout_ sqlsrv_conn* conn TSRMLS_DC )
...
+}
+
+void configure_azure_key_vault( sqlsrv_conn* conn, BYTE config_attr, const char* config_value, size_t key_size )
+{
+    BYTE akv_data[sizeof( CEKEYSTOREDATA ) + 2048 ];
+    CEKEYSTOREDATA *pData = reinterpret_cast<CEKEYSTOREDATA*>( akv_data );
+    pData->name = L"AZURE_KEY_VAULT";
+    pData->data[0] = config_attr;
+    pData->dataSize = 1+key_size;
+
```
yitam: I don't think you actually need to add 1 to dataSize so long as the string is null-terminated. Please double check.

source/shared/core_conn.cpp
```diff
@@ -932,6 +937,70 @@ void determine_server_version( _Inout_ sqlsrv_conn* conn TSRMLS_DC )
...
+    pData->dataSize = sizeof(config_attr) + sizeof(config_value);
+    *reinterpret_cast<DWORD*>(&pData->data[1]) = config_value;
+    
+    core::SQLSetConnectAttr( conn, SQL_COPT_SS_CEKEYSTOREDATA, reinterpret_cast<SQLPOINTER>(pData), SQL_IS_POINTER );
+}
+
+void configure_azure_key_vault( sqlsrv_conn* conn, BYTE config_attr, const char* config_value, size_t key_size )
+{
+    BYTE akv_data[sizeof( CEKEYSTOREDATA ) + 2048 ];
+    CEKEYSTOREDATA *pData = reinterpret_cast<CEKEYSTOREDATA*>( akv_data );
```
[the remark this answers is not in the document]
yitam: Good point. Please double check that with ODBC team :)

source/shared/core_conn.cpp
david-puglielli: I think it is necessary, it's done in the original source and Z_STRLEN does not include the null terminator.

source/shared/core_conn.cpp
```diff
@@ -1005,6 +1066,48 @@ void column_encryption_set_func::func( _In_ connection_option const* option, _In
...
+        case SQLSRV_CONN_OPTION_KEYSTORE_AUTHENTICATION: 
+        {
+            char *value_str = Z_STRVAL_P( value );        
+            CHECK_CUSTOM_ERROR( stricmp( value_str, "KeyVaultPassword" ) && stricmp( value_str, "KeyVaultClientSecret" ), conn, SQLSRV_ERROR_INVALID_AKV_AUTHENTICATION_OPTION )
+            {
+                throw core::CoreException();
+            }
+            conn->ce_option.akv_auth = value;
+            conn->ce_option.akv_mode = stricmp( value_str, "KeyVaultPassword" ) ? AKVCFG_AUTHMODE_CLIENTKEY : AKVCFG_AUTHMODE_PASSWORD;
+            conn->ce_option.akv_required = true;
```
yitam: Setting AKV mode here please double check if indeed the user has given you the correct string... you can't just assume the other must be client key mode

source/shared/core_conn.cpp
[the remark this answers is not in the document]
yitam: Please define a constant or reuse a similar constant for 2048

source/shared/core_sqlsrv.h
```diff
@@ -1055,8 +1055,13 @@ struct stmt_option;
 // This holds the various details of column encryption. 
 struct col_encryption_option {
     bool            enabled;            // column encryption enabled, false by default
+    SQLINTEGER      akv_mode;
+    zval_auto_ptr   akv_auth;
+    zval_auto_ptr   akv_id;
```
yitam: Is akv_auth is still needed now that you have akv_mode? You can initialize akv_mode to an invalid value ?

test/functional/sqlsrv/MsHelper.inc
```diff
@@ -359,6 +359,17 @@ function connect($options = array(), $disableCE = false)
         if (isColEncrypted()) {
             $connectionOptions = array_merge($connectionOptions, array("ColumnEncryption" => "Enabled"));
         }
+        if ($keystore == 'akv') {
+            $akv_options = array("KeyStoreAuthentication"=>$AKVKeyStoreAuthentication);
```
yitam: Change keystore possible value for AKV from only `akv` to two diff choices (e.g. `akv_key`, `akv_pwd`, or something like this). Please make the corresponding changes in the Helper methods and/or constants.

test/functional/sqlsrv/MsSetup.inc
```diff
@@ -44,4 +44,11 @@ if (isset($_ENV['MSSQL_SERVER']) || isset($_ENV['MSSQL_USER']) || isset($_ENV['M
 $keystore = "none";         // key store provider, acceptable values are none, win, ksp, akv
 $dataEncrypted = false;     // whether data is to be encrypted
```
yitam: So here, possible values for $keystore can be `akv_key`, `akv_pwd` or whatever you have defined in MsHelper.inc / MsCommon-mid-refactor, etc. These two values will identify the test branches in test lab

test/functional/sqlsrv/MsSetup.inc
```diff
@@ -44,4 +44,11 @@ if (isset($_ENV['MSSQL_SERVER']) || isset($_ENV['MSSQL_USER']) || isset($_ENV['M
...
 
+// for Azure Key Vault
+$AKVKeyStoreAuthentication = 'TARGET_AKV_AUTH';   // can be KeyVaultPassword or KeyVaultClientSecret
+$AKVPrincipalName = 'TARGET_AKV_PRINCIPAL_NAME';  // for use with KeyVaultPassword
```
yitam: Thus, based on $keystore entry in MsSetup.inc, the value of $AKVKeyStoreAuthentication will be updated in Test Lab

source/shared/core_conn.cpp
david-puglielli: It's verfied a few lines previously that KeyStoreAuthentication is set to one of its two acceptable values.

PR #770, opened 2018-05-08
Updated Apache instructions, install order, fixed links

david-puglielli:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/770)


david-puglielli:
Yes, those links are no longer present in the document.

Issue #699, opened 2018-03-02
bind


fmart8:
+## Driver version or file name
 +Please tell us what the PHP driver version or file name is.
 +5.2.0RC1

 +## SQL Server version
 +Please tell us what the SQL Server version is.
 + SQL Server 2012 x64

 +## Client operating system
 +Please tell us what oprating system the client program is running on.
 +Windows 10 64 bits

 +## PHP version
 +Please tell us which version of PHP you are running.
 +7.2.2 64bits Non Thread Safe

 +## Microsoft ODBC Driver version
 +Please tell us which version of the Microsoft ODBC Driver you are using.
 +13

 +## Table schema
 +Please tell us the table schema
 +Might not be related to a specific table. Please see script.
 
+## Problem description
 +Please share more details with us.
 +Can't get output params values from a stored prcedure; the routine returns garbage or
 the script hangs.
I tried narrowing the problem down and it seams that it only happend when the procedure:
  - Has a BEGIN TRY, no matter where;
  - Has an INSERT INTO SELECT clause that queries a table (it doesn't happen with non-table queries, like SELECT 'value' )

Tested under:
   - Windows 10;
   - SQL Server 2012 x64;
   - PHP 7.2.2 64 bits.
And:
   - Windows Server 2012 R2;
   - SQL Server 2012 x64;
   - PHP 7.2.2 64 bits.

It __did work on this machine__, that runs PHP7.1:
   - Windows Server 2008 Small Business Server Service Pack 2;
   - SQL Server 2008;
   - PHP 7.1.5 64 bits.

 +## Expected behavior and actual behavior
 +Please tell us what should happen and what happened instead
 + $outParam should return a non-zero integer.
Instead, it throws an error saying "An error occurred translating string for an output param to UTF-8: O parametro está incorreto." (incorrect parameter).

 +## Repro code
 +Please share repro code with us, or tell us how to reproduce the issue.
Please try the following script:
<?php
################################################################################
################################################################################

$instance = '...';
$cfg = [
		'Database'			=> '...',
		'ConnectionPooling'	=> 1,
		'CharacterSet'		=> 'UTF-8',
		'UID'				=> '...',
		'PWD'				=> '...',
	];
$conn = sqlsrv_connect($instance, $cfg);
if( $conn === false )
{
     echo "Could not connect.\n";
     die( print_r( sqlsrv_errors(), true));
}
echo "Connected...\n";

################################################################################

// Recreate tables and routines.
[...]

PR #773, opened 2018-05-14
Added test case for Issue 699

yitam:
Related to #699

### DIFF
--- a/test/functional/sqlsrv/srv_699_out_param_integer.phpt
+++ b/test/functional/sqlsrv/srv_699_out_param_integer.phpt
@@ -1,0 +1,94 @@
+--TEST--
+GitHub issue #699 - binding integer as output parameter failed 
+--DESCRIPTION--
+This test uses the sample stored procedure provided by the user, in which an 
+error situation caused the binding to fail with an irrelevant error message about UTF-8 translation. This test proves that this issue has been fixed.
+--ENV--
+PHPT_EXEC=true
+--SKIPIF--
+<?php require('skipif.inc'); ?>
+--FILE--
+<?php
+
+if (strtoupper(substr(PHP_OS, 0, 3)) === 'LIN') {
+    // This test fails in Linux (unixODBC issue) but not Windows or mac. 
+    // See VSO 3459 for details.
+    echo "Done\n";
+    return;
+}
+
+require_once('MsCommon.inc');
+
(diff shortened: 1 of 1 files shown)


test/functional/sqlsrv/srv_699_out_param_integer.phpt
```diff
@@ -0,0 +1,94 @@
...
+This test uses the sample stored procedure provided by the user, in which an 
+error situation caused the binding to fail with an irrelevant error message about UTF-8 translation. This test proves that this issue has been fixed.
+--ENV--
+PHPT_EXEC=true
+--SKIPIF--
+<?php require('skipif.inc'); ?>
+--FILE--
+<?php
+
+if (strtoupper(substr(PHP_OS, 0, 3)) === 'LIN') {
```
david-puglielli: Is it better to put this in a skipif?

test/functional/sqlsrv/srv_699_out_param_integer.phpt
```diff
@@ -0,0 +1,94 @@
...
+$params = array(array(&$outParam, SQLSRV_PARAM_OUT));
+$stmt = sqlsrv_query($conn, $sql_callSP, $params); 
+if (!$stmt) { 
+    fatalError("Error in calling $procName\n"); 
+} 
+
+while ($res = sqlsrv_next_result($stmt)); 
+
+if ($outParam != 123) {
+    echo "The output param value $outParam is unexpected!\n";
```
david-puglielli: I think display the $outParam here so we can see immediately what it is if it fails.

test/functional/sqlsrv/srv_699_out_param_integer.phpt
yitam: no, because this is supposed to be temporary, I hope :)

test/functional/sqlsrv/srv_699_out_param_integer.phpt
yitam: yes, it is being printed in the echo stmt as _$outParm_

test/functional/sqlsrv/srv_699_out_param_integer.phpt
yitam: I changed this part to add a workaround to make the test pass in Linux

PR #774, opened 2018-05-15
Modified the test to address VSO 2915

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/774)

### DIFF
--- a/test/functional/pdo_sqlsrv/pdo_ae_output_param_decimals.phpt
+++ b/test/functional/pdo_sqlsrv/pdo_ae_output_param_decimals.phpt
@@ -133,12 +133,14 @@ function testOutputDecimals($inout)
                     // call stored procedure
                     $outSql = getCallProcSqlPlaceholders($spname, 2);
                     foreach ($pdoParamTypes as $pdoParamType) {
-                        $det = $rand = 0.0;
+                        // Do not initialize $det or $rand as empty strings 
+                        // See VSO 2915 for details. The string must be a numeric
+                        // string, and to make it work for all precisions, we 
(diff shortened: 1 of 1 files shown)


test/functional/pdo_sqlsrv/pdo_ae_output_param_decimals.phpt
```diff
@@ -133,12 +133,14 @@ function testOutputDecimals($inout)
                     // call stored procedure
                     $outSql = getCallProcSqlPlaceholders($spname, 2);
                     foreach ($pdoParamTypes as $pdoParamType) {
-                        $det = $rand = 0.0;
+                        // Do not initialize $det or $rand as empty strings 
+                        // See VSO 2915 for details. The string must be a numeric
+                        // string, and to make it work for all precisions, we 
+                        // simply set it to a single-digit string.
+                        $det = $rand = '0';
```
david-puglielli: Is this limitation in our driver or is it in PDO?

yitam: Neither, for inout param to work, the variable must be a numeric string because ODBC attempts to convert the string input to a decimal (in this test). If it was an empty string or non-numerical, ODBC will return with this error `Error converting data type nvarchar to numeric.`

Likewise, for decimal columns with precision (1,0), it can only take a single-digit string as input, or ODBC will also return a similar error. For decimal columns with higher precision, say (4,0), then you can provide a string like "123", etc. I decided to use a simple single-digit str input that works for all precisions.

PR #777, opened 2018-05-23
Fixed bigint insertion bug in x86

david-puglielli:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/777)


test/functional/pdo_sqlsrv/values.php
```diff
@@ -792,7 +792,7 @@
                       52.7878, 
                       -1.79E+308, 
                       -3.4E+38, 
-                      -987654321987654321,
+                      PHP_INT_SIZE == 4 ? "-9223372036854775807" : -9223372036854775807,
                       987654321,
```
yitam: for both 32bit and 64 bit bigint values should be quoted as strings because even though it passed in 64 bit accuracy might be lost

PR #782, opened 2018-05-25
Fixed the bvt streams tests

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/782)


david-puglielli:
Wasn't there another error in one test where t output 'Done' instead of another message, and that test failed? I think the test was fine, just the output was different from expected.

yitam:
yes @david-puglielli I've already fixed that expected output without creating a pull request 👍

PR #783, opened 2018-05-26
Added select type conversion test

david-puglielli:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/783)


test/functional/sqlsrv/sqlsrv_ae_type_conversion_select.phpt
```diff
@@ -0,0 +1,302 @@
...
+                $convError = sqlsrv_errors();
+                
+                checkErrors($convError);
+                
+                if (AE\isDataEncrypted()) {
+                    $stmtAE = sqlsrv_query($conn, $selectQueryAE[$i][$j]);
+                    $convError = sqlsrv_errors();
+                    
+                    // if the non-AE conversion fails, certainly the AE conversion
+                    // should fail but only with error 22018.
```
yitam: Please extend the comment to include the meaning of 22018 - Conversion not allowed - such that we don't need to scroll all the way up to know what it means

test/functional/sqlsrv/sqlsrv_ae_type_conversion_select.phpt
```diff
@@ -0,0 +1,302 @@
...
+                            print_r($convError);
+                            fatalError("AE conversion failed with unexpected error message. i=$i, j=$j, v=$v\n");
+                        }
+                    } elseif ($stmtAE == false and $conversionMatrixAE[$i][$j] == 'y') {
+                        print_r(sqlsrv_errors());
+                        fatalError("AE conversion failed, should have succeeded. i=$i, j=$j, v=$v\n");
+                    } elseif ($stmtAE != false and $conversionMatrixAE[$i][$j] == 'x') {
+                        fatalError("AE conversion succeeded, should have failed. i=$i, j=$j, v=$v\n");
+                    } elseif ($stmtAE != false and $conversionMatrixAE[$i][$j] == 'y') {
+                        $row = sqlsrv_fetch_array($stmt, SQLSRV_FETCH_NUMERIC);
```
yitam: Suggestion: rearrange the above if...else block such that:
```
if (!$stmtAE) {
   if ($conversionMatrixAE[$i][$j] == 'x') {
      ...
   } else { // ($conversionMatrixAE[$i][$j] == 'y') 
      ...
   }
} else { // query succeeded
   if ($conversionMatrixAE[$i][$j] == 'x') {
      ...
   } elseif ($conversionMatrixAE[$i][$j] == 'y') {
      ...
   }
}

```

test/functional/sqlsrv/sqlsrv_ae_type_conversion_select.phpt
```diff
@@ -0,0 +1,302 @@
...
+                            $row[0] = rtrim($row[0]);
+                            $rowAE[0] = rtrim($rowAE[0]);
+                        }
+                        
+                        if ($row[0] != $rowAE[0]) {
+                            echo "Values do not match! i=$i, j=$j, v=$v\n";
+                            print_r($row[0]);
+                            print_r($rowAE[0]);
+                            fatalError("Test failed, values do not match\n");
+                        }
```
yitam: Suggestion not to call fatalError() here when a test case failed. Perhaps printing what fail is enough such that the rest of test cases will continue.

test/functional/sqlsrv/sqlsrv_ae_type_conversion_select.phpt
```diff
@@ -0,0 +1,302 @@
...
+                    }
+                }
+            }
+        }
+    }
+    
+    $deleteQuery = "DELETE FROM $tableName";
+    $stmt = sqlsrv_query($conn, $deleteQuery);
```
yitam: Please use `TRUNCATE $tableName` instead
From DOCS: "Removes all rows from a table or specified partitions of a table, without logging the individual row deletions. TRUNCATE TABLE is similar to the DELETE statement with no WHERE clause; however, TRUNCATE TABLE is faster and uses fewer system and transaction log resources."

test/functional/sqlsrv/sqlsrv_ae_type_conversion_select.phpt
```diff
@@ -0,0 +1,302 @@
...
+
+// Conversion matrix for SQL types, based on the conversion chart
+// at https://www.microsoft.com/en-us/download/details.aspx?id=35834
+// i = implicit conversion
+// e = explicit conversion
+// x = conversion not allowed
+// @ = not applicable
+// c = explicit CAST required
+// m = misc
+$conversionMatrix = array(array('@','i','i','i','i','i','i','i','i','i','i','e','e','e','e','i','i','x','x','i','i','i','i','i','i','i','i','i','i','i','e','e','e','i','i'),//binary
```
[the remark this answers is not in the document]
david-puglielli: specialised xml conversions, no longer in the matrix in this test so I've removed it.

test/functional/sqlsrv/sqlsrv_ae_type_conversion_select.phpt
```diff
@@ -0,0 +1,302 @@
...
+        $testValues[] = $values[$v][$i];
+        $testValues[] = $values[$v][$i];
+    }
+
+    // Insert the data using sqlsrv_prepare()
+    $stmt = sqlsrv_prepare($conn, $insertQuery, $testValues);
+    if ($stmt == false) {
+        print_r(sqlsrv_errors());
+        fatalError("sqlsrv_prepare failed\n");
+    }
```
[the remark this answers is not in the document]
david-puglielli: As I recall, it prints only the message itself and not the code?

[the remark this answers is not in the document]
yitam: do you really need the error codes?

PR #788, opened 2018-06-05
Fixed the potential error reported by Prefast code analysis

yitam:
To fix this warning from Prefast: 
pdo_sqlsrv\pdo_dbh.cpp(1303) : warning C6385: Reading invalid data from 'last_insert_id_query':  the readable size is '184' bytes, but '2147483647' bytes may be read.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/788)

### DIFF
--- a/source/pdo_sqlsrv/pdo_dbh.cpp
+++ b/source/pdo_sqlsrv/pdo_dbh.cpp
@@ -1276,7 +1276,7 @@ char * pdo_sqlsrv_dbh_last_id( _Inout_ pdo_dbh_t *dbh, _In_z_ const char *name, 
 
     try {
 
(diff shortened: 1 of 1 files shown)


david-puglielli:
Isn't it possible that any call to strlen_s would produce this error, since we don't specify the length argument?

yitam:
@david-puglielli -- in this particular case `strnlen_s()` was called to determine how many characters are in the input string to be converted. Yet, it should never reach INT_MAX (2147483647), the limit that we use for `strnlen_s()` , because `last_insert_id_query` is either LAST_INSERT_ID_QUERY or SEQUENCE_CURRENT_VALUE_QUERY with the sequence name. However, the code change in this PR is required to get rid of this warning.

PR #801, opened 2018-06-21
 Modified AE fetch phptypes test to insert only one row at a time and loop through php types

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/801)


david-puglielli:
I wonder if we should have this test do what it did originally to make sure there are no regressions in future once odbc 17.2 is released?

yitam:
Yes @david-puglielli we can wait till after ODBC 17.2 is released.

PR #803, opened 2018-06-21
Version 5.3.0

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/803)


david-puglielli:
Should the config.m4 files have the copyright notice at the top as well? The config.w32 files have it...

yitam:
@david-puglielli I do not why the comment section was never in place. I can put one in if you like

david-puglielli:
Yes, I think it should be in the config.m4 files as well.

PR #806, opened 2018-06-26
Changed int to SQLLEN to avoid infinite loop

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/806)


source/shared/core_results.cpp
```diff
@@ -964,7 +964,7 @@ SQLRETURN binary_to_string( _Inout_ SQLCHAR* field_data, _Inout_ SQLLEN& read_so
         // to_copy contains the number of bytes to copy, so we divide the number in half (or quarter)
         // to get the number of hex digits we can copy
         SQLLEN to_copy_hex = to_copy / (2 * extra);
-        for( int i = 0; i < to_copy_hex; ++i ) {
```
david-puglielli: Why would this possibly result in an infinite loop?

yitam: Since to_copy_hex (SQLLEN) maybe 8 bytes and `i` is only 4, if the value of to_copy_hex is greater than MAX_INT, the loop will never end.

PR #807, opened 2018-06-27
Streamlined two very similar large column name tests

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/807)


david-puglielli:
Is there a sqlsrv version of this test?

yitam:
@david-puglielli I don't remember, but only these two tests failed with timeout error against Azure

PR #810, opened 2018-07-04
Fix for PHP 7.3

remicollet:
About **1st commit**, from UPGRADING.INTERNALS

```
  a. array_init() and array_init_size() are not functions anymore.
     They don't return any values.

```

About **2nd commit**, -Wformat warning are strangely not raised in PHP < 7.3, but point to real issue. The commit only fix the [-Werror=format-security] with break builds with hardened build options.

About **3rd commit**, from UPGRADING.INTERNALS

```
  k. zend_fcall_info_cache.initialized is removed. zend_fcall_info_cache is
     initialized if zend_fcall_info_cache.function_handler is set.

```

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/810)

Issue #804, opened 2018-06-25
PHP 7.3 Support


brainfoolong:
I am here just to ask for some information about 7.3 timeline. As always, i am an early tester for every new PHP release and we still require SQL support on windows :)

david-puglielli:
@brainfoolong Support for PHP 7.3 is in the pipeline but no timeline has been set yet. Please stay tuned.

brainfoolong:
Thank you for your answer.

Jan-E:
The ZVAL_INTERNAL_DTOR raised an error in PHP 7.3.0 alpha 4.
https://github.com/Microsoft/msphpsql/blob/master/source/pdo_sqlsrv/pdo_dbh.cpp#L555
or in the dev branch
https://github.com/Microsoft/msphpsql/blob/dev/source/pdo_sqlsrv/pdo_dbh.cpp#L549
```
ext\pdo_sqlsrv\pdo_dbh.cpp(549): error C2065: 'ZVAL_INTERNAL_DTOR': undeclared identifier
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\`
```
zval_internal_dtor(zvalue) is defined in
https://github.com/php/php-src/blob/master/Zend/zend_variables.h#L94

Probably the ZVAL_INTERNAL_DTOR must be changed into ZVAL_PTR_DTOR or ZVAL_INTERNAL_PTR_DTOR
https://github.com/php/php-src/blob/master/Zend/zend_variables.h#L111

Jan-E:
BTW In PHP 7.2 ZVAL_INTERNAL_DTOR still existed:
https://github.com/php/php-src/blob/PHP-7.2/Zend/zend_variables.h#L113

Jan-E:
ZVAL_INTERNAL_DTOR was removed in this commit:
https://github.com/php/php-src/commit/7eb045d31a108323e3878947419cde94e2ba8645#diff-93200b0a055c7319402b4e44c8ee8cd4

And became public in PHP 7.3.0 alpha 4. @cmb69 Undocumented as well?

yitam:
> ZVAL_INTERNAL_DTOR was removed in this commit:
> php/php-src@7eb045d#diff-93200b0a055c7319402b4e44c8ee8cd4
> And became public in PHP 7.3.0 alpha 4. @cmb69 Undocumented as well?
 
[...]

PR #822, opened 2018-07-25
Changes made to source and tests to support PHP 7.3

yitam:
See related #804 and #810

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/822)

### DIFF
--- a/source/shared/core_sqlsrv.h
+++ b/source/shared/core_sqlsrv.h
@@ -2377,10 +2377,14 @@ namespace core {
 
     inline void sqlsrv_array_init( _Inout_ sqlsrv_context& ctx, _Out_ zval* new_array TSRMLS_DC) 
     {
-        array_init(new_array); /*int zr = ::array_init(new_array);
-        CHECK_ZEND_ERROR( zr, ctx, SQLSRV_ERROR_ZEND_HASH ) {
+#if PHP_VERSION_ID < 70300
+        int zr = ::array_init(new_array);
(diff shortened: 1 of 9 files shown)


source/shared/core_sqlsrv.h
```diff
@@ -2377,10 +2377,13 @@ namespace core {
 
     inline void sqlsrv_array_init( _Inout_ sqlsrv_context& ctx, _Out_ zval* new_array TSRMLS_DC) 
     {
-        int zr = ::array_init(new_array);
```
david-puglielli: Is this necessary? What happens in PHP 7.3 if there is an error with array_init?

test/functional/sqlsrv/sqlsrv_errors.phpt
```diff
@@ -119,7 +119,7 @@ sqlsrv_close returns true even if an error happens.
     echo "Test successfully done.\n";
 ?>
 --EXPECTF--
-Warning: sqlsrv_close() expects parameter 1 to be resource, boolean given in %Ssqlsrv_errors.php on line %x
+Warning: sqlsrv_close() expects parameter 1 to be resource, bool%S given in %Ssqlsrv_errors.php on line %x
```
david-puglielli: PHP 7.3 changes he output to just `bool` and `int`?

yitam: yes

source/shared/core_sqlsrv.h
yitam: It used to be a function that returns either a SUCCESS or FAILURE but now a macro in 7.3

PR #832, opened 2018-08-17
Used git clone instead to download source from a branch of a tag

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/832)

### DIFF
--- a/buildscripts/buildtools.py
+++ b/buildscripts/buildtools.py
@@ -193,7 +193,7 @@ class BuildUtil(object):
         file.write('@CALL ROBOCOPY ' + source + ' ' + dest + ' /s /xx /xo' + os.linesep)
     
     @staticmethod
-    def download_msphpsql_source(repo, branch, dest_folder = 'Source', clean_up = True):
+    def download_msphpsql_source(repo, branch, dest_folder = 'Source'):
         """Download to *dest_folder* the msphpsql archive of the specified 
         GitHub *repo* and *branch*. The downloaded files will be removed by default.
         """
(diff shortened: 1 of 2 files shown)


buildscripts/buildtools.py
```diff
@@ -193,48 +193,33 @@ def write_lines_to_copy_source(driver, file):
...
-
-            print('Extracting ' + file + ' ...')
-            zip = zipfile.ZipFile(file)
-            zip.extractall()
-            zip.close()
-            
             msphpsqlFolder = os.path.join(temppath, 'msphpsql-' + branch)
+            
+            url = 'https://github.com/' + repo + '/msphpsql.git'
+            command = 'git clone ' + url + ' -b ' + branch + ' --single-branch --depth 1 ' + msphpsqlFolder
```
david-puglielli: git is always in the PATH, right? Is there a check for this somewhere else?

buildscripts/buildtools.py
```diff
@@ -193,48 +193,33 @@ def write_lines_to_copy_source(driver, file):
         file.write('@CALL ROBOCOPY ' + source + ' ' + dest + ' /s /xx /xo' + os.linesep)
     
     @staticmethod
-    def download_msphpsql_source(repo, branch, dest_folder = 'Source', clean_up = True):
```
david-puglielli: clean_up has been removed entirely and it's not used anywhere else?

yitam: yes, it was a default parameter, which was never used

buildscripts/buildtools.py
yitam: I'm not sure what you meant?

david-puglielli: Is there a check somewhere else in the script that git is actually installed and in the PATH?

yitam: This is meant to be a sample script, which assumes Git is in the path. FYI please check README .

PR #833, opened 2018-08-20
Modified the error handling to make it more flexible

yitam:
In most cases error messages are not long, so a small buffer usually suffices. Only with lengthy messages do we handle them differently.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/833)

### DIFF
--- a/source/shared/core_sqlsrv.h
+++ b/source/shared/core_sqlsrv.h
@@ -175,7 +175,7 @@ const int SQL_SERVER_MAX_TYPE_SIZE = 0;
 const int SQL_SERVER_MAX_PARAMS = 2100;
 // increase the maximum message length to accommodate for the long error returned for operand type clash
 // or for conversion of a long string
-const int SQL_MAX_ERROR_MESSAGE_LENGTH = SQL_MAX_MESSAGE_LENGTH * 8;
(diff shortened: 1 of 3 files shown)


source/shared/core_sqlsrv.h
```diff
@@ -1889,14 +1889,20 @@ namespace core {
         // and return a more helpful message prepended to the ODBC errors if that error occurs
         if( !SQL_SUCCEEDED( r )) {
 
-            SQLCHAR err_msg[SQL_MAX_ERROR_MESSAGE_LENGTH + 1] = {'\0'};
+            SQLCHAR err_msg[SQL_MAX_MESSAGE_LENGTH + 1] = {'\0'};
             SQLSMALLINT len = 0;
             
             SQLRETURN rtemp = ::SQLGetDiagField( stmt->handle_type(), stmt->handle(), 1, SQL_DIAG_MESSAGE_TEXT, 
-                                             err_msg, SQL_MAX_ERROR_MESSAGE_LENGTH, &len );
+                                             err_msg, SQL_MAX_MESSAGE_LENGTH, &len );
```
david-puglielli: So it's not necessary to use SQL_MAX_ERROR_MESSAGE_LENGTH here?

yitam: Yes, because the sole purpose of this check is to compare the error message to ODBC_CONNECTION_BUSY_ERROR, which is very short

PR #837, opened 2018-08-28
Adding supporting for Azure AD access token

yitam:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/837)

### DIFF
--- a/source/pdo_sqlsrv/pdo_dbh.cpp
+++ b/source/pdo_sqlsrv/pdo_dbh.cpp
@@ -38,6 +38,7 @@ namespace PDOConnOptionNames {
 
 const char Server[] = "Server";
 const char APP[] = "APP";
+const char AccessToken[] = "AccessToken";
 const char ApplicationIntent[] = "ApplicationIntent";
 const char AttachDBFileName[] = "AttachDbFileName";
 const char Authentication[] = "Authentication";
@@ -184,6 +185,15 @@ const connection_option PDO_CONN_OPTS[] = {
         sizeof( ODBCConnOptions::APP ),
         CONN_ATTR_STRING,
         conn_str_append_func::func 
+    },
+    {
+        PDOConnOptionNames::AccessToken,
+        sizeof( PDOConnOptionNames::AccessToken ),
+        SQLSRV_CONN_OPTION_ACCESS_TOKEN,
+        ODBCConnOptions::AccessToken,
+        sizeof( ODBCConnOptions::AccessToken), 
+        CONN_ATTR_STRING,
+        access_token_set_func::func
     },
     { 
         PDOConnOptionNames::ApplicationIntent,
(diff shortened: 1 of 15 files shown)


source/shared/core_conn.cpp
```diff
@@ -759,35 +765,53 @@ void build_connection_string_and_set_conn_attr( _Inout_ sqlsrv_conn* conn, _Inou
...
-            if( pwd != NULL ) {
-                escaped = core_is_conn_opt_value_escaped( pwd, strnlen_s( pwd ));
-                CHECK_CUSTOM_ERROR( !escaped, conn, SQLSRV_ERROR_UID_PWD_BRACES_NOT_ESCAPED ) {
+        // Add the server name
+        common_conn_str_append_func( ODBCConnOptions::SERVER, server, strnlen_s( server ), connection_string TSRMLS_CC );
+       
+        // if uid is not present then we use trusted connection -- but not when access token is used, because they are incompatible
+        if (!access_token_used) {
+            if (uid == NULL || strnlen_s(uid) == 0) {
+                connection_string += CONNECTION_OPTION_NO_CREDENTIALS;  //  "Trusted_Connection={Yes};"
```
david-puglielli: Where is CONNECTION_OPTION_NO_CREDENTIALS defined?

source/shared/core_conn.cpp
```diff
@@ -1172,3 +1196,54 @@ size_t core_str_zval_is_true( _Inout_ zval* value_z )
...
+    //  typedef struct AccessToken
+    //  {
+    //      unsigned int dataSize;
+    //      char data[];
+    //  } ACCESSTOKEN;
+    //
+    // NOTE: The ODBC Driver version 13.1 only supports this authentication on Windows.
+    //
+    // A valid access token byte string must be expanded so that each byte is followed by a 0 padding byte, 
+    // similar to a UCS-2 string containing only ASCII characters
```
david-puglielli: It sounds like the foregoing comments are from a Microsoft Docs document - if so please provide a link to it.

source/pdo_sqlsrv/pdo_dbh.cpp
```diff
@@ -38,6 +38,7 @@ namespace PDOConnOptionNames {
 
 const char Server[] = "Server";
 const char APP[] = "APP";
+const char AccessToken[] = "AccessToken";
```
david-puglielli: Is this an ODBC keyword as well?

source/shared/core_conn.cpp
yitam: At the beginning of the same file. It was there all along but wasn't used.

source/shared/core_conn.cpp
yitam: ok will do

source/pdo_sqlsrv/pdo_dbh.cpp
yitam: No, but similar to TraceFile and TraceOn, this one is also a pre-connection attribute.

PR #839, opened 2018-09-03
Remove Web server prerequisite

Majkl578:
Web server is not required to use the MSSQL Driver and it's definitely a _prerequisite_.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/839)

PR #841, opened 2018-09-04
Modified README re user's suggestion

yitam:
See pull request #839

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/msphpsql/841)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,8 @@ For full details on the system requirements for the drivers, see the [system req
 
 On the client machine:
 - PHP 7.0.x, 7.1.x, or 7.2.x (7.2.0 and up on Unix, 7.2.1 and up on Windows)
-- If using a Web server such as Internet Information Services (IIS) or Apache, it must be configured to run PHP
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,7 +46,7 @@ For full details on the system requirements for the drivers, see the [system req
 
 On the client machine:
 - PHP 7.0.x, 7.1.x, or 7.2.x (7.2.0 and up on Unix, 7.2.1 and up on Windows)
-- A Web server such as Internet Information Services (IIS) is required. Your Web server must be configured to run PHP
+- If using a Web server such as Internet Information Services (IIS) or Apache, it must be configured to run PHP
```
david-puglielli: This should come at the end of the list (after the ODBC driver requirement); since the ODBC driver is a genuine prerequisite, it should come before any 'if' configuration requirements.

yitam:
Done. Please check again, @david-puglielli

Majkl578:
This is still inaccurate: PHP FPM, which tends to be a majority these days, is not a webserver -> a) there is nothing to configure in the web server (rather FPM itself), b) restarting the web server won't help.

David-Engel:
@Majkl578 If PHP FPM is not a web server and the prerequisite is qualified as "If using a Web server", then the prerequisite does not apply to PHP FPM. So how is it still inaccurate?

Also, do you have data showing that PHP FPM is the majority now? (I'm genuinely curious on that point.)

Issue #648, opened 2017-12-28
PDO_SQLSRV returns everything as a string


nemchik:
## Driver version or file name
4.3.0+9905

## SQL Server version
Irrelevant

## Client operating system
Windows

## PHP version
7.1.7 x64 (From WebPI)

## Microsoft ODBC Driver version
phpinfo() shows ODBCVER | 0x0350

## Table schema
Irrelevant

## Problem description
PDO_SQLSRV returns everything as a string.
Reference the post here https://stackoverflow.com/a/13756696/1384186 by Robert Calhoun on Dec 7 '12 and the point number 2 here https://github.com/Microsoft/msphpsql/issues/60 and the first paragraph of this page https://docs.microsoft.com/en-us/sql/connect/php/how-to-retrieve-date-and-time-type-as-strings-using-the-sqlsrv-driver stating 

> It is an error to use the ReturnDatesAsStrings connection option with the PDO_SQLSRV driver.

The SQLSRV driver apparently returns data as the correct type. The PDO_SQLSRV driver should also return data as the correct type. As it is now data returned from the PDO_SQLSRV driver that should be datetime or int is returned as string and must then be converted to to the correct data type manually. 

## Expected behavior and actual behavior
Expected: PDO_SQLSRV returns data in the type matching what it is in the database.
Actual: PDO_SQLSRV returns everything as a string.

## Repro code
I am unsure where this might be in your repo.

yitam:
Yes @nemchik PDO_SQLSRV returns dates as strings. However, for numeric columns like integers or floats (except decimals or numerics because we want to preserve accuracy) you can set the attribute `PDO::SQLSRV_ATTR_FETCHES_NUMERIC_TYPE` to true in order to fetch them as numeric values:  

```    
$conn = new PDO("sqlsrv:Server=$server;database=$dbName;", $uid, $pwd, $options); 
$conn->setAttribute(PDO::SQLSRV_ATTR_FETCHES_NUMERIC_TYPE, true);
```

Hope this helps.

nemchik:
@yitam `PDO::SQLSRV_ATTR_FETCHES_NUMERIC_TYPE` accomplishes everything that I personally use except the date data types. Even with `PDO::SQLSRV_ATTR_FETCHES_NUMERIC_TYPE` dates are still returned as strings when using PDO, however dates are returned as dates with using the non PDO SQLSRV driver (as pointed out on the StackOverflow link).

[...]

PR #842, opened 2018-09-11
Feature request - new flag for pdo_sqlsrv to be able to return datetime as objects

yitam:
#648

### DIFF
--- a/source/pdo_sqlsrv/pdo_stmt.cpp
+++ b/source/pdo_sqlsrv/pdo_stmt.cpp
@@ -213,61 +213,63 @@ void meta_data_free( _Inout_ field_meta_data* meta )
 zval convert_to_zval( _In_ SQLSRV_PHPTYPE sqlsrv_php_type, _Inout_ void** in_val, _In_opt_ SQLLEN field_len )
 {
     zval out_zval;
-    ZVAL_UNDEF( &out_zval );
+    ZVAL_UNDEF(&out_zval);
 
-    switch( sqlsrv_php_type ) {
-       
-        case SQLSRV_PHPTYPE_INT:
-        case SQLSRV_PHPTYPE_FLOAT:       
-        {
-            if( *in_val == NULL ) {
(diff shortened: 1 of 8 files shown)


test/functional/pdo_sqlsrv/pdo_fetch_datetime_time_as_objects.phpt
```diff
@@ -0,0 +1,227 @@
+--TEST--
```
david-puglielli: Why do these types need to be in a separate test? The tests are very similar...

source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -339,6 +347,11 @@ void stmt_option_fetch_numeric:: operator()( _Inout_ sqlsrv_stmt* stmt, stmt_opt
     pdo_stmt->fetch_numeric = ( zend_is_true( value_z )) ? true : false;
 }
 
+void stmt_option_fetch_datetime:: operator()( _Inout_ sqlsrv_stmt* stmt, stmt_option const* /*opt*/, _In_ zval* value_z TSRMLS_DC )
+{
+    pdo_sqlsrv_stmt *pdo_stmt = static_cast<pdo_sqlsrv_stmt*>( stmt );
+    pdo_stmt->fetch_datetime = ( zend_is_true( value_z )) ? true : false;
+}
```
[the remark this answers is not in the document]
yitam: In `php_pdo_sqlsrv.h` near line 247 the constructor of pdo_sqlsrv_stmt takes the attributes from sqlsrv_conn

test/functional/pdo_sqlsrv/pdo_fetch_datetime_time_as_objects.phpt
yitam: Yes, they are very similar, only the inputs are generated differently, using some different fetch styles. Let me try to combine these two tests.

PR #854, opened 2018-09-26
Clear AKV data after setting the connection attribute or when exception is thrown

yitam:


source/shared/core_conn.cpp
```diff
@@ -1120,6 +1124,7 @@ void ce_akv_str_set_func::func(_In_ connection_option const* option, _In_ zval*
 {
     SQLSRV_ASSERT(Z_TYPE_P(value) == IS_STRING, "Azure Key Vault keywords accept only strings.");
 
+    const char *value_str = Z_STRVAL_P(value);
```
david-puglielli: I'm not clear on why this has been moved out of the switch bloc...

yitam: Because it's now used in all cases, unlike before when only one case needed it.

david-puglielli: But does this fix the problem with empty strings? Or is it calling akv_reset further up that does it?

yitam: What empty string problem? The lines below check the length and would have thrown exception, right?

david-puglielli: I'm just wondering what would have caused the apparent leak (whether it's a false positive or not), because both the previous code and your changes here should have thrown an exception. In any case, this code is more robust.

yitam: Thanks! This change ensures that the strings are freed whenever exceptions are thrown.

PR #855, opened 2018-09-28
Save meta data for the fetched result set

yitam:
Retrieve metadata upon fetching and save a copy to use it whenever possible. Move the core metadata member from pdo_stmt to the parent class to be shared in sqlsrv side too

### DIFF
--- a/source/sqlsrv/stmt.cpp
+++ b/source/sqlsrv/stmt.cpp
@@ -1832,11 +1832,11 @@ void fetch_fields_common( _Inout_ ss_sqlsrv_stmt* stmt, _In_ zend_long fetch_typ
             // The meta data field name is already null-terminated, and the field name len is correct.
             field_name_len = stmt->current_meta_data[i]->field_name_len;
             field_names[i].name = static_cast<char*>(sqlsrv_malloc(field_name_len, sizeof(char), 1));
-            memcpy_s(field_names[i].name, field_name_len * sizeof(char), stmt->current_meta_data[i]->field_name, field_name_len);
+            memcpy_s((void*)field_names[i].name, (field_name_len * sizeof(char)), (void*)stmt->current_meta_data[i]->field_name, field_name_len);
             field_names[i].name[field_name_len] = '\0';     // null terminate the field name after the memcpy
             field_names[i].len = field_name_len;            // field_name_len should not need to include the null char
         }
-		
+
         stmt->fetch_field_names = field_names;
         stmt->fetch_fields_count = num_cols;
         field_names.transferred();
--- a/source/sqlsrv/stmt.cpp
+++ b/source/sqlsrv/stmt.cpp
@@ -471,7 +471,7 @@ PHP_FUNCTION( sqlsrv_field_metadata )
 
     try {
 
-    // get the number of fields in the resultset
+    // get the number of fields in the resultset and its metadata if not exists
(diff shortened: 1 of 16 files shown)


source/sqlsrv/stmt.cpp
```diff
@@ -460,26 +472,25 @@ PHP_FUNCTION( sqlsrv_field_metadata )
     try {
 
     // get the number of fields in the resultset
-    num_cols = core::SQLNumResultCols( stmt TSRMLS_CC );
+    SQLSMALLINT num_cols = get_resultset_meta_data(stmt);
```
david-puglielli: This is where you get the full metadata and not just the number of fields, isn't it? The comment should be updated to reflect that.

source/sqlsrv/stmt.cpp
```diff
@@ -460,26 +472,25 @@ PHP_FUNCTION( sqlsrv_field_metadata )
...
         core::sqlsrv_array_init( *stmt, &field_array TSRMLS_CC );
 
-        core::sqlsrv_add_assoc_string( *stmt, &field_array, FieldMetaData::NAME, 
-                                       reinterpret_cast<char*>( core_meta_data->field_name.get() ), 0 TSRMLS_CC );
+        // add the field name to the associative array but keep a copy
+        core::sqlsrv_add_assoc_string(*stmt, &field_array, FieldMetaData::NAME,
+                                      reinterpret_cast<char*>(core_meta_data->field_name.get()), 1 TSRMLS_CC);
 
-        core_meta_data->field_name.transferred();
+        // core_meta_data->field_name.transferred();
```
david-puglielli: Why is this commented out but left in?

source/sqlsrv/stmt.cpp
```diff
@@ -1772,40 +1818,25 @@ void fetch_fields_common( _Inout_ ss_sqlsrv_stmt* stmt, _In_ zend_long fetch_typ
...
-            field_names[i].len = field_name_len + 1;
-            field_name.reset();
+        field_names = static_cast<sqlsrv_fetch_field_name*>(sqlsrv_malloc(num_cols * sizeof(sqlsrv_fetch_field_name)));
+        for (int i = 0; i < num_cols; ++i) {
+            // The meta data field name is already null-terminated, and the field name len is correct.
+            field_name_len = stmt->current_meta_data[i]->field_name_len;
+            field_names[i].name = static_cast<char*>(sqlsrv_malloc(field_name_len, sizeof(char), 1));
+            memcpy_s((void*)field_names[i].name, (field_name_len * sizeof(char)), (void*)stmt->current_meta_data[i]->field_name, field_name_len);
+            field_names[i].name[field_name_len] = '\0';     // null terminate the field name after the memcpy
+            field_names[i].len = field_name_len;            // field_name_len should not need to include the null char
```
david-puglielli: The old code seems to correct for wide char encoding. Is that no longer necessary with your changes?

yitam: Because when getting the meta data the field names have been converted if necessary

source/sqlsrv/stmt.cpp
david-puglielli: `if it exists`?

source/sqlsrv/stmt.cpp
david-puglielli: In core_sqlsrv_field_metadata?

Issue #415, opened 2017-05-31
Money, Decimal datatypes


genyslt:
Hello,

## Issue

I have several fields on sql table:

```
CREATE TABLE MyBalance
(
BalanceId INT IDENTITY PRIMARY KEY,
Total MONEY,
Percents DECIMAL(5,2)
)

go

INSERT INTO MyBalance (Total, Percents) VALUES (100, 20)

go

```
PHP Code:

```
$query = 'SELECT Total, Percents FROM MyBalance';
$stmt = sqlsrv_query($conn, $query);

$res_array = array();
	while( $row = sqlsrv_fetch_array( $stmt, SQLSRV_FETCH_NUMERIC ) ) {
		$res_array[] = $row;
	}
	sqlsrv_free_stmt($stmt);
print_r($res_array);
die();
```
The result on $res_array have some issue:

1. The money fields returns 4 digits after point:

100.0000

2. The decimal fields don't returns 0 at the start:

.00

## Question

1. The decimal field must return with 0 at start: 0.00
2. The money field must return two digits after point (IF modulus operand == 0)

Try it select on sql server management studio:

```
SELECT *
FROM MyBalance
```
At the management studio the result is good.

The result i see on the manager and on $res_array must be same.

## System info

**PHP version:** 7.1.5
**Linux version:** Ubuntu 16.04.2 LTS
**sqlsrv_client_info:**
[DriverName] => libmsodbcsql-13.1.so.8.0
[DriverODBCVer] => 03.52
[DriverVer] => 13.01.0008
[ExtensionVer] => 4.0.8.0
**sqlsrv_server_info:**
[SQLServerVersion] => 12.00.2000

v-kaywon:
@genyslt ,

The answer to both questions is that PHP does not support decimal type. All numbers that are not integers are represented as float in PHP. To preserve the precision of the decimal or money types, the values our driver return is a stringified number of exact precision that is specified when you create your table (money has a scale of 4, and your decimal type has a scale of 2).

genyslt:
It's not normal way to get ".00" number. The prices on the market also going scale of 2. 

On http://php.net/manual/en/book.mssql.php driver - every works fine. I trying to migrate from php 5.6 and mssql driver to php 7.1.5 and sqlsrv driver - and get some missunderstandings.

Can it be a configuration setting? 

Full internet writing about this problem, every programmer repairs results to show to the user , so i think this is very actual problem and driver must solve this problem.

[...]

PR #873, opened 2018-11-01
Feature request 415 for pdo_sqlsrv

yitam:
Related to #415 and #861 - the fetched values of decimal and money types are always strings in order to preserve accuracy and precision, unlike other primitive numeric types that can be retrieved as numbers.

With this newly introduced statement attribute PDO::SQLSRV_ATTR_FORMAT_DECIMALS, decimal strings can be formatted. Note that this expects an integer value from the range [0, 38] and only affects the decimal/numeric data of fetching results. No effect on other operations like insertion or update.

1. By default, decimal data will be returned with the original precision and scale.
2. The data column original scale still takes precedence – for example, if the user
specifies 3 decimal digits for a column of decimal(5,2), the result still shows only 2
decimals to the right of the dot.
3. After formatting, the missing leading zeroes will be padded.
4. The underlying data will not be altered, but formatted results may likely be rounded
up (e.g. .2954 will be displayed as 0.30 if only two decimals are to be shown)
5. Do not support output (inout) params for this feature.


test/functional/pdo_sqlsrv/pdostatement_format_decimals.phpt
```diff
@@ -0,0 +1,385 @@
...
+    // as numbers by default. When fetched as strings, no formatting is done even with
+    // the statement option FormatDecimals set
+    $epsilon = 0.001;
+    $values = array();
+    for ($i = 0; $i < 5; $i++) {
+        $n1 = rand(1, 100);
+        $n2 = rand(1, 100);
+        $neg = ($i % 2 == 0) ? -1 : 1;
+
+        $n = $neg * $n1 / $n2;
```
david-puglielli: This is going to do integer multiplication and division - it's not clear if that's what we want for this test, won't this just be zero half the time (because of the rand())?

test/functional/pdo_sqlsrv/pdostatement_format_decimals.phpt
```diff
@@ -0,0 +1,385 @@
...
+
+    $query = "SELECT CONVERT(smallmoney, $values[0]),
+                     CONVERT(money, $values[1]),
+                     CONVERT(smallmoney, $values[2]),
+                     CONVERT(money, $values[3]),
+                     CONVERT(smallmoney, $values[4]),
+                     CONVERT(money, $values[5])";
+
+    $stmt = $conn->query($query);
+    $results = $stmt->fetch(PDO::FETCH_NUM);
```
david-puglielli: Fetching the results is done in verifyMoneyValues isn't it? In any case $results is not passed to it...

test/functional/pdo_sqlsrv/pdostatement_format_decimals.phpt
yitam: No, actually the results are floats, not integers. For example, running the test in debug mode gave me these values

```
testFloatTypes: -0.059999999, -0.059999999
testFloatTypes: 1.7391304, 1.7391304
testFloatTypes: -3.2333333, -3.2333333
testFloatTypes: 0.048780487804878, 0.048780487804878
testFloatTypes: -0.56701030927835, -0.56701030927835
```

test/functional/pdo_sqlsrv/pdostatement_format_decimals.phpt
yitam: Good point! Let me check

source/shared/core_stmt.cpp
```diff
@@ -2293,27 +2303,39 @@ void finalize_output_parameters( _Inout_ sqlsrv_stmt* stmt TSRMLS_DC )
                 str_len = output_param->original_buffer_len - null_size;
             }
 
-            // if it's not in the 8 bit encodings, then it's in UTF-16
-            if( output_param->encoding != SQLSRV_ENCODING_CHAR && output_param->encoding != SQLSRV_ENCODING_BINARY ) {
-				bool converted = convert_zval_string_from_utf16(output_param->encoding, value_z, str_len);
```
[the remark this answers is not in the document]
david-puglielli: You resolved this conversation but forgot to hit reply I think.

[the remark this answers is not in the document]
yitam: This is the only place where the old function was called, which was just a wrapper, as you said. Since the converted string may need to be formatted, it is done before copying to the `zval` value.

PR #905, opened 2018-12-10
Drop tests related to fake passwords

yitam:


david-puglielli:
Have we added a task to the backlog to rework these tests in the future?

yitam:
yes @david-puglielli there exists a backlog item to redesign these tests

Issue #900, opened 2018-12-07
Using output parameter results in invalid data from client memory when not assigned in SP


mathieuk:
## PHP Driver version or file name
5.3.0

 ## SQL Server version
SQL2016
 
 ## Client operating system
 Linux / CentOS 7

 ## PHP version
PHP 7.2.11
 
 ## Microsoft ODBC Driver version
unixODBC 2.3.6
ODBC Driver 17 for SQL Server
 
## Problem description
When calling a stored procedure with a string output that remains untouched during the stored procedure, the driver will return memory that it shouldn't (but from a specific location).

## Expected behavior and actual behavior
Expected:
When the output parameter is not assigned it should become NULL. 

Actual:
Memory from a specific location is returned which may return sensitive information. For instance: I've seen my PHP code (from unrelated bits) returned. 

## Repro code or steps to reproduce
 
```php
<?php
$pdo = new PDO(/* ... */);
$pdo->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);

$pdo->query("CREATE OR ALTER PROCEDURE nullOutputBug(@OUTPUT NVARCHAR(255) OUTPUT) AS SELECT 1,2,3");

$output = null;

$stmt = $pdo->prepare("nullOutputBug @OUTPUT = :output");
$stmt->bindParam('output', $output, PDO::PARAM_STR, 32);

$result = $stmt->execute();


echo "========================\n";
echo "Output variable now contains:\n\n";

echo $output;
echo "\n\n========================\n";

?>
```

This now consistently returns the following for me:

```
$ php test.php
========================
Output variable now contains:

xecutetputBug(@OUTPUT NVARCHAR(255) OUTPUT) AS SELECT 1,2,3

========================
```

Varying the parameter length will change the location it reads from and the length of data returned. As said, I've had chunks of my PHP code returned but also (parts) of my database name.

yitam:
Thanks @mathieuk yes I can reproduce this.

yitam:
@mathieuk in your particular example when a result set is waiting to be processed (`SELECT 1, 2, 3`), the output param is not set until [closeCursor](https://docs.microsoft.com/sql/connect/php/pdostatement-closecursor) is called to finalize the output parameter's value. FYI, this page explains why [closing the cursor ](https://docs.microsoft.com/sql/odbc/reference/develop-app/closing-the-cursor) is required

[...]

PR #907, opened 2018-12-11
Initialize output param buffer when allocating extra space

yitam:
See related issue #900

### DIFF
--- a/test/functional/sqlsrv/sqlsrv_900_output_param_memory_data.phpt
+++ b/test/functional/sqlsrv/sqlsrv_900_output_param_memory_data.phpt
@@ -1,0 +1,81 @@
+--TEST--
+GitHub issue 900 - output parameter displays data from memory when not finalized
+--DESCRIPTION--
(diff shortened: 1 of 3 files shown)


test/functional/sqlsrv/sqlsrv_900_output_param_memory_data.phpt
```diff
@@ -0,0 +1,81 @@
...
+    if (!sqlsrv_execute($stmt)) {
+        fatalError("getOutputParam: failed to execute procedure $storedProcName");
+    }
+
+    // The output param should be doubled in size for wide characters;
+    // for max fields it should be the maximum anticipated size 8000.
+    // However, it should not contain any data so after trimming it
+    // should be merely an empty string
+    $len = strlen($output);
+    $expectedLen = ($isMax)? 8000 :  ($size * 2);
```
david-puglielli: Why is there no comparison to the max size of 8000 in the pdo test?

yitam: good point! I should redesign the sqlsrv test to fix the size like the pdo test

PR #911, opened 2019-01-03
Dropped dbname variable and set QUOTED_IDENTIFIER to ON

yitam:
To prepare these scripts for future development purpose:
1. Modify the scripts to SET QUOTED_IDENTIFIER to 'ON' -- run sqlcmd with -I and bcp with -q
1. Also drop 'Use DbName' or the use of dbname variable

### DIFF
--- a/test/functional/setup/168256.sql
+++ b/test/functional/setup/168256.sql
@@ -1,6 +1,3 @@
-USE $(dbname) 
-GO
-
 IF EXISTS (SELECT * FROM sys.objects 
 WHERE object_id = OBJECT_ID(N'[dbo].[168256]') AND type in (N'U'))
 
--- a/test/functional/setup/exec_sql_scripts.py
+++ b/test/functional/setup/exec_sql_scripts.py
@@ -1,10 +1,7 @@
 #!/usr/bin/env python3
-# contains helper methods  
+# contains helper methods
 import os
-import sys
 import subprocess
-import platform
-import argparse
 from subprocess import Popen, PIPE
 
 def executeCommmand(inst_command):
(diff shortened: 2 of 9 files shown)


test/functional/setup/setup_dbs.py
```diff
@@ -1,21 +1,15 @@
...
 import argparse
-from subprocess import Popen, PIPE
 from exec_sql_scripts import *
 
 def setupTestDatabase(conn_options, dbname, azure):
     sqlFiles = ['test_types.sql', '168256.sql', 'cd_info.sql', 'tracks.sql']
-    
-    # for Azure, must specify the database for the sql scripts to work
-    if (azure.lower() == 'yes'):
-        conn_options += ' -d ' + dbname
```
david-puglielli: Why is this not needed anymore?

test/functional/setup/168256.sql
```diff
@@ -1,6 +1,3 @@
-USE $(dbname) 
-GO
-
```
david-puglielli: All of these changes are to make the scripts work with DW - but then we are using these scripts for all the other platforms we test on, correct? I assume we've tested these scripts in Bamboo?

yitam: With this change we no longer need the clumsy workaround outside Windows platforms

test/functional/setup/exec_sql_scripts.py
```diff
@@ -15,29 +12,17 @@ def executeCommmand(inst_command):
...
 
 def executeSQLscript(sqlfile, conn_options, dbname):
-    if platform.system() == 'Windows':
-        executeSQLscriptWindows(sqlfile, conn_options, dbname)
-    elif platform.system() == 'Linux' or platform.system() == 'Darwin':
-        executeSQLscriptUnix(sqlfile, conn_options, dbname)
-    
-def executeSQLscriptWindows(sqlfile, conn_options, dbname):
-    inst_command  = 'sqlcmd ' + conn_options + ' -i ' + sqlfile + ' -v dbname =' + dbname
+    inst_command  = 'sqlcmd -I ' + conn_options + ' -i ' + sqlfile + ' -d ' + dbname
```
yitam: The variable dbname is always appended at the end of `sqlcmd`

test/functional/setup/setup_dbs.py
yitam: Please see my other comment

PR #913, opened 2019-01-04
Skipped the non-applicables tests against Azure Data Warehouse

yitam:
Skipped tests that depend on the existence of pre-populated tables and modified the tests that do not have to rely on any of those tables. Also skipped those tests that have to test max types or types that are not supported in Azure Data Warehouse: 

https://docs.microsoft.com/en-us/azure/sql-data-warehouse/sql-data-warehouse-tables-data-types


test/functional/sqlsrv/test_warning_errors2.phpt
```diff
@@ -15,7 +15,7 @@ if( $conn === false ) {
     die( print_r( sqlsrv_errors(), true ));
 }
 
-$stmt = sqlsrv_prepare( $conn, "SELECT * FROM [cd_info]");
```
david-puglielli: Were there only a few tests that used cd_info and test_streamable_types? The three or four which are modifed here can't be the only ones can they?

yitam: You're right there are not many of those. Though there are other pre-populated tables used in sqlsrv tests, for pdo side there are less than a handful.

test/functional/pdo_sqlsrv/pdo_azure_ad_authentication.phpt
```diff
@@ -12,47 +12,41 @@ require_once("MsSetup.inc");
...
     echo "Query failed.\n";
-}
-else
-{
-    $result = $stmt->fetch();
-    var_dump( $result );
+} else {
+    $result = $stmt->fetch(PDO::FETCH_NUM);
+    $edition = $result[0];
+    var_dump($edition);
```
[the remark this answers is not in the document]
yitam: I just used a random query that guarantees to return numeric string of a single digit in all platforms

PR #918, opened 2019-01-15
Changed version 5.6.0

yitam:


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright(c) 2018 Microsoft Corporation
+Copyright(c) 2019 Microsoft Corporation
```
david-puglielli: Is this the only place the year is found?

PR #920, opened 2019-01-25
Added new tests for setting client buffer size related to issue 228

yitam:


test/functional/sqlsrv/srv_228_sqlsrv_clientbuffermaxkbsize.phpt
```diff
@@ -62,19 +80,24 @@ $size = 2;
...
+dropTable($conn, $tableName1);
+dropTable($conn, $tableName2);
+
 sqlsrv_close($conn);
 print "Done"
 ?>
 
 --EXPECT--
-bool(false)
+Setting for ClientBufferMaxKBSize was non-int or non-positive.
```
david-puglielli: This error message is not the same as the one in the PDO tests above. Is this correct? It looks like it's the same error, just the message is slightly different.

yitam: Yes they are different. The error messages have always been there just no tests for checking them.

PR #926, opened 2019-02-07
Modified linux and mac instructions for 5.6.0 RTW

david-puglielli:


david-puglielli:
Yes, it's identical except for the links.

PR #930, opened 2019-02-13
add Language option on connect

gdegoulet:


david-puglielli:
Thanks for this PR @gdegoulet. Please be aware of how the language functionality works: 

There are three layers of messages that may be returned when using the PHP drivers. One layer comes from the PHP driver, and because the PHP driver has not been localised, those messages will always be in English. Another layer comes from the ODBC driver (and driver manager). To get messages from the ODBC driver in other languages, you need to use a localised version of the driver in the language you prefer. These localised versions are available to download from the Microsoft Download Center. The third layer is messages that come from SQL Server itself. When the `Language` option is passed to the ODBC driver, it affects only the messages received from SQL Server. So if the user connects with the `Language` option, messages may be returned in multiple languages because they could come from any of the three layers.

However, this is not much different from the current user experience with the ODBC driver, since you would still get messages in different languages if you pass a different language as a connection attribute than a localised version of the ODBC driver uses. The difference is that with your PR, messages from SQL Server can be set directly from a PHP script, but messages from the PHP driver would still be in English.

We want to gauge how useful adding a `Language` option is - so, considering all the info above, do you believe the `Language` option should be added to the PHP driver?

gdegoulet:
Hello David : for an very old website (LAMP) , we used php 5.6 with freetds driver.
Now we are working on a new release of our website on php 7.2/7.3 with the new sqlsrv driver : we encountered problem with date format on insert operation : our first workaround was to set language after connect to mssql server : something like this : <?php  sqlsrv_query($this->bdd, 'SET LANGUAGE us_english'); ?>

you said "When the Language option is passed to the ODBC driver, it affects only the messages received from SQL Server" : not only, it seems to change the database behaviours on date string, number format, ... ?

After some search, we discovered that there is an language option (and some other) on the connect function from sqlsrv driver but not in php extension ?
You ask "why should we add this option ?" I would say "why not ?" why all useful driver options are not present on the php extension side ?

I'm not mssql server expert : it's maybe not the good way to solve our problems ? look forward to reading your tips.

Thanks

david-puglielli:
Ok we will merge this PR shortly, but not before the next preview release 5.6.1, which fixes a problem with the PECL packaging. That will be in the next few days most likely. Please stay tuned.

Issue #937, opened 2019-02-25
SQL_ASSERT(sqlsrv_field_metadata): 'Meta data vector out of sync'


obmsch:
(SQLServer(2016)/Win10(1809)/PHP(7.2.15)/MSODBC(17.3.1.1)

This started with 5.6 (5.3 had no issues). This is on a multi layer application (MantisBT->ADOdb->sqlsvr), so getting the real source of this is not obviuos. The triggering sequence ($sql = 'UPDATE ...'):
```
  $stm = sqlsrv_query($connection, $sql);   [Ok]
  $fieldmeta = sqlsrv_field_metadata($stmt) [SQL_ASSERTs]
```
Looking at the docs:
// sqlsrv_field_metadata( resource $stmt )
// 
// Retrieves metadata for the fields of a prepared statement. For information
// about preparing a statement, see sqlsrv_query or sqlsrv_prepare. Note that
// sqlsrv_field_metadata can be called on any prepared statement, pre- or
// post-execution.
//
// Parameters
// $stmt: A statement resource for which field metadata is sought.
//
// Return Value
// retrieve an array of metadata for the current result set on a statement.  Each element of the 
// array is a sub-array containing 5 elements accessed by key:
//  name - name of the field.
//  type - integer of the type.  Can be compared against the SQLSRV_SQLTYPE constants.
//  size - length of the field.  null if the field uses precision and scale instead.
//  precision - number of digits in a numeric field.  null if the field uses size.
//  scale - number of decimal digits in a numeric field.  null if the field uses sizes.
//  is_nullable - if the field may contain a NULL instead of a value
// false is returned if an error occurs retrieving the metadata

This looks like an API (contract) violation. Neither documented, nor deprecated or warned on.

yitam:
Yes @obmsch we can reproduce this with UPDATE / DELETE statements that return no fields. We will fix this. That being said, is there any reason for calling sqlsrv_field_metadata() for UPDATE / DELETE statements?

mrbarletta:
Thanks @yitam for your quick response!. let me chime in:

We had this issue as well first with Store Procedures that return no data and also with updates, in my case it was PHP ADODB 5.21 trying to do that automatically after a query execution, we spent hours debugging and finally found this call to be the one failing (without any errors).

We used version 5.6 on Ubuntu 18.04 with php 7.2, we started receiving 500 errors without any trace of what could have gone wrong.

[...]

PR #940, opened 2019-02-27
Issue 937 - fixed ASSERT and added new tests

yitam:
See #937

### DIFF
--- a/test/functional/pdo_sqlsrv/pdo_937_metadata.phpt
+++ b/test/functional/pdo_sqlsrv/pdo_937_metadata.phpt
@@ -1,0 +1,115 @@
+--TEST--
+GitHub issue 937 - getting metadata will not fail after an UPDATE / DELETE statement
+--DESCRIPTION--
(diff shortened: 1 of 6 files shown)


test/functional/pdo_sqlsrv/pdo_937_metadata.phpt
```diff
@@ -0,0 +1,115 @@
...
+    $stmt = $conn->prepare($tsql);
+    $stmt->execute();
+    $numCol = $metadata = $stmt->columnCount();
+    echo "Number of columns after UPDATE: $numCol\n";
+
+    $tsql = "SELECT * FROM $tableName";
+    $stmt = $conn->query($tsql);
+    $numCol = $metadata = $stmt->columnCount();
+    for ($i = 0; $i < $numCol; $i++) {
+        $metadata = $stmt->getColumnMeta($i);
```
david-puglielli: This is the only place that getColumnMeta is called? In the sqlsrv test sqlsrv_field_metadata is called several times for both non-empty and empty result sets. Is it preferable to call columnCount first in PDO?

PR #949, opened 2019-03-07
Onboarding to Azure Pipelines

yitam:


azure-pipelines.yml
```diff
@@ -0,0 +1,309 @@
...
+      cp php-sdk\phpdev\vc15\x64\php-${env:VERSION}-src\run-tests.php $(Build.SourcesDirectory)\test\functional\pdo_sqlsrv
+      dir *sqlsrv*.dll
+      cp *sqlsrv*.dll C:\tools\php\ext\
+    displayName: 'Build drivers for the latest version of PHP $(phpVersion)'
+  
+  - script: |
+      echo extension=php_sqlsrv.dll >> C:\tools\php\php.ini
+      echo extension=php_pdo_sqlsrv.dll >> C:\tools\php\php.ini
+      php --ri sqlsrv
+      php --ri pdo_sqlsrv
```
david-puglielli: Is this the correct php? Does the VM have php already installed?

azure-pipelines.yml
```diff
@@ -0,0 +1,309 @@
...
+      python setup_dbs.py -dbname $(sqlsrv_db)
+      python setup_dbs.py -dbname $(pdo_sqlsrv_db)
+    displayName: 'Set up test databases'
+    condition: false
+
+  - script: |
+      cd $(Build.SourcesDirectory)\test\functional\sqlsrv
+      php run-tests.php -P sqlsrv_client_info.phpt
+      cd $(Build.SourcesDirectory)\test\functional\pdo_sqlsrv
+      php run-tests.php -P pdo_getAttribute_clientInfo.phpt
```
david-puglielli: Only running these two tests? Why are we going to the trouble of setting up test databases if we just do smoke testing?

azure-pipelines.yml
```diff
@@ -0,0 +1,309 @@
...
+      sudo update-alternatives --set phpdbg /usr/bin/phpdbg$(phpVersion)
+      sudo update-alternatives --set php-cgi /usr/bin/php-cgi$(phpVersion)
+      sudo update-alternatives --set phar.phar /usr/bin/phar.phar$(phpVersion)
+      php -version
+    displayName: 'Use PHP version $(phpVersion)'
+
+  - script: |
+      echo install ODBC and dependencies
+      sudo apt-get purge unixodbc
+      sudo sudo apt autoremove
```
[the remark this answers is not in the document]
yitam: Thanks I will fix that

azure-pipelines.yml
```diff
@@ -0,0 +1,309 @@
...
+  - script: |
+      brew tap
+      brew tap homebrew/core
+      brew install autoconf automake libtool
+      brew install php@$(phpVersion)
+      php -v
+    displayName: 'Install PHP'
+
+  - script: |
+      echo ready to build extensions
```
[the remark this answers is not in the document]
yitam: because no plan to run tests in macOS

azure-pipelines.yml
```diff
@@ -0,0 +1,309 @@
...
+    displayName: 'Run SQL Server for Linux'
+
+  - script: |
+      sudo sed -i 's/# en_US ISO-8859-1/en_US ISO-8859-1/g' /etc/locale.gen
+      sudo locale-gen en_US
+      sudo locale-gen en_US.UTF-8
+      export LANG='en_US.UTF-8' 
+      export LANGUAGE='en_US:en' 
+      export LC_ALL='en_US.UTF-8'
+    displayName: 'Add locale iso-8859-1'
```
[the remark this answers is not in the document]
yitam: Maybe I will rename this step, which uncomments ISO-8859 as well as generates the locales

azure-pipelines.yml
yitam: Yes, it has php already installed, as in the Linux agent

azure-pipelines.yml
yitam: The smoke tests were there to make sure the extensions were loaded correctly. That being said, in Windows the conditions are mostly false, meaning to skip them. Those steps were kept for further improvements in the future

david-puglielli: ok that's fine.

Issue #570, opened 2017-10-18
Fetching a varbinary field as a stream using client buffer in sqlsrv gives "Invalid cursor state" error


yitam:
Tested in Windows with the latest sqlsrv driver (tested with fields like varbinary(512) or varbinary(max)). Without using client buffer, we can save the output to a jpeg file and then open it with any photo app. However, when using client buffer, the error is 
```
Array
(
    [0] => Array
        (
            [0] => 24000
            [SQLSTATE] => 24000
            [1] => 0
            [code] => 0
            [2] => [Microsoft][ODBC Driver Manager] Invalid cursor state
            [message] => [Microsoft][ODBC Driver Manager] Invalid cursor state
        )

)
```

This is the repro:

```
$connectionInfo = array("Database"=>"AdventureWorks2014", "UID"=>$uid, "PWD"=>$pwd);
$conn = sqlsrv_connect($server, $connectionInfo);
$tsql = "SELECT ThumbNailPhoto FROM Production.ProductPhoto WHERE ProductPhotoID = ?";
$productPhotoID = 70;
$params = array($productPhotoID);

$stmt = sqlsrv_prepare($conn, $tsql, $params, array("Scrollable"=>SQLSRV_CURSOR_CLIENT_BUFFERED));
if ($stmt === false) {
    echo "Error in preparing statement.";
    die(print_r(sqlsrv_errors(), true));
}
if (!sqlsrv_execute($stmt)) {
    echo "Error in statement execution.";
    die (print_r(sqlsrv_errors(), true));
}

$readAsStr = false; // if we set this to true, we can see that $field is not empty!
if (sqlsrv_fetch($stmt, SQLSRV_SCROLL_FIRST)) {
    if ($readAsStr) {
        $field = sqlsrv_get_field($stmt, 0, SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR));
        var_dump($field);
    } else {
        $image = sqlsrv_get_field($stmt, 0, SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY));
        header("Content-Type: image/jpg");
        if (!fpassthru($image)) {
            print_r(sqlsrv_errors());
        }
    }
} else {
    echo "Error in retrieving data.";
    die (print_r(sqlsrv_errors(), true));
}
```

david-puglielli:
[...]

PR #952, opened 2019-03-13
Fixed the error in Issue 570 

yitam:
See #570 for detailed description of the issue

The result set may have been buffered, so instead of calling SQLGetData() directly, call get_data() defined in the derived classes, which might either call SQLGetData() or retrieve the data already buffered.

### DIFF
--- a/test/functional/sqlsrv/srv_570_fetch_varbinary.phpt
+++ b/test/functional/sqlsrv/srv_570_fetch_varbinary.phpt
@@ -102,9 +102,6 @@ function runTest($conn, $columnType, $path)
     }
     do {
         $read = sqlsrv_send_stream_data($stmt);
-        if ($read === false) {
-            fatalError("Something went wrong trying to insert");
-        }
     } while ($read);
     sqlsrv_free_stmt($stmt);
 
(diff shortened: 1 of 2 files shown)


test/functional/sqlsrv/srv_570_fetch_varbinary.phpt
```diff
@@ -0,0 +1,146 @@
...
+
+    $params = array($data, SQLSRV_PARAM_IN, SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY));
+    $stmt = sqlsrv_query($conn, $tsql, array($params));
+    if ($stmt === false) {
+        fatalError("Failed to insert image into $tableName");
+    }
+    do {
+        $read = sqlsrv_send_stream_data($stmt);
+        if ($read === false) {
+            fatalError("Something went wrong trying to insert");
```
david-puglielli: Won't fatalError be called when running out of  stream data normally? sqlsrv_send_stream_data returns false when no more data is to be sent, but that's not a fatal error.

yitam: You're right. I'll fix this.

Issue #955, opened 2019-03-15
Packages haven't been deployed to windows.php.net


morozov:
The [5.6.0](https://windows.php.net/downloads/pecl/releases/sqlsrv/5.6.0/) directory exists on windows.php.net since February 22 but it's empty:

![image](https://user-images.githubusercontent.com/59683/54406252-9e83fe80-4696-11e9-80b6-f39376c194b5.png)

The fact that https://pecl.php.net/rest/r/sqlsrv/stable.txt points to 5.6.0 [breaks](https://ci.appveyor.com/project/doctrine/dbal/builds/23090110/job/exu86rhatneyhwb7) build automation (e.g. on [doctrine/dbal](https://github.com/doctrine/dbal)).

One of the published [build logs](https://windows.php.net/downloads/pecl/releases/sqlsrv/5.6.0/logs/php_sqlsrv-5.6.0-7.1-nts-vc14-x64-logs.zip) contains:

> C:\php-snap-build\php-src\php71\x64\php-src\configure.js(5499, 13) Microsoft JScript runtime error: 'PHP_PDO_SQLSRV' is undefined

yitam:
Thanks @morozov for bringing this to our attention. We will get back to you on this.

david-puglielli:
This should now be fixed with the latest release, 5.6.1. Please let us know if there are any issues.

morozov:
@david-puglielli thank you. The 5.6.1 builds are available but the 5.6.0 still aren't. Are they expected to be deployed as well?

david-puglielli:
5.6.1 itself is the fix for the deployment failure, so 5.6.0 will not be deployed.

david-puglielli:
Closing this issue. Please reopen if necessary.

PR #956, opened 2019-03-15
Fix issue 955 - errors building sqlsrv alone

yitam:
#955


source/sqlsrv/config.w32
```diff
@@ -27,7 +27,7 @@ if( PHP_SQLSRV != "no" ) {
     if (CHECK_LIB("odbc32.lib", "sqlsrv") && CHECK_LIB("odbccp32.lib", "sqlsrv") && 
         CHECK_LIB("version.lib", "sqlsrv") && CHECK_LIB("psapi.lib", "sqlsrv")&&
         CHECK_HEADER_ADD_INCLUDE( "core_sqlsrv.h", "CFLAGS_SQLSRV", configure_module_dirname + "\\shared")) {
-            if (PHP_PDO_SQLSRV == "no" || PHP_SQLSRV_SHARED) {
+            if (PHP_SQLSRV_SHARED || PHP_PDO_SQLSRV == "no") {
```
david-puglielli: Why would this make a difference?

yitam: When only building sqlsrv, `PHP_PDO_SQLSRV` is undefined, thus breaking the build.

PR #965, opened 2019-03-29
Added more checks for error conditions

yitam:


test/functional/sqlsrv/sqlsrv_data_to_str.phpt
```diff
@@ -59,6 +59,16 @@ large types to strings of 1MB size.
         die( "sqlsrv_get_field(6) failed." );
     }
 
+    $str = sqlsrv_get_field( $stmt, 0, SQLSRV_PHPTYPE_STRING("UTF") );
+    if ($str === false) {
+        $error = sqlsrv_errors()[0]['message'];
+        if ($error !== 'Invalid type') {
```
david-puglielli: Invalid type is not very informative... maybe we should reword the message in util.cpp?

yitam: I agree, but it may not be a good idea to modify an old error message since few versions ago.

PR #972, opened 2019-04-11
Fix for redundant calls to SQLNumResultCols and SQLRowCount

david-puglielli:
https://docs.microsoft.com/sql/odbc/reference/develop-app/was-a-result-set-created

### DIFF
--- a/source/pdo_sqlsrv/pdo_stmt.cpp
+++ b/source/pdo_sqlsrv/pdo_stmt.cpp
@@ -694,7 +694,7 @@ int pdo_sqlsrv_stmt_fetch( _Inout_ pdo_stmt_t *stmt, _In_ enum pdo_fetch_orienta
 
         // support for the PDO rowCount method.  Since rowCount doesn't call a method, PDO relies on us to fill the 
         // pdo_stmt_t::row_count member
-        if( driver_stmt->past_fetch_end || driver_stmt->cursor_type != SQL_CURSOR_FORWARD_ONLY ) {
+        if( driver_stmt->past_fetch_end) {// || driver_stmt->cursor_type != SQL_CURSOR_FORWARD_ONLY ) {
 
             stmt->row_count = core::SQLRowCount( driver_stmt TSRMLS_CC );
 
--- a/source/pdo_sqlsrv/pdo_stmt.cpp
+++ b/source/pdo_sqlsrv/pdo_stmt.cpp
@@ -595,23 +595,22 @@ int pdo_sqlsrv_stmt_execute( _Inout_ pdo_stmt_t *stmt TSRMLS_DC )
             stmt->row_count = 0;
             driver_stmt->column_count = 0;
             driver_stmt->row_count = 0;
-            driver_stmt->columns_rows_obtained = true;
         }
         else {
-            if (driver_stmt->columns_rows_obtained == false)
-            {
+            if (driver_stmt->column_count == ACTIVE_NUM_COLS_INVALID) {
                 stmt->column_count = core::SQLNumResultCols( driver_stmt TSRMLS_CC );
+                driver_stmt->column_count = stmt->column_count;
+            }
+            else {
+                stmt->column_count = driver_stmt->column_count;
+            }
 
+            if (driver_stmt->row_count == ACTIVE_NUM_ROWS_INVALID) {
                 // return the row count regardless if there are any rows or not
                 stmt->row_count = core::SQLRowCount( driver_stmt TSRMLS_CC );
-                
-                driver_stmt->column_count = stmt->column_count;
                 driver_stmt->row_count = stmt->row_count;
-                driver_stmt->columns_rows_obtained = true;
             }
-            else
-            {
-                stmt->column_count = driver_stmt->column_count;
+            else {
                 stmt->row_count = driver_stmt->row_count;
             }
         }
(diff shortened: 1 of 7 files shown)


source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -694,7 +709,7 @@ int pdo_sqlsrv_stmt_fetch( _Inout_ pdo_stmt_t *stmt, _In_ enum pdo_fetch_orienta
 
         // support for the PDO rowCount method.  Since rowCount doesn't call a method, PDO relies on us to fill the 
         // pdo_stmt_t::row_count member
-        if( driver_stmt->past_fetch_end || driver_stmt->cursor_type != SQL_CURSOR_FORWARD_ONLY ) {
+        if( driver_stmt->past_fetch_end) {
```
yitam: Please add a comment why the second OR condition is no longer required (for future references when we looked back to understand the reason of removing this)

source/shared/core_sqlsrv.h
```diff
@@ -1438,8 +1438,10 @@ struct sqlsrv_stmt : public sqlsrv_context {
     bool has_rows;                        // Has_rows is set if there are actual rows in the row set
     bool fetch_called;                    // Used by core_sqlsrv_get_field to return an informative error if fetch not yet called 
     int last_field_index;                 // last field retrieved by core_sqlsrv_get_field
-    bool past_next_result_end;            // core_sqlsrv_next_result sets this to true when the statement goes beyond the 
-                                          // last results
+    bool past_next_result_end;            // core_sqlsrv_next_result sets this to true when the statement goes beyond the last results
+    bool columns_rows_obtained;           // Whether or not SQLNumResultCols and SQLRowCount have been called for the active result
+    short column_count;                   // Number of columns in the current result set obtained from SQLNumResultCols
+    long row_count;                       // Number of rows in the current result set obtained from SQLRowCount
     unsigned long query_timeout;          // maximum allowed statement execution time
```
yitam: What's the chance of column_count or row_count ever be negative? If not, consider setting them to -1 (or any negative constant) such that you don't need the boolean variable `columns_rows_obtained`. This makes calling the two ODBC APIs independent of each other

source/shared/core_stmt.cpp
```diff
@@ -821,7 +827,13 @@ bool core_sqlsrv_fetch( _Inout_ sqlsrv_stmt* stmt, _In_ SQLSMALLINT fetch_orient
         }
         // First time only
         if ( !stmt->fetch_called ) {
-            SQLSMALLINT has_fields = core::SQLNumResultCols( stmt TSRMLS_CC );
+            SQLSMALLINT has_fields;
+            if (stmt->columns_rows_obtained) {
+                has_fields = stmt->column_count;
+            } else {
+                has_fields = core::SQLNumResultCols( stmt TSRMLS_CC );
+            }
```
yitam: See my previous suggestion. Thus, call SQLNumResultCols() only if column_count is equal to that negative constant

source/sqlsrv/util.cpp
```diff
@@ -196,7 +196,7 @@ ss_error SS_ERRORS[] = {
           
     {
         SQLSRV_ERROR_NO_FIELDS,
-        { IMSSP, (SQLCHAR*)"The active result for the query contains no fields.", -28, false }
+        { IMSSP, (SQLCHAR*)"The active result for the query contains no fields, so no result set was created.", -28, false }
     },
```
yitam: I suggest not to change the existing error message

source/sqlsrv/stmt.cpp
```diff
@@ -1791,7 +1791,11 @@ SQLSMALLINT get_resultset_meta_data(_Inout_ sqlsrv_stmt * stmt)
 
     if (num_cols == 0) {
         getMetaData = true;
-        num_cols = core::SQLNumResultCols(stmt TSRMLS_CC);
+        if (stmt->columns_rows_obtained == false) {
+            num_cols = core::SQLNumResultCols(stmt TSRMLS_CC);
+        } else {
+            num_cols = stmt->column_count;
```
yitam: See my comment above

source/shared/core_sqlsrv.h
david-puglielli: row_count can be -1 because SQLRowCount returns -1 if it can't tell how many rows are in the result. column_count is never negative but is a SQLSMALLINT so not unsigned.

yitam: In this case, consider to initialize them to any negative constant(s). My suggestions are ACTIVE_RESULT_NUM_COLS_INVALID (-99) and/or ACTIVE_RESULT_NUM_ROWS_INVALID (-99). Something like that.

source/shared/core_stmt.cpp
```diff
@@ -828,7 +826,7 @@ bool core_sqlsrv_fetch( _Inout_ sqlsrv_stmt* stmt, _In_ SQLSMALLINT fetch_orient
         // First time only
         if ( !stmt->fetch_called ) {
             SQLSMALLINT has_fields;
-            if (stmt->columns_rows_obtained) {
+            if (stmt->column_count != ACTIVE_NUM_COLS_INVALID) {
                 has_fields = stmt->column_count;
             } else {
                 has_fields = core::SQLNumResultCols( stmt TSRMLS_CC );
```
yitam: After calling SQLNumResultCols, do you set stmt->column_count? If not, when?

source/sqlsrv/stmt.cpp
yitam: Again, see my earlier comment setting stmt->column_count after calling SQLNumResultCols

PR #974, opened 2019-04-19
Added batch query test

david-puglielli:


test/functional/sqlsrv/sqlsrv_batch_query.phpt
```diff
@@ -0,0 +1,179 @@
+--TEST--
+Test a batch query with different cursor types
+--DESCRIPTION--
+Verifies that batch queries don't work with dynamic, static, and keyset
+server-side cursors, and checks that correct column and row counts are
+returned otherwise
+--SKIPIF--
```
yitam: Please include the ODBC reference page(s) to this test that provide(s) reasons why batch queries will return errors with server-side cursors

test/functional/sqlsrv/sqlsrv_batch_query.phpt
```diff
@@ -0,0 +1,179 @@
...
+<?php require('skipif.inc'); ?>
+--FILE--
```
yitam: Since you use AE helper methods please use the appropriate skipif file

test/functional/sqlsrv/sqlsrv_batch_query.phpt
```diff
@@ -0,0 +1,179 @@
...
+$expectedCols = 1;
+$expectedRows = 5;
+$expectedResultSets = 5;
+
+//expected errors
+$noCursor = array(array('01000', 16954, '[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Executing SQL directly; no cursor.'),
+                  array('01S02', 0, '[Microsoft][ODBC Driver 17 for SQL Server]Cursor type changed'));
+$wrongCursor = array(array('IMSSP', -50, 'This function only works with statements that have static or keyset scrollable cursors.'));
+$noNextResult = array(array('IMSSP', -22, 'There are no more rows in the active result set.  Since this result set is not scrollable, no more data may be retrieved.'));
+
```
yitam: Please omit `ODBC Driver 17 for SQL Server` in the expected error messages because this test should pass when testing against other ODBC versions. Thus, comparing a substring of an error message should suffice (such as using fnmatch(), etc.)

test/functional/sqlsrv/sqlsrv_batch_query.phpt
```diff
@@ -0,0 +1,179 @@
...
+    $stmt = AE\insertRow($conn, $tableName, $inputs[$i]);
+    sqlsrv_free_stmt($stmt);
+}
+
+$query = "SELECT c1_int from batch_query_test;
+          SELECT c2_tinyint from batch_query_test;
+          SELECT c3_smallint from batch_query_test;
+          SELECT c4_bigint from batch_query_test;
+          SELECT c5_bit from batch_query_test;";
+
```
yitam: Did you mean to use `$tableName` in `$query`? Also, to be consistent please use upper caps for tsql keywords

test/functional/sqlsrv/sqlsrv_batch_query.phpt
```diff
@@ -0,0 +1,179 @@
...
+        if ($cursor == 'forward' or $cursor == 'buffered') {
+            fatalError("Statement execution failed unexpectedly with a ".$cursor." cursor\n");
+        } else {
+            checkErrors($noCursor);
+            continue;
+        }
+    }
+
+    // Check the column and row count before and after running through
+    // all results
```
yitam: Please also explain why you have to check the column and row count before and after

test/functional/sqlsrv/sqlsrv_batch_query.phpt
```diff
@@ -0,0 +1,179 @@
...
+    $numResultSets = 1;
+    while ($next = sqlsrv_next_result($stmt)) {
+        checkColumnsAndRows($stmt, $cursor, $wrongCursor);
+        
+        while ($res = sqlsrv_fetch($stmt))
+        { }
+    
+        checkColumnsAndRows($stmt, $cursor, $wrongCursor);
+        ++$numResultSets;
+    }
```
yitam: The above is quite similar. Perhaps we should use `do... while ()` loop instead? Also, consider verifying the data fetched is correct (either with `sqlsrv_get_field` or `sqlsrv_fetch_array`)?

test/functional/sqlsrv/sqlsrv_batch_query.phpt
```diff
@@ -0,0 +1,179 @@
...
+
+    // We expect an error if sqlsrv_next_result returns false,
+    // but not if it returns null (i.e. if we are genuinely at
+    // the end of all the result sets with a buffered cursor)
+    if ($next === false) {
+        if ($cursor == 'forward') {
+            checkErrors($noNextResult);
+        } else {
+            fatalError("sqlsrv_next_result failed with a ".$cursor." cursor\n");
+        }
```
yitam: This comment applies to all error messages. With double quotes `$cursor` would be evaluated so you can simply do something like this 
`fatalError("sqlsrv_next_result failed with a $cursor cursor\n");`

test/functional/sqlsrv/sqlsrv_batch_query.phpt
```diff
@@ -0,0 +1,204 @@
...
+            continue;
+        }
+    }
+
+    $numResultSets = 0;
+
+    // Check the column and row count before and after running through
+    // each result set, because some cursor types may return the number
+    // of rows only after fetching all rows in the result set
+    do {
```
yitam: I'm not sure I understand the logic of checking number of rows twice.  Here the cursor is either buffered or forward, and buffered cursor should always return the right number of rows, correct? I thought forward cursors always return false?

david-puglielli: On the PDO side, one of the server side cursors would return the number of rows once it was at the end of a result set, though I don't remember if it was a forward cursor. But I could easily imagine this is the kind of thing whose behaviour might change in the future, so I think we should leave it like this so that if the behaviour does change, it will trigger an error and force us to look at it again.

test/functional/sqlsrv/sqlsrv_batch_query.phpt
```diff
@@ -0,0 +1,204 @@
...
+    {
+        $inputs[$i][$colName[$j]] = $data[$j][$i];
+    }
+}
+
+for ($i=0; $i < sizeof($inputs); ++$i) {
+    $stmt = AE\insertRow($conn, $tableName, $inputs[$i]);
+    sqlsrv_free_stmt($stmt);
+}
+
```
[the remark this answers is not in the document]
david-puglielli: The comment on this function in MsHelper.inc indicates it has to be an associative array, with the column names as keys. The other tests that call insertRow all appear to use associative arrays as well, and the function looks like it relies on the inputs being an associative array.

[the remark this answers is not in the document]
yitam: Still, `$inputs` multi array is not used again after this part of the test, so I think you can simplify the insertion part a bit

PR #977, opened 2019-04-23
Used different skipif conditions for these two tests that require AE connections

yitam:


test/functional/pdo_sqlsrv/pdo_569_query_varcharmax.phpt
```diff
@@ -5,23 +5,32 @@ Verifies that the problem is no longer reproducible.
...
 try {
-    // This test requires to connect with the Always Encrypted feature
-    // First check if the system is qualified to run this test
-    $dsn = getDSN($server, null);
-    $conn = new PDO($dsn, $uid, $pwd);
-    if (!isAEQualified($conn)) {
-        echo "Done\n";
-        return;
-    }
-    unset($conn);
```
david-puglielli: Why do these tests fail with this part included? Should they not just return here? It looks like the new skipif section does the same thing as this...

yitam: Because occasionally it didn't return and kept on going

PR #978, opened 2019-04-24
Modified get column meta method to reference saved metadata

yitam:
Also, field name will be reset in meta_data_free() when the statement gets destroyed

### DIFF
--- a/source/pdo_sqlsrv/pdo_stmt.cpp
+++ b/source/pdo_sqlsrv/pdo_stmt.cpp
@@ -478,7 +478,6 @@ int pdo_sqlsrv_stmt_describe_col( _Inout_ pdo_stmt_t *stmt, _In_ int colno TSRML
    
     // Set the name
     column_data->name = zend_string_init( (const char*)core_meta_data->field_name.get(), core_meta_data->field_name_len, 0 );
-    core_meta_data->field_name.reset();     
 
     // Set the maxlen
     column_data->maxlen = ( core_meta_data->field_precision > 0 ) ? core_meta_data->field_precision : core_meta_data->field_size;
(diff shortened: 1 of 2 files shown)


source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -1066,10 +1065,12 @@ int pdo_sqlsrv_stmt_get_col_meta( _Inout_ pdo_stmt_t *stmt, _In_ zend_long colno
         // initialize the array to nothing, as PDO requires us to create it
         core::sqlsrv_array_init( *driver_stmt, return_value TSRMLS_CC );
 
-        sqlsrv_malloc_auto_ptr<field_meta_data> core_meta_data;
-
-        core_meta_data = core_sqlsrv_field_metadata( driver_stmt, (SQLSMALLINT) colno TSRMLS_CC );
+        field_meta_data* core_meta_data;
 
+        // metadata should have been saved earlier
```
david-puglielli: How do we know the metadata is always saved before it gets to this point? Is pdo_sqlsrv_stmt_describe_col for sure called before pdo_sqlsrv_stmt_get_col_meta?

source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -1066,10 +1065,12 @@ int pdo_sqlsrv_stmt_get_col_meta( _Inout_ pdo_stmt_t *stmt, _In_ zend_long colno
         // initialize the array to nothing, as PDO requires us to create it
         core::sqlsrv_array_init( *driver_stmt, return_value TSRMLS_CC );
 
-        sqlsrv_malloc_auto_ptr<field_meta_data> core_meta_data;
-
-        core_meta_data = core_sqlsrv_field_metadata( driver_stmt, (SQLSMALLINT) colno TSRMLS_CC );
+        field_meta_data* core_meta_data;
```
david-puglielli: Why are we not using the smart pointer here?

source/pdo_sqlsrv/pdo_stmt.cpp
yitam: Yes, see pdo_stmt_describe_columns in pdo_stmt.c

source/pdo_sqlsrv/pdo_stmt.cpp
yitam: Because if we use smart pointer the entire data will be transferred. See sqlsrv_malloc_auto_ptr definition

PR #984, opened 2019-04-29
Added more pdo tests to verify different error conditions

yitam:


test/functional/pdo_sqlsrv/pdo_fetch_column_twice.phpt
```diff
@@ -0,0 +1,150 @@
...
+{
+    try {
+        $tsql = "SELECT * FROM $tableName";
+        $stmt = $conn->query($tsql);
+        $result = $stmt->fetchColumn($col);
+        echo "fetchColumnOutOfBound1: expected fetchColumn to throw an exception\n";
+        unset($stmt);
+    } catch (PDOException $e) {
+        $error1 = '*General error: Invalid column index';
+        $error2 = '*An invalid column number was specified.';
```
david-puglielli: So either error might be returned?

test/functional/pdo_sqlsrv/pdo_output_decimal.phpt
```diff
@@ -42,7 +42,7 @@ try {
     $expected1 = "7.4";
     $expected2 = "7";
     if ($outValue1 == $expected1 && $outValue2 == $expected2) {
-        echo "Test Successfully\n";
+        echo "Test Successfully done\n";
```
david-puglielli: Or just 'Test successful'?

test/functional/pdo_sqlsrv/pdo_fetch_column_twice.phpt
yitam: Yes, different error is returned depends on running with run-tests.php or not

PR #985, opened 2019-05-03
Fixed memory issues with data classification

yitam:


source/shared/core_util.cpp
```diff
@@ -485,8 +493,9 @@ namespace data_classification {
             convert_sensivity_field(stmt, encoding, idptr, idlen, (char**)&id);
             pair->id = id;
 
-            pairs.push_back(pair.get());
+            name_id_pair* a_pair = pair;
             pair.transferred();
+            pairs->push_back(a_pair);
```
david-puglielli: This seems to be the only place where converting pairs to a pointer has any effect, but it's not clear to me why this would solve the memory issues?

source/shared/core_util.cpp
```diff
@@ -493,9 +493,8 @@ namespace data_classification {
             convert_sensivity_field(stmt, encoding, idptr, idlen, (char**)&id);
             pair->id = id;
 
-            name_id_pair* a_pair = pair;
+            pairs.push_back(pair.get());
```
david-puglielli: This is not pairs->push_back?

yitam: Already corrected that in previous commit

Issue #432, opened 2017-06-09
Invalid UTF-8 name in connection string produces wrong error message in Linux


david-puglielli:
See [sqlsrv_connStr.phpt](https://github.com/Microsoft/msphpsql/blob/dev/test/sqlsrv/sqlsrv_connStr.phpt). When the server name is invalid in UTF-8, the connection attempt fails with an error message as follows:

```
Array
(
    [0] => Array
        (
            [0] => HYT00
            [SQLSTATE] => HYT00
            [1] => 0
            [code] => 0
            [2] => [unixODBC][Microsoft][ODBC Driver 13 for SQL Server]Login timeout expired
            [message] => [unixODBC][Microsoft][ODBC Driver 13 for SQL Server]Login timeout expired
        )

    [1] => Array
        (
            [0] => 08001
            [SQLSTATE] => 08001
            [1] => 11002
            [code] => 11002
            [2] => [unixODBC][Microsoft][ODBC Driver 13 for SQL Server]TCP Provider: Error code 0x2AFA
            [message] => [unixODBC][Microsoft][ODBC Driver 13 for SQL Server]TCP Provider: Error code 0x2AFA
        )

    [2] => Array
        (
            [0] => 08001
            [SQLSTATE] => 08001
            [1] => 11002
            [code] => 11002
            [2] => [unixODBC][Microsoft][ODBC Driver 13 for SQL Server]A network-related or instance-specific error has occurred while establishing a connection to SQL Server. Server is not found or not accessible. Check if instance name is correct and if SQL Server is configured to allow remote connections. For more information see SQL Server Books Online.
            [message] => [unixODBC][Microsoft][ODBC Driver 13 for SQL Server]A network-related or instance-specific error has occurred while establishing a connection to SQL Server. Server is not found or not accessible. Check if instance name is correct and if SQL Server is configured to allow remote connections. For more information see SQL Server Books Online.
        )

)
```

This implies that the error is network-related, when in fact it is because the server name is not valid. The expected error message should correctly indicate that `An error occurred translating the connection string to UTF-16`. This may be fixed by replacing `ToUtf16()` with `ToUtf16Strict()` in `convert_string_from_default_encoding()` in [core_util.cpp](https://github.com/Microsoft/msphpsql/blob/dev/source/shared/core_util.cpp#L388), but this change causes other test failures so further investigation is required.

david-puglielli:
This has been fixed. Closing issue.

PR #989, opened 2019-05-09
Issue 432 fix

david-puglielli:


david-puglielli:
So should we add use_strict to the other conversion functions as well (like convert_string_from_utf16)?

yitam:
> So should we add use_strict to the other conversion functions as well (like convert_string_from_utf16)?

Do you think it's necessary at this point? Or something we can revisit as needed?

david-puglielli:
I say revisit as needed, no need to add it yet.

test/functional/sqlsrv/sqlsrv_connStr.phpt
```diff
@@ -77,89 +95,10 @@ $c = connect(array(
 if ($c !== false) {
     fatalError("sqlsrv_connect(4) should have failed");
 }
-print_r(sqlsrv_errors());
+checkErrors($utf16Error);
 
 echo "Test succeeded.\n";
 
 ?>
 --EXPECTF--
```
yitam: Glad you finally fixed this test, but one minor point. Please change from EXPECTF to EXPECT

david-puglielli:
Issue #432 fixed.

PR #991, opened 2019-05-13
Fixed two failing tests

yitam:


yitam:
> What is the problem with pdo_035?

It failed in AE test when the first column was supposed to be integer but the original test tried to bind param with binary encoding

PR #1013, opened 2019-07-24
fix for 7.4.0beta1

remicollet:
From UPGRADE.INTERNALS

```
 y. The read and write operations of php_stream_ops now return ssize_t, with
     negative values indicating an error.
```

This si for beta1 which will be announced tomorrow


nikic:
While this fixes the compilation error, I think some additional changes are needed to properly propagate errors. In particular the two `return 0`s in the catch blocks should become `return -1` to indicate that an error happened.

yitam:
Hi @remicollet 

Many thanks for your contribution but as @nikic mentioned above, we will fix this properly to support PHP 7.4 beta. Also, we do not plan to release a preview just yet, because the preview will include new feature(s) and bug fixes.

yitam:
For your information @remicollet we will close this pull request. For more details please see the related pull request #1015. Thanks!

Issue #999, opened 2019-06-05
PHP 7.4 support


remicollet:
For now, build fails

```
/dev/shm/BUILD/php74-php-sqlsrv-5.6.1/NTS/sqlsrv/shared/core_sqlsrv.h: In function 'void core::sqlsrv_add_index_zval(sqlsrv_context&, zval*, zend_ulong, zval*)':
/opt/remi/php74/root/usr/include/php/Zend/zend_API.h:421:2: error: expected id-expression before '(' token
  (zend_hash_index_update(Z_ARRVAL_P(arg), index, value) ? SUCCESS : FAILURE)
  ^
/dev/shm/BUILD/php74-php-sqlsrv-5.6.1/NTS/sqlsrv/shared/core_sqlsrv.h:2405:20: note: in expansion of macro 'add_index_zval'
         int zr = ::add_index_zval( array, index, value );
                    ^~~~~~~~~~~~~~
/dev/shm/BUILD/php74-php-sqlsrv-5.6.1/NTS/sqlsrv/shared/core_sqlsrv.h: In function 'void core::sqlsrv_add_next_index_zval(sqlsrv_context&, zval*, zval*)':
/opt/remi/php74/root/usr/include/php/Zend/zend_API.h:433:2: error: expected id-expression before '(' token
  (zend_hash_next_index_insert(Z_ARRVAL_P(arg), value) ? SUCCESS : FAILURE)
  ^
/dev/shm/BUILD/php74-php-sqlsrv-5.6.1/NTS/sqlsrv/shared/core_sqlsrv.h:2413:20: note: in expansion of macro 'add_next_index_zval'
         int zr = ::add_next_index_zval( array, value );
                    ^~~~~~~~~~~~~~~~~~~

```

cmb69:
`add_next_index_zval` has been [replaced by a macro](https://github.com/php/php-src/commit/85e1df6fdcaf307c5e6b241df4d859e916806d2a). That doesn't seem to be mentioned in [UPGRADING.INTERNALS](https://github.com/php/php-src/blob/PHP-7.4/UPGRADING.INTERNALS), though.

yitam:
Thanks @remicollet and @cmb69 
We will investigate and get back to you.

cmb69:
This issue *might* get resolved on the PHP side, see php/php-src#4250.

yitam:
Thanks @cmb69 for the reference! We will keep an eye on this.

remicollet:
[...]

PR #1015, opened 2019-07-25
Addressed various issues with PHP 7.4 beta1

yitam:
Related to #999

### DIFF
--- a/source/shared/core_stmt.cpp
+++ b/source/shared/core_stmt.cpp
@@ -988,7 +988,7 @@ void core_sqlsrv_sensitivity_metadata( _Inout_ sqlsrv_stmt* stmt TSRMLS_DC )
             return;
         }
 
-        if (stmt->current_sensitivity_metadata != NULL) {
+        if (stmt->current_sensitivity_metadata) {
             // Already cached, so return
             return;
         }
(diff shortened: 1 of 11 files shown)


source/shared/core_stmt.cpp
```diff
@@ -2428,6 +2428,7 @@ void finalize_output_parameters( _Inout_ sqlsrv_stmt* stmt TSRMLS_DC )
     zend_string* key = NULL;
     void* output_param_temp = NULL;
 
+    try {
     ZEND_HASH_FOREACH_KEY_PTR( params_ht, index, key, output_param_temp ) {
```
david-puglielli: This block should now be indented.

source/shared/core_stmt.cpp
```diff
@@ -2553,9 +2554,15 @@ void finalize_output_parameters( _Inout_ sqlsrv_stmt* stmt TSRMLS_DC )
         }
         value_z = NULL;
     }  ZEND_HASH_FOREACH_END();
+    }
+    catch (core::CoreException&) {
+        // empty the hash table since it's been processed
+        zend_hash_clean(Z_ARRVAL(stmt->output_params));
 
+        throw;
```
david-puglielli: Should we DIE() here or just throw?

yitam: No we should throw rather than die here because it might not be a fatal error at this point. Let the calling function(s) handle the exception(s)

source/shared/core_stmt.cpp
yitam: I originally did that but the changes are hard to spot, so I reverted and decided to make cosmetic changes later

Issue #1027, opened 2019-08-30
Specifying `'QueryTimeout'` issues a `LOCK_TIMEOUT` on the server


gjdanis:
In the pdo_sqlsrv and sqlsrv drivers it looks like setting `'QueryTimeout'` actually issues a `LOCK_TIMEOUT` command on the server. Here's an example with PDO:

```
$pdo = new PDO(...);
$pdo->setAttribute(PDO::SQLSRV_ATTR_QUERY_TIMEOUT, 30);
$pdo->exec(
  '
    WAITFOR DELAY \'00:00:40\'; SELECT 1
  '
);
```

This likely isn't what we want as read queries that don't take any locks and exceed this timeout won't be killed.

Is there a recommended approach to ensure that read queries don't exceed a user specified query timeout?

gjdanis:
It also looks like every statement will produce a command to `SET LOCK_TIMEOUT` resulting in an extra roundtrip to the server. Is it possible to make this part of the DSN so that the command need only be issued once but can be overridden by an individual statement?

yitam:
Hi @gjdanis, where did you see that `LOCK_TIMEOUT` was used?

gjdanis:
Hey @yitam I opened up SQL Server Profiler - you should see it executed before each query.

yitam:
hi @gjdanis 

Using your repro script, I could see that the query took roughly 40s to return. It seems that query timeout has no effect on PDO::exec(). 

That being said, if using a prepared statement, then query timeout works. For example,

```
$t0 = microtime(true);

$options = array(PDO::SQLSRV_ATTR_QUERY_TIMEOUT => 2);
$query = 'WAITFOR DELAY \'00:00:40\'; SELECT 1';

$stmt = $conn->prepare($query, $options);
echo $stmt->getAttribute( constant( "PDO::SQLSRV_ATTR_QUERY_TIMEOUT" ) ) . PHP_EOL;  
$result = $stmt->execute();

$t1 = microtime(true);

$elapsed = $t1 - $t0;

echo "Time elapsed: $elapsed secs with result:\n";
var_dump($result);

print "Done\n";
```

The output I got was

```
2
Time elapsed: 2.0018880367279 secs with result:
bool(false)
Done
```

Will investigate some more and get back to you on this.

gjdanis:
Hey @yitam any update on this?

I can confirm that statement options seem to work. Here's the trace I ran and it looks like we issue this setting prior to running the query.
[...]

PR #1037, opened 2019-09-12
Modified how drivers handle query timeout settings

yitam:
In PDO. the query timeout setting is inherited from the corresponding connection attribute, and the user may override the query timeout setting more than once using statement option or via PDOStatement::setAttribute().

### DIFF
--- a/test/functional/pdo_sqlsrv/pdo_1027_query_timeout.phpt
+++ b/test/functional/pdo_sqlsrv/pdo_1027_query_timeout.phpt
@@ -1,0 +1,198 @@
+--TEST--
+GitHub issue 1027 - PDO::SQLSRV_ATTR_QUERY_TIMEOUT had no effect on PDO::exec()
+--DESCRIPTION--
+This test verifies that setting PDO::SQLSRV_ATTR_QUERY_TIMEOUT correctly should affect PDO::exec() as in the case for PDO::prepare() (as statement attribute or option).
+--ENV--
+PHPT_EXEC=true
+--SKIPIF--
+<?php require('skipif_mid-refactor.inc'); ?>
+--FILE--
+<?php
+require_once("MsSetup.inc");
+require_once("MsCommon_mid-refactor.inc");
+
+const _DELAY = 5;
+
+$message = '*Invalid value timeout specified for option PDO::SQLSRV_ATTR_QUERY_TIMEOUT.';
+$delay = _DELAY;
+$query = "WAITFOR DELAY '00:00:$delay'; SELECT 1";
+$error = '*Query timeout expired';
+
+function testTimeoutAttribute($conn, $timeout, $statementLevel = false)
+{
+    global $message;
+
+    $error = str_replace('timeout', $timeout, $message);
+
+    try {
+        if ($statementLevel) {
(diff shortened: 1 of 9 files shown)


test/functional/pdo_sqlsrv/pdo_1027_query_timeout.phpt
```diff
@@ -0,0 +1,198 @@
...
+    $error = str_replace('timeout', $timeout, $message);
+
+    try {
+        if ($statementLevel) {
+            trace("statement option expects error: $error\n");
+            $options = array(PDO::SQLSRV_ATTR_QUERY_TIMEOUT => $timeout);
+            $sql = 'SELECT 1';
+            $stmt = $conn->prepare($sql, $options);
+        } else {
+            trace("connection attribute expects error: $error\n");
```
david-puglielli: I'm confused, do we expect a timeout error on 'SELECT 1'?

yitam: This method set invalid timeout values so the setting should have already failed

test/functional/pdo_sqlsrv/pdo_1027_query_timeout.phpt
```diff
@@ -0,0 +1,198 @@
...
+    testTimeoutAttribute($conn, 'xyz');
+    testTimeoutAttribute($conn, -99, true);
+    testTimeoutAttribute($conn, 'abc', true);
+}
+
+function checkTimeElapsed($message, $t0, $t1, $expectedDelay)
+{
+    $elapsed = $t1 - $t0;
+    $diff = abs($elapsed - $expectedDelay);
+    $leeway = 1.0;
```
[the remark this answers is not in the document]
yitam: this occasionally happened esp when test servers and agents are not the same instances

test/functional/pdo_sqlsrv/pdo_1027_query_timeout.phpt
```diff
@@ -0,0 +1,198 @@
...
+    $conn->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
+    
+    // if timeout is 0 it means no timeout 
+    $delay = ($timeout > 0) ? $timeout : _DELAY;
+
+    $result = null;
+    $t0 = microtime(true);
+
+    try {
+        $result = $conn->exec($query);
```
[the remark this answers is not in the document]
yitam: If timeout expired, an exception would have thrown. Does it make sense?

test/functional/pdo_sqlsrv/pdo_1027_query_timeout.phpt
david-puglielli: But it wouldn't return `Query timeout expired`, it should be `Invalid timeout value...` (so fnmatch should be comparing to $message not to $error)

yitam: oh... here in this function there's a local variable $error as well. My bad I should have used a diff name for this local variable

yitam:
See related issue #1027

PR #1045, opened 2019-10-19
Always Encrypted v2 support

david-puglielli:


test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+    }
+
+    $query = substr($query, 0, -6)."\n)";
+
+    return $query;
+}
+
+// The INSERT query for the table
+function formulateSetupQuery($tableName, &$dataTypes, &$colNames, &$colNamesAE)
+{
```
yitam: "SetupQuery" actually means the "InsertQuery" right? Again, use "construct" rather than "formulate" to be consistent with the above functions

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+    foreach ($dataTypes as $type) {
+        $colName1 = $colNames[$type].", ";
+        $colName2 = $colNamesAE[$type].", ";
+        $queryTypes .= $colName1;
+        $queryTypes .= $colName2;
+        $valuesString .= "?, ?, ";
+    }
+
+    $queryTypes = substr($queryTypes, 0, -2).")";
+    $valuesString = substr($valuesString, 0, -2).")";
```
yitam: "-2" here is to rid of the extra comma and space chars? Please add a comment here too

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+    for ($v = 0; $v < sizeof($testValues['bigint']); ++$v) {
+        $insertValues = array();
+
+        foreach ($dataTypes as $type) {
+            $insertValues[] = $testValues[$type][$v];
+            $insertValues[] = $testValues[$type][$v];
+        }
+
+        // Insert the data using sqlsrv_prepare()
+        try {
```
yitam: Update the comment to use the PDO prepare

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+            } elseif (isEnclaveEnabled($key)) {
+                $e = $error->errorInfo;
+                checkErrors($e, array('42000', '33546'));
+            } elseif (!isEnclaveEnabled($key)) {
+                $e = $error->errorInfo;
+                checkErrors($e, array('42000', '33277'));
+            } else {
+                print_r($error);
+                die("AE statement execution when it shouldn't!");
+            }
```
yitam: The message for die() is missing a word like "failed"
See another similar one below

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+    }
+
+    $AEres = $AEstmt->fetchAll(PDO::FETCH_NUM);
+    $nonAEres = $nonAEstmt->fetchAll(PDO::FETCH_NUM);
+    $AEcount = count($AEres);
+    $nonAEcount = count($nonAEres);
+
+    if ($type == 'char' or $type == 'nchar') {
+        // char and nchar may not return the same results - at this point
+        // we've verified that statement execution works so just return
```
yitam: Is the difference something to be fixed in the near future? 
If so, please put a TODO here 
If not, please justify

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+    } else {
+        if ($AEcount != 0) {
+            $i = 0;
+            foreach ($AEres as $AEr) {
+                if ($AEr[0] != $nonAEres[$i][0]) {
+                    print_r("AE and non-AE results are different! For field $i, got AE result ".$AEres[$i][0]." and non-AE result ".$nonAEres[$i][0]."\n");
+                }
+                ++$i;
+            }
+        }
```
yitam: For each of the cases above please elaborate the message for easier troubleshooting later -- so that we know this particular error belongs to which case, etc. For example, see if you can extend the message "Too many AE results blah blah blah"

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+    }
+}
+
+// testCompare selects based on a comparison in the WHERE clause and compares
+// the results between encrypted and non-encrypted columns, checking that the
+// results are identical
+function TestCompare($conn, $tableName, $comparisons, $dataTypes, $colNames, $thresholds, $length, $key, $encryptionType, $attestation)
+{
```
yitam: Please rename the function to be consistent with the one in the comment.
Also, this function has many arguments, so please briefly explain what each argument is for.
You can use the function getInsertSqlComplete() in sqlsrv/MsHelper.inc  as an example

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+    foreach ($comparisons as $comparison) {
+        foreach ($dataTypes as $type) {
+
+            // Unicode data types require the PHPTYPE to be specified to UTF-8 string
+            // and the correct collation. Otherwise we get different results between
+            // the encrypted and non-encrypted columns (probably because the collation
+            // was only changed in the encryption query).
+            $string = dataTypeIsStringMax($type);
```
yitam: The comment is seemingly specific for sqlsrv test?

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+            compareResults($AEstmt, $nonAEstmt, $key, $encryptionType, $attestation, $comparison);
+        }
+    }
+}
+
+// testPatternMatch selects based on a pattern in the WHERE clause and compares
+// the results between encrypted and non-encrypted columns, checking that the
+// results are identical
+function TestPatternMatch($conn, $tableName, $patterns, $dataTypes, $colNames, $length, $key, $encryptionType, $attestation)
+{
```
yitam: Again, please rename the function to be consistent with the one in the comment. Also, add brief descriptions for each argument. See my review comment above

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+    foreach ($dataTypes as $type) {
+
+        // Pattern matching doesn't work in AE for non-string types
+        // without an explicit cast
+        if (!dataTypeIsStringMax($type)) {
+            continue;
```
yitam: So does it mean we do not plan to test pattern matching for non-string types? Please indicate it in the comment

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+                                  $pattern."%",
+                                  "%".$pattern,
+                                  "%".$pattern."%",
+                                  );
+
+            foreach ($patternArray as $spattern) {
+
+                $unicode = dataTypeIsUnicode($type);
+                $collate = $unicode ? " COLLATE Latin1_General_BIN2" : "";
+                $PDOType = getPDOType($type);
```
yitam: This COLLATE clause is required for unicode comparisons only? Is it always the case or something specific for AE related comparisons?

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+    }
+}
+
+function checkErrors($errors, ...$codes)
+{
+    $codeFound = false;
+
+    foreach ($codes as $code) {
+        if ($code[0]==$errors[0] and $code[1]==$errors[1]) {
+            $codeFound = true;
```
yitam: Once the code is found, shouldn't we break the for loop?

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+        print_r($codes);
+        echo "\n";
+        die("Error code not found.\n");
+    }
+}
+
+function isEnclaveEnabled($key)
+{
+    if (strpos($key, '-enclave') !== false) return true;
+    else return false;
```
yitam: Should always use curly braces for any if else statements, but for this one and the several cases below, you can simply do something like `return (strpos($key, '-enclave') !== false)`, etc.

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+
+function getPDOType($type)
+{
+    switch($type) {
+        case "bigint":
+        case "integer":
+        case "smallint":
+        case "tinyint":
+            return PDO::PARAM_INT;
+            break;
```
yitam: After `return` no need to `break`, isn't it?

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
```diff
@@ -0,0 +1,532 @@
...
+        case "varbinary(max)":
+            return PDO::PARAM_LOB;
+            break;
+        default:
+            die("Case is missing for $type type in getPDOType.\n");
+    }
+}
+
+function getSQLType($type, $length)
+{
```
yitam: This function is unlikely needed

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
david-puglielli: It's a TODO.

test/functional/pdo_sqlsrv/pdo_AE_functions.inc
david-puglielli: For AE comparisons in Unicode, a Latin1_General_BIN2 collation is required. I don't think it's required for ASCII data.

test/functional/pdo_sqlsrv/pdo_aev2_encrypt_plaintext.phpt
```diff
@@ -0,0 +1,123 @@
...
+                    print_r($error);
+                    die("Creating table failed when it shouldn't have!\n");
+                }
+
+                insertValues($conn, $insertQuery, $dataTypes, $testValues);
+
+                if ($count == 0)
+                {
+                    // Split the data type array, because for some reason we get an error
+                    // if the query is too long (>2000 characters)
```
yitam: Put a TODO here to monitor this known issue
Please also move the opening curly bracket to the end of `if` clause

test/functional/pdo_sqlsrv/pdo_aev2_encrypt_plaintext.phpt
```diff
@@ -0,0 +1,123 @@
...
+                            } else {
+                                print_r($error);
+                                die("Encrypting failed when it shouldn't have!\n");
+                            }
+                        }
+                    }
+                }
+
+                if ($encryption_failed) continue;
+
```
yitam: Use camel style for variables and curly brackets for `if` please

test/functional/pdo_sqlsrv/pdo_aev2_keywords.phpt
```diff
@@ -0,0 +1,54 @@
...
+--DESCRIPTION--
+This test checks that connection fails when ColumnEncryption is set to nonsense,
+or when it is set to a bad protocol. Then it checks that connection succeeds when
+the attestation URL is bad.
+--FILE--
+<?php
+include("MsSetup.inc");
+include("AE_v2_values.inc");
+include("pdo_AE_functions.inc");
+
```
yitam: `require_once`  + skipif
re your comment above, can you find a better word for `bad` to describe the URL?

test/functional/sqlsrv/sqlsrv_AE_functions.inc
```diff
@@ -366,10 +367,15 @@ function testPatternMatch($conn, $tableName, $patterns, $dataTypes, $colNames, $
 
             foreach ($patternarray as $spattern) {
 
+                // Unicode operations with AE require the PHPTYPE to be specified to
+                // UTF-8 and the Latin1_General_BIN2 collation. If the COLLATE
+                // clause is left out, we get different results between the
+                // encrypted and non-encrypted columns (probably because the
+                // collation was only changed in the encryption query).
                 // We must pass the length of the pattern matching string
```
yitam: For cases like pattern matching in sqlsrv, what happens if the length of pattern string is not passed in sqlsrv_prepare()?

PR #1046, opened 2019-10-23
Added the include file to ansi tests

yitam:
These tests sometimes failed outside Windows due to the required include file missing.


david-puglielli:
Why do these tests only 'sometimes' fail without the required include? If the required file is not included, they should always fail, no?

yitam:
@david-puglielli don't know. It fails consistently in Catalina and in some Linux distros when directly run without using run-tests.php

PR #1049, opened 2019-11-01
Saved php types with metadata when fetching

yitam:


david-puglielli:
This looks good to me, but do we have tests that could ferret out issues with caching php types?

yitam:
@david-puglielli many existing fetching tests go through this path, but let me see if I can add more tests.

source/pdo_sqlsrv/pdo_stmt.cpp
```diff
@@ -782,14 +782,11 @@ int pdo_sqlsrv_stmt_get_col_data( _Inout_ pdo_stmt_t *stmt, _In_ int colno,
 
         // set the encoding if the user specified one via bindColumn, otherwise use the statement's encoding
         // save the php type for next use
-        if (driver_stmt->current_meta_data[colno]->sqlsrv_php_type.typeinfo.type != SQLSRV_PHPTYPE_INVALID) {
-            sqlsrv_php_type = driver_stmt->current_meta_data[colno]->sqlsrv_php_type;
-        }
```
david-puglielli: Did this cause problems with the new tests?

yitam: Yes, as shown in the pdo test, the user may choose different encoding or type for every single fetch

PR #1058, opened 2019-11-16
Updated all CIs

yitam:


appveyor.yml
```diff
@@ -23,7 +23,7 @@ environment:
         TEST_PHP_SQL_SERVER: (local)\SQL2017
         SQL_INSTANCE: SQL2017
         PHP_VC: 15
-        PHP_MAJOR_VER: 7.2
+        PHP_MAJOR_VER: 7.3
         PHP_MINOR_VER: 11
         PHP_EXE_PATH: x64\Release_TS
         THREAD: ts
```
david-puglielli: PHP_MAJOR_VER below is still 7.1?

azure-pipelines.yml
```diff
@@ -60,7 +60,7 @@ jobs:
```
david-puglielli: Below we install ODBC 17.3 - should this be updated?

appveyor.yml
yitam: Yes, for now, until after 7.4 is released. In that case we will change the matrix to drop the image VS 2015

azure-pipelines.yml
yitam: Did you mean Windows? Not for now because the job is only a placeholder for future development (see the condition is `false`)

PR #1062, opened 2019-11-27
Update linux mac instructions for php 7.4

yitam:


Linux-mac-install.md
```diff
@@ -1,50 +1,50 @@
...
 
-These instructions install PHP 7.3 by default. Note that some supported Linux distros default to PHP 7.0 or earlier, which is not supported for the PHP drivers for SQL Server -- please see the notes at the beginning of each section to install PHP 7.1 or 7.2 instead.
+These instructions install PHP 7.4 by default. Note that some supported Linux distros default to PHP 7.0 or earlier, which is not supported for the PHP drivers for SQL Server -- please see the notes at the beginning of each section to install PHP 7.2 or above.
```
david-puglielli: We should say `some supported distros default to PHP 7.1 or earlier, which is not supported for the latest version of the PHP drivers...`. In fact we should probably emphasize somewhere that these instructions don't apply to earlier versions of the drivers?

PR #1070, opened 2019-12-20
Fixed the skipif wordings and styles

yitam:
The money_format() function is deprecated in PHP 7.4, so use intl NumberFormatter

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -128,6 +128,7 @@ jobs:
 
   - script: |
       echo ready to build extensions
+      sudo apt-get install -y php$(phpVersion)-intl
       cd $(Build.SourcesDirectory)/source
       chmod a+x packagize.sh
       ./packagize.sh
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -59,6 +59,8 @@ jobs:
     displayName: 'Build and install drivers'
 
 - job: Linux
+  variables:
+    phpver: 7.3
   pool:
     vmImage: 'ubuntu-18.04'
   steps:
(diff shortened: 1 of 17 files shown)


azure-pipelines.yml
```diff
@@ -72,13 +74,13 @@ jobs:
       architecture: 'x64'
 
   - script: |
-      sudo update-alternatives --set php /usr/bin/php$(phpVersion)
-      sudo update-alternatives --set phar /usr/bin/phar$(phpVersion)
-      sudo update-alternatives --set phpdbg /usr/bin/phpdbg$(phpVersion)
-      sudo update-alternatives --set php-cgi /usr/bin/php-cgi$(phpVersion)
-      sudo update-alternatives --set phar.phar /usr/bin/phar.phar$(phpVersion)
+      sudo update-alternatives --set php /usr/bin/php$(phpver)
```
david-puglielli: I assume this is because PHP 7.4 is not yet available in the standard repos?

test/functional/pdo_sqlsrv/skipif.inc
```diff
@@ -1,4 +1,4 @@
 <?php
-if (!extension_loaded("pdo") || !extension_loaded('pdo_sqlsrv'))
-    die("PDO driver cannot be loaded; skipping test.\n");
-?>
+if (!extension_loaded("pdo_sqlsrv")) {
+    die("skip Extension not loaded");
+}
```
david-puglielli: why get rid of the closing `?>`?

yitam: Because the closing tag of a PHP block at the end of a file is actually optional

azure-pipelines.yml
yitam: yes, as of today :)

Issue #1079, opened 2020-01-17
Unknown type in sqlsrv_buffered_query::sqlsrv_buffered_query When querying extended properties


Beakerboy:
`EXEC sp_addextendedproperty @name=N'MS_Description', @value='Schema table description may contain "quotes" and could be long—very long indeed.',@level0type = N'Schema', @level0name = 'dbo',@level1type = N'Table', @level1name = 'test48992550test_table'`

`SELECT prop.value FROM information_schema.tables tbl
  LEFT JOIN sys.extended_properties prop
    ON prop.major_id = object_id(tbl.table_schema + '.' + tbl.table_name) 
    AND prop.minor_id = 0
    AND prop.name = 'MS_Description'
  WHERE tbl.table_type = 'base table' AND tbl.table_schema='dbo' AND tbl.table_name='test48992550test_table'`

Produces the error:
`Drupal\KernelTests\Core\Database\SchemaTest::testSchema PHPUnit\Framework\Exception: PHP Fatal error:  Unknown type in sqlsrv_buffered_query::sqlsrv_buffered_query in /home/travis/build/Beakerboy/drupal-project/web/drivers/lib/Drupal/Driver/Database/sqlsrv/Statement.php on line 67`

PHP Driver version or file name
sqlsrv-5.6.1
pdo_sqlsrv-5.6.1

SQL Server version
mssql-server-2019 xenial

Client operating system
Ubuntu

PHP version
7.2.15

yitam:
>/home/travis/build/Beakerboy/drupal-project/web/drivers/lib/Drupal/Driver/Database/sqlsrv/Statement.php on line 67

hi @Beakerboy please provide few lines before and after this line 67

Beakerboy:
Wow! Such fast response:

`public function execute($args = [], $options = []) {
    if (isset($options['fetch'])) {
      if (is_string($options['fetch'])) {
        // Default to an object. Note: db fields will be added to the object
        // before the constructor is run. If you need to assign fields after
        // the constructor is run, see http://drupal.org/node/315092.
        $this->setFetchMode(PDO::FETCH_CLASS, $options['fetch']);
      }
      else {
        $this->setFetchMode($options['fetch']);
      }
    }

    $logger = $this->dbh->getLogger();
    $query_start = microtime(TRUE);

    // If parameteres have already been binded
    // to the statement and we pass an empty array here
    // we will get a PDO Exception.
    if (empty($args)) {
      $args = NULL;
    }

[...]

PR #1080, opened 2020-01-20
Include sql_variant type for buffered queries

yitam:
Fixed the issue reported in #1079 and #51

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_install:
   - docker pull mcr.microsoft.com/mssql/server:2019-latest
  
 install:
-  - docker run -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=Password123' -p 1433:1433 --name=$TEST_PHP_SQL_SERVER -d mcr.microsoft.com/mssql/server:2019-CTP3.2-ubuntu
+  - docker run -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=Password123' -p 1433:1433 --name=$TEST_PHP_SQL_SERVER -d mcr.microsoft.com/mssql/server:2019-latest
   - docker build --build-arg PHPSQLDIR=$PHPSQLDIR -t msphpsql-dev -f Dockerfile-msphpsql .
 
 before_script:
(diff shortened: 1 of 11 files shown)


test/functional/sqlsrv/srv_1079_sql_variant_buffered_queries.phpt
```diff
@@ -0,0 +1,62 @@
...
+  ["Type"]=>
+  int(-150)
+  ["Size"]=>
+  int(10)
+  ["Precision"]=>
+  NULL
+  ["Scale"]=>
+  NULL
+  ["Nullable"]=>
+  int(1)
```
david-puglielli: So sqlsrv_field_metadata does not return the `sqlsrv:decl_type` that identifies a sql_variant type?

yitam: Yes, in the [sqlsrv_field_metadata documentation](https://docs.microsoft.com/sql/connect/php/sqlsrv-field-metadata?view=sql-server-ver15) the TYPE returns the numeric value of the sql type, and here ` ["Type"]=>  int(-150)`, which is `SQL_SS_VARIANT`

.travis.yml
```diff
@@ -20,10 +20,10 @@ env:
   - TEST_PHP_SQL_PWD=Password123
 
 before_install:
-  - docker pull mcr.microsoft.com/mssql/server:2019-GA-ubuntu-16.04
+  - docker pull mcr.microsoft.com/mssql/server:2019-latest
```
david-puglielli: This installs the ubuntu version of SQL Server?

yitam: Yes, please scroll down this [page for your reference ](https://hub.docker.com/_/microsoft-mssql-server)

Beakerboy:
Thanks for fixing it so quickly! What’s the typical timeline for rolling these fixes out to PECL? Which versions will it be applied to?

yitam:
@Beakerboy consider yourself lucky this time ;) as this should be merged to the next stable release shortly

PR #1082, opened 2020-01-22
Changed version 5.8.0 

yitam:
Updated versions and the year in license


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright(c) 2019 Microsoft Corporation
```
david-puglielli: Is this really the only place the date is found?

yitam: yes I suppose so, unless you think otherwise?

PR #1086, opened 2020-01-31
5.8.0 rtw docs

david-puglielli:


Linux-mac-install.md
```diff
@@ -248,18 +364,18 @@ brew install autoconf automake libtool
...
 brew install apache2
 ```
-To find the Apache configuration file for your Apache installation, run 
+To find the Apache configuration file, `httpd.conf`, for your Apache installation, run 
 ```
 apachectl -V | grep SERVER_CONFIG_FILE
 ``` 
-and substitute the path for `httpd.conf` in the following commands:
+The following commands append the required configuration to `httpd.conf`. Be sure to substitute the path returned by the preceding command in place of `/usr/local/etc/httpd/httpd.conf`:
 ```
```
yitam: Thought we should not subsitute the httpd.conf path?

Linux-mac-install.md
```diff
@@ -216,15 +290,57 @@ sudo systemctl restart apache2
...
 To test your installation, see [Testing your installation](#testing-your-installation) at the end of this document.
 
-## Installing the drivers on macOS Sierra, High Sierra, Mojave, and Catalina
+## Installing the drivers on Alpine 3.11
```
yitam: Please specify this is experimental

Linux-mac-install.md
```diff
@@ -1,23 +1,27 @@
...
+The following instructions assume a clean environment and show how to install PHP 7.x, the Microsoft ODBC driver, the Apache web server, and the Microsoft Drivers for PHP for SQL Server on Ubuntu 16.04, 18.04, and 19.10, RedHat 7 and 8, Debian 8, 9, and 10, Suse 12 and 15, and macOS 10.13, 10.14, and 10.15. These instructions advise installing the drivers using PECL, but you can also download the prebuilt binaries from the [Microsoft Drivers for PHP for SQL Server](https://github.com/Microsoft/msphpsql/releases) Github project page and install them following the instructions in [Loading the Microsoft Drivers for PHP for SQL Server](https://docs.microsoft.com/sql/connect/php/loading-the-php-sql-driver). For an explanation of extension loading and why we do not add the extensions to php.ini, see the section on [loading the drivers](https://docs.microsoft.com/sql/connect/php/loading-the-php-sql-driver#loading-the-driver-at-php-startup).
```
yitam: In the above paragraph please add Alpine (experimental)

Linux-mac-install.md
david-puglielli: It will likely change though if the user installs Apache via brew, and I don't think earlier versions of macOS (before Catalina) shipped with Apache, so this default path may not be correct for all users.

david-puglielli: Correction - previous versions did ship with apache. Will modify the apachectl path to get the brew provided executable.

PR #1092, opened 2020-02-07
Made some tests more robusts across platforms

yitam:


test/functional/sqlsrv/MsCommon.inc
```diff
@@ -84,6 +84,12 @@ function isDaasMode()
     return ($daasMode ? true : false);
 }
 
+function isLocaleDisabled()
+{
+    global $daasMode, $localeDisabled;
+    return ($daasMode || $localeDisabled);
+}
+
```
david-puglielli: We don't use isDaasMode anymore now right? We can remove it

test/functional/pdo_sqlsrv/pdo_test_non_LOB_types.phpt
```diff
@@ -60,19 +64,19 @@ unset($stmt);
...
-)
\ No newline at end of file
+\(
+    \[0\] => 9223372036854775807
+    \[1\] => 2147483647
+    \[2\] => 32767
+    \[3\] => 255
+    \[4\] => 1
+    \[5\] => 9999999999999999999999999999999999999
+    \[6\] => 922337203685477.5807
```
david-puglielli: I think the backslashes are needed here too (and in a few other places) to match a '.'

test/functional/sqlsrv/srv_007_login_timeout.phpt
```diff
@@ -38,7 +41,7 @@ do {
             echo "Connection failed at $elapsed secs. Leeway is $leeway sec but the difference is $diff\n";
         } else {
             // The test will fail but this helps us decide if this test should be redesigned
-            echo "$numAttempts\t";
+            echo "$numAttempts: $diff\n";
```
david-puglielli: I think we could label the output to make it immediately obvious what happens if the test fails in a full test run, otherwise we just get two numbers. Something like `echo "Attempts: $numAttempts, Time difference: $diff\n"`.

test/functional/pdo_sqlsrv/pdostatement_fetchAll.phpt
```diff
@@ -180,9 +180,9 @@ array(2) {
     [5]=>
     string(10) "STRINGCOL2"
     ["FloatCol"]=>
-    string(18) "222.22200000000001"
+    string(%d) "222.222%S"
     [6]=>
-    string(18) "222.22200000000001"
+    string(%d) "222.222%S"
```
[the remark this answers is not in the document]
yitam: yes

test/functional/pdo_sqlsrv/pdostatement_fetchAll.phpt
```diff
@@ -425,7 +425,7 @@ string(10) "STRINGCOL2"
 string(23) "2000-11-11 11:11:11.223"
 string(10) "STRINGCOL2"
 string(10) "STRINGCOL2"
-string(18) "222.22200000000001"
+string(%d) "222.222%S"
```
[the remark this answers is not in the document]
yitam: So far, yes

test/functional/sqlsrv/MsCommon.inc
yitam: We still use it in specific tests for Azure

PR #1094, opened 2020-02-12
Xplat alpine changes

david-puglielli:


david-puglielli:
`uname -n` gives the hostname, which could be anything. But I'll add a check of os-release.

source/pdo_sqlsrv/config.m4
```diff
@@ -84,6 +84,12 @@ if test "$PHP_PDO_SQLSRV" != "no"; then
       PDO_SQLSRV_SHARED_LIBADD="$PDO_SQLSRV_SHARED_LIBADD -Wl,-z,now"
   fi
 
+  IS_ALPINE_1=`uname -a | cut -f 4 -d ' ' | cut -f 2 -d '-'`
+  IS_ALPINE_2=`cat /etc/os-release | grep ID | grep alpine | cut -f 2 -d '='`
+  if test "${IS_ALPINE_1}" = "Alpine" || test "${IS_ALPINE_2}" = "alpine"; then
+      AC_DEFINE(__MUSL__, 1, [ ])
+  fi
```
yitam: Forgot to mention that perhaps you can first check `uname` and if it's Darwin simply skips?

PR #1103, opened 2020-02-25
Revised a few existing tests that are flawed

yitam:
While converting tests from PR #720 discovered that the designs were flawed. These tests are rewritten and expanded to include non-AE testing as well.

### DIFF
--- a/test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_datetime.phpt
+++ b/test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_datetime.phpt
@@ -39,7 +39,7 @@ function testOutputParam($conn, $spname, $direction, $dataType, $sqlType)
 {
     // The driver does not support these types as output params, simply return
     if (isDateTimeType($sqlType) || isLOBType($sqlType)) {
-        return;
+        return true;
     }
     
     $sqlTypeConstant = get_sqlType_constant($sqlType);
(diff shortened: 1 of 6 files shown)


test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_datetime.phpt
```diff
@@ -1,111 +1,161 @@
...
+    $tbname = GetTempTableName("", false);
+    $colMetaArr = array(new AE\ColumnMeta($dataType, "c_det"), new AE\ColumnMeta($dataType, "c_rand", null, false));
+    AE\createTable($conn, $tbname, $colMetaArr);
+
+    // Create a Store Procedure
+    $spname = 'selectAllColumns';
+    createProc($conn, $spname, "@c_det $dataType OUTPUT, @c_rand $dataType OUTPUT", "SELECT @c_det = c_det, @c_rand = c_rand FROM $tbname");
+
+    // insert a row
+    $inputValues = array_slice(${explode("(", $dataType)[0] . "_params"}, 1, 2);
```
david-puglielli: Could we have an example here of what the input looks like? Expressions like this are tricky to parse by looking.

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_datetime.phpt
```diff
@@ -1,111 +1,161 @@
...
+    $r;
+    $stmt = AE\insertRow($conn, $tbname, array( $colMetaArr[0]->colName => $inputValues[0], $colMetaArr[1]->colName => $inputValues[1] ), $r);
+    if ($r === false) {
+        is_incompatible_types_error($dataType, "default type");
```
david-puglielli: is_incompatible_types_error has the same issue you fixed elsewhere - it compares SQLSTATE to a non-string 22018.

test/functional/sqlsrv/sqlsrv_ae_output_param_sqltype_numeric.phpt
```diff
@@ -107,17 +96,19 @@ function testOutputParam($conn, $spname, $direction, $dataType, $sqlType, $input
...
-            // an error when converting from one type to another and the
-            // latter indicating an incompatible conversion error
-            // Either one is acceptable for the purpose of this test
-            $success = ($errors[0]['SQLSTATE'] === '22018' || $errors[0]['SQLSTATE'] === '42000');
+            // we still get errors from the server -- if so, it might
+            // return either SQLSTATE '42000' or '22018' (operand type 
+            // clash but only happens with some certain types)
+            // E.g. when converting a bigint to int or an int to numeric, 
+            // SQLSTATE '42000' is returned, indicating an error when 
+            // converting from one type to another.
```
david-puglielli: I'm still not clear on why we would get errors if the types are supposed to be compatible. Do the errors occur only in AE?

yitam: It only happens with non-AE, and there are many possible combinations -- will investigate further in another task

PR #1106, opened 2020-03-16
Modified some BVT tests to be configurable

yitam:


yitam:
> Why server2? Otherwise looks good to me.

Because some tests require connecting to two servers

PR #1107, opened 2020-03-24
Set logger for driver API

yitam:


test/functional/sqlsrv/sqlsrv_connect_log_to_file.phpt
```diff
@@ -0,0 +1,56 @@
+--TEST--
+Tset functions return FALSE for errors with logging
```
v-mabarw:
```suggestion
Test functions return FALSE for errors with logging
```

test/functional/sqlsrv/sqlsrv_connect_logs.phpt
```diff
@@ -0,0 +1,49 @@
+--TEST--
+Tset functions return FALSE for errors with logging
```
v-mabarw:
```suggestion
Test functions return FALSE for errors with logging
```

source/sqlsrv/conn.cpp
```diff
@@ -845,13 +844,15 @@ PHP_FUNCTION( sqlsrv_close )
...
+    // Without sqlsrv_close(), this function is invoked by php during the final clean up stage. 
+    // To prevent memory/resource leaks, no more logging at this point.
+    //LOG_FUNCTION( "sqlsrv_conn_dtor" );
 
     // get the structure
     ss_sqlsrv_conn *conn = static_cast<ss_sqlsrv_conn*>( rsrc->ptr );
     SQLSRV_ASSERT( conn != NULL, "sqlsrv_conn_dtor: connection was null");
 
-    SET_FUNCTION_NAME( *conn );
+    conn->set_func("sqlsrv_conn_dtor");
```
v-mabarw: Would \_FN\_ work here as well?

yitam: Not this one because the macro LOG_FUNCTION is commented but I can certainly use `__func__`
Thanks :)

PR #1127, opened 2020-04-16
cleanup TSRMLS usage

remicollet:
This macro are only used in PHP 5
Still present in PHP 7 (for BC)
Dropped in PHP 8

So seems useful to cleanup to prepare to PHP 8


remicollet:
Notice: build with PHP 8.0.0-dev is ok, but not really the goal of this PR, and not tested

yitam:
Resolved conflicts and merged. See #1130

PR #1275, opened 2021-07-01
Fix warning compiling core_stmt.cpp

mlocati:
Let's fix this compilation warning:

```
source/sqlsrv/shared/core_stmt.cpp:3104:25: warning: converting to non-pointer type 'long int' from NULL [-Wconversion-null]
         buffer_length = NULL;
                         ^~~~
```


yitam:
Thanks @mlocati for the contribution.

mlocati:
That was an easy one 😉

Issue #1310, opened 2021-10-03
work-around requested for Microsoft SQL Server MSODBCSQL handling of NULL


gjcarrette:
## PHP Driver version or file name
php-sqlsrv-5.8.0
## SQL Server version
Microsoft SQL Server 2017 (RTM-CU21) (KB4557397) - 14.0.3335.7
Microsoft SQL Server 2016 (SP2-CU14) (KB4564903) - 13.0.5830.85
## Client operating system
CentOS Linux release 7.6.1810 (Core)
Linux 3.10.0-957.12.1.el7.x86_64 #1 SMP Mon Apr 29 14:59:59 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
## PHP version
PHP 7.4.3
## Microsoft ODBC Driver version
Name        : msodbcsql17
Version     : 17.7.1.1
Build Date  : Fri 15 Jan 2021 02:56:50 PM EST
## Table schema
None required to run the test.
## Problem description
The Microsoft SQL Server ISNULL function seems to badly handle a NULL value transmitted to it
using SQLBindParameter with a data type of SQL_CHAR and length of 1. Because the only
purpose of PDO_SQLSRV is to interact with the Microsoft flagship relational database product
I am asking for a work-around to be implemented so that a NULL is transmitted
using SQLBindParameter with data type of SQL_VARCHAR and length of 0.  This work
around has been proven with the SQLRelay open source file [src/connections/odbc.cpp](https://github.com/davidwed/sqlrelay/blob/master/src/connections/odbc.cpp#L2407) starting at
about line 2407 where there is extensive commentary about this issue.

This work around would speed up the full retirement of the PDO_DBLIB freetds driver at the company where I work. In order to continue using Microsoft SQL Server I can supply a patch to PDO_SQLSRV if such a patch would be accepted in principal rather than rejected as not needed due to the possibility of fixing Microsoft SQL Server instead.

## Expected behavior and actual behavior
`select cast(ISNULL(:K, -1) as int) as K`
is expected to return -1 when K is NULL. Instead it gives
`[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Conversion failed when converting the varchar value ='*' to data type int.`

## Repro code or steps to reproduce
Run the attached php file with environment variables set to the database server, user credentials and database used for testing. Also attached is a stdout and an odbc trace.
[...]

PR #1311, opened 2021-10-04
adjust sql_data_type and column_size for NULL parameters

gjcarrette:
This is a proposed fix for https://github.com/microsoft/msphpsql/issues/1310

### DIFF
--- a/source/shared/core_stmt.cpp
+++ b/source/shared/core_stmt.cpp
@@ -2150,7 +2150,8 @@ void sqlsrv_param::process_null_param(_Inout_ zval* param_z)
         // Use the encoding to guess whether the sql_type is binary type or char type. For NULL cases,
         // if the server type is a binary type, then the server expects the sql_type to be binary type
         // as well, otherwise an error stating "Implicit conversion not allowed.." is thrown by the
-        // server. For all other server types, setting the sql_type to sql_char works fine.
+        // server. For all other server types, setting the sql_type to sql_varchar works fine.
+        // It must be varchar with column size 0 for ISNULL to work properly.
         sql_data_type = (encoding == SQLSRV_ENCODING_BINARY) ? SQL_BINARY : SQL_VARCHAR;
     }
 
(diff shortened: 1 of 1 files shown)


yitam:
@gjcarrette 

Please make changes to the method sqlsrv_param::process_null_param() instead (line 2146), because that is for input parameters:

```
--- a/source/shared/core_stmt.cpp
+++ b/source/shared/core_stmt.cpp
@@ -2151,13 +2151,13 @@ void sqlsrv_param::process_null_param(_Inout_ zval* param_z)
         // if the server type is a binary type, than the server expects the sql_type to be binary type
         // as well, otherwise an error stating "Implicit conversion not allowed.." is thrown by the
         // server. For all other server types, setting the sql_type to sql_char works fine.
-        sql_data_type = (encoding == SQLSRV_ENCODING_BINARY) ? SQL_BINARY : SQL_CHAR;
+        sql_data_type = (encoding == SQLSRV_ENCODING_BINARY) ? SQL_BINARY : SQL_VARCHAR;
     }

     c_data_type = (encoding == SQLSRV_ENCODING_BINARY) ? SQL_C_BINARY : SQL_C_CHAR;

     if (column_size == SQLSRV_UNKNOWN_SIZE) {
-        column_size = 1;
+        column_size = (encoding == SQLSRV_ENCODING_BINARY) ? 1 : 0;
         decimal_digits = 0;
     }
     buffer = NULL;

```

gjcarrette:
Yes, that makes sense, because the SQLRelay source file [odbc.cpp](https://github.com/davidwed/sqlrelay/blob/master/src/connections/odbc.cpp#L2472) that I mentioned in the issue also  uses SQL_C_BINARY for both binary or non-binary T-SQL NULL. 

So one thing we didn't do here is make a phpt file, a test case, that fails with the old code but passes with the new code.
To tell the truth I did not use the make test targets in the msphpsql tree, I used the test file that I distributed with the issue.

PR #1330, opened 2021-11-17
Fix restoring PDO::ATTR_ERRMODE after PDO::lastInsertId() call failed

### LINKED ISSUE #1331: PDO::ATTR_ERRMODE is implicitly changed to PDO::ERRMODE_SILENT after PDO::lastInsertId() call fails
## PHP version

PHP 8.1.0RC6

Note that GitHub Actions PHP image looks to use...

- `pdo_sqlsrv` on PHP 8.1 and later
- `pdo_dblib` on PHP 8.0 and before

## Driver and SQL Server versions

`PDO` dump Information provided Laravel's `dump()` function:

```php
PDO {#335
  inTransaction: false
  attributes: {
    CASE: NATURAL
    ERRMODE: SILENT
    PERSISTENT: false
    DRIVER_NAME: "sqlsrv"
    SERVER_INFO: array:3 [
      "CurrentDatabase" => "testing"
      "SQLServerVersion" => "15.00.4178"
      "SQLServerName" => "e6cf78d63667"
    ]
    ORACLE_NULLS: NATURAL
    CLIENT_VERSION: array:4 [
      "DriverName" => "libmsodbcsql-17.8.so.1.1"
      "DriverODBCVer" => "03.52"
      "DriverVer" => "17.08.0001"
      "ExtensionVer" => "5.10.0-beta1"
    ]
    SERVER_VERSION: "15.00.4178"
    STATEMENT_CLASS: array:1 [
      0 => "PDOStatement"
    ]
    EMULATE_PREPARES: false
    STRINGIFY_FETCHES: false
    DEFAULT_FETCH_MODE: BOTH
  }
}
```

## Problem description

Described here: [[PoC] PDO::ATTR_ERRMODE is implicitly changed to PDO::ERRMODE_SILENT after PDO::lastInsertId() call fails by mpyw · Pull Request #1 · mpyw/laravel-retry-on-duplicate-key](https://github.com/mpyw/laravel-retry-on-duplicate-key/pull/1)

> ## Overview
> 
> SQL Server handles auto-incrementing values differently than other RDBMS. I accidentally forgot to change the value of the `$incrementing` property of Model that is not auto-incremented. In this case, `PDO::lastInsertId()` is unexpectedly called and its call will fail. This is the intended design.
> 
> However, I found that after `PDO::lastInsertId()` failed, **`PDO::ATTR_ERRMODE` was implicitly changed to `PDO::ERRMODE_EXCEPTION` and never restored**. I'm convinced that this is a bug, so I submit a pull request to the `pdo_sqlsrv` repository.

## Expected behavior and actual behavior

After `PDO::lastInsertId()` fails,

### Expected

`PDO::ERRMODE_SILENT` is replaced by `PDO::ERRMODE_EXCEPTION` to restore original value.

### Actual

`PDO::ERRMODE_SILENT` is never replaced.

## Repro code or steps to reproduce

https://github.com/mpyw/laravel-retry-on-duplicate-key/blob/8fe6bc3ec72c37886e85b9e0817d6cedbe089db3/tests/Test.php

mpyw:
- fixes #1331

### DIFF
--- a/source/pdo_sqlsrv/pdo_dbh.cpp
+++ b/source/pdo_sqlsrv/pdo_dbh.cpp
@@ -1598,6 +1598,9 @@ zend_string * pdo_sqlsrv_dbh_last_id(_Inout_ pdo_dbh_t *dbh, _In_ const zend_str
 
         driver_stmt->~sqlsrv_stmt();
     } catch( core::CoreException& ) {
+        // restore error handling to its previous mode
+        dbh->error_mode = prev_err_mode;
+
         // copy any errors on the statement to the connection so that the user sees them, since the statement is released
         // before this method returns
         strcpy_s( dbh->error_code, sizeof( dbh->error_code ),
--- a/test/functional/pdo_sqlsrv/pdo_lastInsertId_restore_original_errmode_after_failed.phpt
+++ b/test/functional/pdo_sqlsrv/pdo_lastInsertId_restore_original_errmode_after_failed.phpt
@@ -32,15 +32,21 @@ try {
     dropTable($conn, "table1");
     dropTable($conn, "table2");
     dropTable($conn, "table3");
+
+    // Should trigger exception
+    $tsql = "SELECT * FROM dummy";
+    $conn->exec($tsql);
+
     unset($conn);
 } catch (PDOException $e) {
-    var_dump($e);
+    print_r($e->getMessage());
     exit;
 }
 
 
 ?>
---EXPECT--
-int(2)
-int(2)
-int(2)
+--EXPECTREGEX--
+int\(2\)
+int\(2\)
+int\(2\)
+.*Invalid object name \'dummy\'\.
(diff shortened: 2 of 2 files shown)


mpyw:
@yitam Hi! Where should I add new tests?

yitam:
Thanks @mpyw but we only accept PRs to `dev` branch. For new tests please add them to `test\functional\pdo_sqlsrv`

yitam:
@mpyw your change looks good. Thanks for catching this.

For your test, it's similar to our existing test "pdo_lastInsertId.phpt", but your calls to `lastInsertId()` would not end up in the exception would they?

To prove that your change has any effect, please put something like the following after one of the calls to `lastInsertId()`, to force an exception to occur:

```
    $tsql = "SELECT * FROM dummy";
    $conn->exec($tsql);
```

Also, change the catch block to be something like this:

```
} catch (PDOException $e) {
    var_dump($e->getMessage());
    exit;
}
```

mpyw:
@yitam Thanks! How do I confirm the actual message? 

https://github.com/microsoft/msphpsql/runs/4245732315

yitam:
@mpyw you don't have to match the exact message, but this is my suggestion (please drop the test tables first):

```
    dropTable($conn, "table3");

    $tsql = "SELECT * FROM dummy";
    $conn->exec($tsql);

    unset($conn);
} catch (PDOException $e) {
    print_r($e->getMessage());
    exit;
}


?>
--EXPECTREGEX--
string\(3\) "200"
string\(3\) "102"
(string\(0\) ""|bool\(false\))
.*Invalid object name \'dummy\'\.
```

PR #1404, opened 2022-07-28
Create pdo_1391_string_truncation_after_short_string.phpt

talkinnl:
Regresssion tests: Use the same Statement to insert a row with short strings first, and longer strings second.

- Issue also happens with VARCHAR( >=4002 ). Not just Max or NVARCHAR
- Issues 1391, 1393 (and possibly 1371 ?) might all have the same root cause.


absci:
Your changes have been merged into `dev` (#1408), so I'll close this.

PR #1468, opened 2023-08-03
Fixed error when using PDO::ATTR_STRINGIFY_FETCHES

SakiTakamachi:
This fixes the issue reported here.

https://github.com/laravel/framework/issues/47937

Fixes for errors caused by PDO changes in php8.1.22 and 8.2.9.

test code
```
<?php

// my env
$pdo = new PDO('sqlsrv:server=mssql-server;database=test;', 'test_user', 'p@ssw0rd');
$pdo->setAttribute(PDO::ATTR_STRINGIFY_FETCHES, false);
```

### DIFF
--- a/source/pdo_sqlsrv/pdo_dbh.cpp
+++ b/source/pdo_sqlsrv/pdo_dbh.cpp
@@ -1323,14 +1323,6 @@ bool pdo_sqlsrv_dbh_set_attr(_Inout_ pdo_dbh_t *dbh, _In_ zend_long attr, _Inout
             break;
 #endif
 
-#if (PHP_VERSION_ID >= 80122 && PHP_VERSION_ID < 80200) || PHP_VERSION_ID >= 80209
-            case PDO_ATTR_STRINGIFY_FETCHES:
-            {
-                // do nothing
-            }
-            break;
-#endif
-
             // Not supported
             case PDO_ATTR_FETCH_TABLE_NAMES:
             case PDO_ATTR_FETCH_CATALOG_NAMES:
--- a/source/pdo_sqlsrv/pdo_dbh.cpp
+++ b/source/pdo_sqlsrv/pdo_dbh.cpp
@@ -1323,6 +1323,14 @@ bool pdo_sqlsrv_dbh_set_attr(_Inout_ pdo_dbh_t *dbh, _In_ zend_long attr, _Inout
             break;
 #endif
 
+#if (PHP_VERSION_ID >= 80122 && PHP_VERSION_ID < 80200) || PHP_VERSION_ID >= 80209
+            case PDO_ATTR_STRINGIFY_FETCHES:
+            {
+                // do nothing
+            }
+            break;
+#endif
+
             // Not supported
             case PDO_ATTR_FETCH_TABLE_NAMES:
             case PDO_ATTR_FETCH_CATALOG_NAMES:


GrahamCampbell:
Why does this PR change nothing?

SakiTakamachi:
A test was crashing unexpectedly, so I'm temporarily running it without any changes.
Revert when the current test is finished.

SakiTakamachi:
The tests seem to fail even without any changes, and there seems to be some unstable tests.

SakiTakamachi:
@v-makouz 
If you don't mind, could you please review it?

SakiTakamachi:
This test will fail even if you didn't make any changes.

I don't understand why this is the case, just by looking at it lightly.

absci:
Thanks for your contribution, there're some tests may fail intermittently due to connection. I'll rerun it.

SakiTakamachi:
@absci 

Thank you for confirmation.

Am I correct in thinking that I have to rerun the connection test many times until it succeeds?

youkidearitai:
@absci 

Hi, I heard PHP 8.1.22 and 8.2.8 throws Exception when `PDO::ATTR_STRINGIFY_FETCHES`.
I would appreciate it if you could pass this pull request.

Regards.

absci (approved):

SakiTakamachi:
Thank you for confirming and merging!

jarkko-hautakorpi:
Are there perhaps some undocumented changes coming, or was it a bug and no changes intended?
https://learn.microsoft.com/en-us/sql/connect/php/pdo-setattribute?view=sql-server-ver16

SakiTakamachi:
> This documentation is still valid?

Yes, it is.

This is a fix that removes the error caused by changes in PDO core.

AbrahamBrookes:
I also got hit with this issue, thanks for the PR and the merge!

pirex360:
Hi, I have also :  `SQLSTATE[IMSSP]: An invalid attribute was designated on the PDO object` error.
- In laravel v9 and PHP 8.1.22 with MSODBC 18, the plesk has auto-updated and the driver for sqlsrv doesn't work anymore.

I tried everything, removing all PDO::ATTR or PDO::SQLSRV, and the result is the same! Here is a sample of the connection data:

`array(
                'driver'        => 'sqlsrv',
                'host'          => 'xxx.xxx.xxx.xxx',
                'port'          => 999999,
                'database'      => 'testdb',
                'username'      => 'johndoe',
                'password'      => 'test',
                'charset'       => 'utf8',
                'prefix'        => '',
                'prefix_indexes' => true,
                'options'       => array(
                    PDO::ATTR_EMULATE_PREPARES => true,
                    PDO::ATTR_STRINGIFY_FETCHES => false,
                    PDO::ATTR_CASE => PDO::CASE_LOWER,
                    PDO::SQLSRV_ATTR_FETCHES_NUMERIC_TYPE => true,
                    PDO::SQLSRV_ATTR_FORMAT_DECIMALS => true,
                ),
                'encrypt' => 'yes',     // added for MSSQL 18 to work
                'trust_server_certificate' => true, // added for MSQL 18 to work
            )`

I also tried upgrading to PHP 8.2.8 and the result is the same.
I have pdo_sqlsrv and sqldrv extensions on 5.10.1


Can anybody help me?

SakiTakamachi:
@pirex360 

> I also tried upgrading to PHP 8.2.8 and the result is the same.

Is it really php8.2.8?
This issue occurs on 8.2.9 and later.

SakiTakamachi:
@pirex360 

This fix is ​​still in the process of being merged into the dev branch and has not been released.

It will be included in the next release.

If you update to php8.2.9 again before the release, of course the problem will reoccur.

Using the driver that is new version after release will permanently fix this issue.

ssswang:
@pirex360 
If you upgrade to php 8.2.9, you will get `SQLSTATE[IMSSP]: An invalid attribute was designated on the PDO object` because of `PDO::ATTR_STRINGIFY_FETCHES => false,`

In my own limited experiment and testing, it is safe to remove this line. Please correct me if I am wrong.

SakiTakamachi:
@ssswang 

Yes, you are right.

In the current php specification, the initial value of `PDO::ATTR_STRINGIFY_FETCHES` is false.

Therefore, it can be controlled by specifying `SQLSRV_ATTR_FETCHES_NUMERIC_TYPE`.

You can safely delete the PDO::ATTR_STRINGIFY_FETCHES line.

neclimdul:
"after release" :disappointed:

ssswang:
@SakiTakamachi 
Arigato, Sakichi. Thank you for confirming.
It's 5 AM in Japan. Don't work too hard. ^^

pirex360:
> @pirex360 If you upgrade to php 8.2.9, you will get `SQLSTATE[IMSSP]: An invalid attribute was designated on the PDO object` because of `PDO::ATTR_STRINGIFY_FETCHES => false,`
> 
> In my own limited experiment and testing, it is safe to remove this line. Please correct me if I am wrong.
> 
> This is also a solution wrote by SakiTakamachi using a try catch block without removing `PDO::ATTR_STRINGIFY_FETCHES => false,` [laravel/framework#47937 (comment)](https://github.com/laravel/framework/issues/47937#issuecomment-1671772276)

It's weird, In my testing... I remove all the options array from the connection :
```
 'options'       => array(
                    PDO::ATTR_EMULATE_PREPARES => true,
                    PDO::ATTR_STRINGIFY_FETCHES => false,
                    PDO::ATTR_CASE => PDO::CASE_LOWER,
                    PDO::SQLSRV_ATTR_FETCHES_NUMERIC_TYPE => true,
                    PDO::SQLSRV_ATTR_FORMAT_DECIMALS => true,
                ),
```

And still, the error appears...

SakiTakamachi:
It doesn't make much sense to test it in Laravel. Because the initial value is set inside Laravel, it is not possible to completely remove the option specification.

https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Connectors/SqlServerConnector.php#L19

SakiTakamachi:
And removing an option doesn't mean setting it to false, it means not specifying anything.

pirex360:
> And removing an option doesn't mean setting it to false, it means not specifying anything.

Sure, You are right... it fallback to the defaults... but doesn't the default supported by the driver?
The key is your "Because the initial value is set inside Laravel, it is not possible to completely remove the option specification.", it points to the problem of Laravel defaults and not Driver Default.

Thanks for your time, it was enlightenment.

jarkko-hautakorpi:
@pirex360 It's not a Laravel problem, because it's a bug in the driver, which is missing the parameter, that it's suppose to have. But, if you like to change Laravel, I belive it's the file vendor/laravel/framework/src/Illuminate/Database/Connectors/SqlServerConnector.php on line 19
`PDO::ATTR_STRINGIFY_FETCHES => false,`
https://github.com/jimrubenstein/laravel-framework/blob/f7ba54c132319631163ac0c98101da0a3b16c3ca/src/Illuminate/Database/Connectors/SqlServerConnector.php#L19
In some cases this is a very useful and important parameter to have.

JGlueck-WIKA:
@absci Any news on when the next release will take place?

SakiTakamachi:
@absci 

Is it difficult to release a patch version that addresses this issue as 5.11.1 first?

absci:
> @absci
> 
> Is it difficult to release a patch version that addresses this issue as 5.11.1 first?

I'll discuss it with management levels. Maybe we can do a hotfix release.

SakiTakamachi:
>I'll discuss it with management levels. Maybe we can do a hotfix release.

I apologize for the trouble.
Thanks so much.

skylord123:
> I'll discuss it with management levels. Maybe we can do a hotfix release.

This is currently breaking the build process for my company. Since we build our container off Ubuntu using Ondrej repo we are stuck on PHP 8.1.22 that has this problem. So a hotfix would be a godsend to getting this resolved.

Also thanks @SakiTakamachi for a good job finding and fixing this.

kokoshneta:
> > @absci
> > Is it difficult to release a patch version that addresses this issue as 5.11.1 first?
> 
> I'll discuss it with management levels. Maybe we can do a hotfix release.

Since this is currently breaking every single Laravel install where the developer is unable to control the minor version of PHP (e.g., Plesk-hosted – once you’re on 8.2.9, it seems you can’t downgrade back to 8.2.8 without Plesk auto-re-upgrading), a hotfix would be highly appreciated.

JGlueck-WIKA:
For those of you with a Laravel installation also suffering from this problem at the moment - the [workaround suggested here](https://github.com/laravel/framework/issues/47937#issuecomment-1678200201) helped us to circumnavigate that issue for now.

skylord123:
> Here is the easiest und final solution: https://github.com/SakiTakamachi/laravel-sqlsrv-err-avoid
> 
> install it and the problem is gone.

Not a final solution, just a bandaid. Basically just ignores the problem this driver is currently causing. The final solution would be to get a patched version of msphpsql.

I have over 25+ services so manually installing this bandaid until this is fixed then removing it is quite a large effort.

David-Engel:
Think this PR should not have been merged as-is. The fix should be made in PHP (the offending PR should be reverted), where the break happened. This change will be fine if it is needed for PHP 8.3. But don't make everyone else patch out-of-band for an upstream error.

Changes that break backwards compatibility should only be made on version changes where you might expect to have to test for version compatibility. Upgrading from PHP 8.2.8 to PHP 8.2.9 shouldn't break downstream apps.

SakiTakamachi:
@David-Engel 

All other pdo exts do not raise an error under the same conditions.

Only PDO SQLSRV raises an error on its own.

So I personally don't think this is a BC break.

It takes a lot of effort to consider all downstream proprietary specifications in order to make upstream modifications.

SakiTakamachi:
Even if you look at the PDO documentation, it doesn't say that setAttribute throw an exception, and it states that it simply returns false on failure.

https://www.php.net/manual/en/pdo.setattribute.php#refsect1-pdo.setattribute-returnvalues

SakiTakamachi:
@David-Engel 
This is a problem that this repository should solve, as the proprietary error throwing spec is causing this issue.
I really hope for an early hotfix release.
If you want a fix at the underlying specification level rather than a monkey patch, I'm ready to help with that fix.

SakiTakamachi:
I have created an issue because I can't keep talking forever in this PR.
https://github.com/microsoft/msphpsql/issues/1474

Issue #1491, opened 2023-12-04
HY104 Invalid precision value when reusing prepared statement


Leiothrix1:
Versions:
PHP: 8.2.13
sqlsrv: 5.11.1
ODBC driver: 18.3.2.1
SQL Server: 14.0.3038.14
Server OS: Windows Server 2016

Test table:
CREATE TABLE dbo.TestTable (
	TestColumn VARCHAR (8000) NULL
)

Test code: attached
[TestCase.txt](https://github.com/microsoft/msphpsql/files/13550867/TestCase.txt)

Expected output: two identical rows inserted into the database.

Actual output:
The first row is inserted in the database, and the second fails with error:
HY104 [Microsoft][ODBC Driver 18 for SQL Server]Invalid precision value

This code works fine if SQLSRV_SQLTYPE_VARCHAR() is set <= 4000 or to MAX.  If it is between 4001->8000 it produces the error above on the second insert.  This also works if the SQL type is not specified in the prepared statement.  Note that there is nothing changing between the inserts, it is just sqlsrv_execute being executed twice.

v-makouz:
I can't seem to reproduce this one. Can you provide an ODBC trace?

Leiothrix1:
Sorry I didn't know that option even existed or I would have included that originally.

Trace is attached.

[trace.txt](https://github.com/microsoft/msphpsql/files/13574089/trace.txt)

Thanks!

v-makouz:
Strangely, in my tests, I could reproduce this error with 5.10 version of the PHP driver, but not 5.11. But you're already using 5.11, so I will investigate this further

v-makouz:
I was able to reproduce this, we are now looking into it.

Issue #1448, opened 2023-03-23
[Bug] [Microsoft][ODBC Driver 17 for SQL Server]Invalid character value for cast specification


thsmrtone1:
**PHP version**  8.0.28
**PHP SQLSRV version**  5.10.0 and 5.11.0
**Microsoft ODBC Driver version**  msodbcsql17 -- 17.8.1.1
**SQL Server version**  Azure SQL Database
**Client operating system**  Mac OS 10.15.7

**Table schema**  

```sql
create table test (
    testCol nvarchar(10)
);
```

**Problem description**  

When attempting to use a statement to insert two records into a table, and values have different data types, an error is produced.

**Expected behavior and actual behavior**  

actual: [Microsoft][ODBC Driver 17 for SQL Server]Invalid character value for cast specification
expected: (success)

**Repro code or steps to reproduce**

The following reproduces consistently when running sqlsrv v5.10.0 and v5.11.0

```php
$conn = sqlsrv_connect('XXXX.database.windows.net,1433', [
    'Database' => 'XXXX',
    'CharacterSet' => 'UTF-8',
    'UID' => 'XXXX',
    'PWD' => 'XXXX',
    'ReturnDatesAsStrings' => 1,
]);

$v0 = 1000;
$stmt = sqlsrv_prepare($conn, 'INSERT INTO [test] (testCol) VALUES (?);', [&$v0]);
sqlsrv_execute($stmt); // 1000 is inserted
$v0 = 'abcd';
sqlsrv_execute($stmt);

$error = sqlsrv_errors(SQLSRV_ERR_ERRORS);
echo $error[0][2] ?? ''; // [Microsoft][ODBC Driver 17 for SQL Server]Invalid character value for cast specification
```

**!important! Please note that the above snippet DOES NOT produce an error in sqlsrv v5.9.0.** Bug must have been introduced in v5.10.0

v-makouz:
I can reproduce this, so I'll lok into what might be causing it

v-makouz:
Update: I haven't fully figured out the exact change that caused this bug, still looking, but the reason is that it on the second insert the driver doesn't try to detect the type of the variable and uses the setting from the first insert. So '1000' is detected as BIGINT, everything is fine, then tries to insert 'abcd' as BIGINT, which doesn't work. 

I suspect it binds the variable and doesn't rebind when it's a different type. I'm trying to find the change that is the culprit.

absci:
[...]

PR #1497, opened 2024-01-16
Fix invalid precision error

MichaelSun90:
Fix for issue #1491


MichaelSun90:
Upon discuses, we decide to not do the SQL data type reset step within bind params function that initially there for fixing #1448. With that change, the find_param return a param with SQL unknow data type when execute the prepared query for the second time which will later be process into a SQL wvarchar type. When the logic try to bind the wvarchar with the input 8000 size which is twice the max size for a wvarchar type can hold cause the invalid precision error. This type process is related to a older version of utf8 encoding handling. The varchar type able to hold utf8 and utf16 encoding for newer version SQL server (later then 2019), and the handling for this and backward compatibility may come in a later. If the logic does not reset the SQL type unknow, the find_param will return the stored datatype varchar, and 8000 size will no longer be a problem.  For the issue #1448, if user properly handles the input param data, then the current logic should handles the task without a problem. Also, the error :"Invalid character value for cast specification" returned for that issue seems properly informed the user the cause for the error.

absci (approved):

Issue #1498, opened 2024-01-18
New attribute for sql bigint and decimal/numeric types to php int and float types


startork:
### Is your feature request related to a problem? Please describe.
We are changing our pdo driver from pdo_dblib to pdo_sqlsrv and have issues with bigints and decimals being cast to strings. We pass these values straight to our frontend which does some small calcs for display (totals etc...). I have read through the issues related to this area such as #291 and absolutely agree that by default these types need to be strings to preserve precision. However, a) bigint is the same size as the php int type on most systems b) as mentioned, to do any sort of operation on these values they must be converted to php floats anyway. There is no built-in way to arithmetically handle decimals as decimals in PHP so I don't see the harm in (optionally) always converting these to floats.

### Describe the solution you'd like
One or two attributes added to allow default types of int and float for bigint and decimal/numeric types.

### Describe alternatives you've considered
Currently I am using a custom version of pdo_sqlsrv with a modified sql_type_to_php_type method.

v-makouz:
If you want to add this option you can submit the PR and we can test it and pull it into the a future version.

PR #1500, opened 2024-01-20
Add fetch_bignumeric attribute

startork:
Related to #1498 

Add in optional bignumeric attribute which converts sql bigints to php ints and sql decimal and numeric to php float

Add basic tests for string vs int/float returns


startork:
@microsoft-github-policy-service agree

startork:
Hi @v-makouz 
Would you be able to give me some advice? I'm unsure whether to
a) Treat bigints as integers straight from the start, using c++ long long types or 
b) to handle similar to datetime, keeping existing code as much as possible and only converting to int when required as a zval.

Thanks,
Starsky

v-makouz:
I think treating it like datetime would be more straightforward. But I see there are still some automated test fails, like the "Numeric value out of range"

vaishnavyogesh:
@startork @v-makouz how can we run tests locally on any change? is there any local development setup guide?

PR #1520, opened 2024-08-20
Update run-perf_tests.py

v-makouz:
Switch from using SHA1 hash to SHA256


theodorejb:
What does this fix?

DanielRuf:
This fixes nothing.

The sha1 hash is just used to create a hashtable (PerformanceResults in `get_driver_id`) and doesn't have to be cryptographically secure.

For tests the performance is more relevant.

theodorejb:
@DanielRuf It's strange that this is a pull request since it is already committed to the dev branch here: https://github.com/microsoft/msphpsql/commit/30ed4b551d8144b2f43503048825f3ad421fb5e4

To me the last 48 commits to this repository (all made by the same individual) are troubling. They all have a commit message of the form "Update {filename}", without any details about what was changed or the reason for it. It looks like someone was just experimenting with incomplete changes which should have been on their own test branch rather than the default branch of the repository.

DanielRuf:
> To me the last 48 commits to this repository (all made by the same individual) are troubling. They all have a commit message of the form "Update {filename}", without any details about what was changed or the reason for it. It looks like someone was just experimenting with incomplete changes which should have been on their own test branch rather than the default branch of the repository.

Probably not best practice, but I don't even see a CONTRIBUTING.md.
Microsoft should enforce some proper rules.

Let's see how development will continue.
So far it's not ideal, that's for sure.

jahnvi480:
@v-makouz Thanks for raising the PR but looks like we have already merged the changes using other branch. Will close this PR

PR #1543, opened 2025-10-19
PHP 8.5 (compilation) support

stancl:
Since https://github.com/php/php-src/pull/17741, dbh->error_mode is stored as uint8_t. We cast the value to pdo_error_mode for the assignment to work.

Since https://github.com/php/php-src/pull/17742, query_stmt_zval has been replaced by query_stmt_obj. The diff in the linked PR shows the new destructor usage. We follow that same usage in this commit.

---

A few notes:
- I have limited experience with php-src, so please review with care. The context in the linked PRs should be sufficient.
- I did not guard the changes in `pdo_dbh.cpp` with PHP version checks. I assume (but do not know) that casting from `pdo_error_mode` to `pdo_error_mode` (in PHP <= 8.4) shouldn't emit any compiler warnings.
- Initially I used just `(pdo_error_mode) dbh->error_mode` but noticed that existing code uses `static_cast`. I don't have experience with C++ but assume this should be correct.

The changes are confirmed to compile for me locally, but this should be confirmed in CI.


stancl:
@microsoft-github-policy-service agree

Jan-E:
It compiles under Windoiws (VS17). I do not have a running sqlsrv database, so I did not run the tests.

theoaksoft:
Hello, When will this PR be integrated? This is the only thing delaying some of us from upgrading.

jahnvi480 (approved):

iBotPeaches (approved):

PR #1561, opened 2026-01-14
Fix pdo_connection_quote.phpt for PHP 8.4 v2

boom-ka:
PHP 8.3 says "cannot be empty", PHP 8.4 says "must not be empty". We use * wildcards to match both.


jahnvi480 (approved):

grant-mkm:
Bit late but you removed `$error2` being set but not from the following `if` block.

PR #1570, opened 2026-02-09
Using vs17 for compilation going forward

jahnvi480:
This pull request simplifies the logic for determining the compiler version in the `buildscripts/buildtools.py` script. The function now always sets the compiler to `vs17` instead of selecting it based on the PHP version.

Build tools logic simplification:

* The `compiler_version` method in `buildscripts/buildtools.py` was updated to always set `self.vc` to `'vs17'`, removing the previous conditional logic that selected the compiler based on the PHP version.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,7 +19,7 @@ jobs:
 - job: macOS
   pool:
     vmImage: 'macos-latest'
-  timeoutInMinutes: 0
+  timeoutInMinutes: 90
   variables:
     phpver: '8.5'
     macOS_sqlsrv_db: 'macos_sqlsrv_testdb'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,6 +19,7 @@ jobs:
 - job: macOS
   pool:
     vmImage: 'macos-latest'
+  timeoutInMinutes: 0
   variables:
     phpver: '8.5'
     macOS_sqlsrv_db: 'macos_sqlsrv_testdb'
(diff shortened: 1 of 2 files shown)


gargsaumya (approved):

azure-pipelines.yml
```diff
@@ -19,6 +19,7 @@ jobs:
 - job: macOS
   pool:
     vmImage: 'macos-latest'
+  timeoutInMinutes: 0
```
David-Engel: We don't want an unlimited timeout. Let's just bump it 50%. (60->90 mins).
```suggestion
  timeoutInMinutes: 90
```

bewithgaurav (approved):

buildscripts/buildtools.py
```diff
@@ -97,14 +97,7 @@ def determine_compiler(self, sdk_dir, vs_ver):
...
         if self.vc == '':
-            VC = 'vc15'
-            version = self.version_label()
-            if version[0] == '8':     # Compiler version for PHP 8.0 or above
-                if version[1] >= '4':
-                    VC = 'vs17'
-                else:
-                    VC = 'vs16'
-            self.vc = VC
+            self.vc = 'vs17'
```
David-Engel, line 100: Is this valid for PHP 8.3? The pipeline isn't exercising PHP 8.3. If we ever want to extend PR validation to all supported PHP versions, we'd have to undo this. (Not proposing we do that now. But this removal seems premature.)

David-Engel (dismissed):
I'm guessing the real purpose of this PR was to get the pipelines running a bit faster to address the macOS job 60 minute timeout failures. That was accomplished by the infinite job timeout of 0 (we should just bump that up incrementally, not to infinite). The brew and git changes might save a minute or two. The vs17-noly compilation won't save any time in the macOS job. I think that change is a bit premature unless PHP 8.3 can also be compiled by vs17.

buildscripts/buildtools.py
jahnvi480: Yes this is valid for PHP8.3 as well I ran it in my local and build the driver and it is working fine

azure-pipelines.yml
jahnvi480: Made the required changes

gargsaumya (approved):

bewithgaurav (approved):

PR #1590, opened 2026-03-26
FIX: Prepared statement silently fails on insert - SET NOCOUNT ON fixes it

jahnvi480:
This pull request addresses a critical issue where unconsumed result sets from prepared statement inserts could cause silent data loss due to implicit transaction rollbacks when using SQL Server with MARS (Multiple Active Result Sets) enabled. The fix ensures that all pending result sets are consumed before statement destruction, preventing unintended rollbacks. New functional tests are also added to verify that inserts persist as expected, even when extra result sets are generated by triggers or special SQL Server settings.

**Bug fix: Prevent silent data loss with unconsumed result sets**
- Ensures all pending result sets are consumed before freeing statement handles in both the `pdo_sqlsrv` and `sqlsrv` drivers, preventing the ODBC driver from rolling back uncommitted transactions when MARS is enabled. This addresses cases where triggers, `SET STATISTICS PROFILE ON`, or `SET NOCOUNT OFF` produce extra result sets. 

**Testing: Add functional tests to verify data persistence**
- Adds `pdo_insert_with_extra_result_sets.phpt` to test that prepared statement inserts persist data even when extra result sets are produced by triggers or SQL Server settings in the `pdo_sqlsrv` driver.
- Adds `sqlsrv_insert_with_extra_result_sets.phpt` to test the same scenarios for the `sqlsrv` driver.


jahnvi480:
@David-Engel , the `catch` block will remain uncovered because triggering it requires `core_sqlsrv_next_result` to throw an exception during destruction which would require something like the server dropping the connection mid-drain. That's not reliably reproducible in a phpt test without external infrastructure (e.g., killing the SQL Server process mid-test).
What should we do in this case?

test/functional/pdo_sqlsrv/pdo_insert_with_extra_result_sets.phpt
```diff
@@ -0,0 +1,112 @@
...
+    if ($count != 3) {
+        echo "FAIL: Expected 3 rows after insert with trigger, got $count\n";
+    } else {
+        echo "Test 4 passed: $count rows inserted via prepared statement with trigger.\n";
+    }
+
+    // Cleanup
+    $conn->exec("DROP TRIGGER trg_insert_$tableName");
+    $conn->exec("DROP TABLE $tableName");
+    $conn = null;
```
David-Engel: Move the cleanup outside the try block to ensure it runs on failures. You can change the statements to DROP TABLE/TRIGGER **IF EXISTS** XXX so they don't error if the objects were never created.

test/functional/pdo_sqlsrv/pdo_stmt_dtor_drain_statistics.phpt
```diff
@@ -0,0 +1,41 @@
+--TEST--
```
David-Engel: What's different about this test? It seems like it duplicates what's already in pdo_insert_with_extra_result_sets.phpt‎.

test/functional/pdo_sqlsrv/pdo_stmt_dtor_drain_multi_resultset.phpt
```diff
@@ -0,0 +1,38 @@
+--TEST--
```
David-Engel: Can this file and pdo_stmt_dtor_drain_results.phpt‎ be combined into the first test file? There is an overhead to each additional test file added. I don't understand the choice to create so many new test files.

test/functional/pdo_sqlsrv/pdo_insert_with_extra_result_sets.phpt
```diff
@@ -0,0 +1,112 @@
...
+statement insert should still persist data. Previously, unconsumed result sets left
+pending on the ODBC statement handle could be cancelled during statement destruction,
+causing an implicit rollback of the insert.
+--SKIPIF--
+<?php require('skipif.inc'); ?>
+--FILE--
+<?php
+require_once('MsCommon.inc');
+
+$tableName = 'pdo_insert_extra_results_test';
```
David-Engel:
```suggestion
$tableName = $tableName = GetTempTableName('pdo_drain_results', false);
```

David-Engel (commented):
The comments on the pdo test files also apply to the ssqlsrv test files.

David-Engel (approved):

PR #1592, opened 2026-03-27
Fix | Access token identity leaking across pooled connections

### LINKED ISSUE #1396: An existing connection in the connection pool is reused unexpectedly even though it has a different access token.

David-Engel:
Fixes #1396

### Problem

When using AccessToken with connection pooling, two issues cause incorrect behavior:

1. Identity cross-contamination: The ODBC Driver Manager matches pooled connections by connection string but does not compare driver-specific attributes like `SQL_COPT_SS_ACCESS_TOKEN`. Connections to the same server with different access tokens share the same pool entry, returning the wrong identity to callers.

2. Use-after-free: The ODBC driver stores only the raw `ACCESSTOKEN` pointer — it never copies the token data. The previous code freed the token immediately after `SQLDriverConnect` returned. If the DM pooled the physical connection and idle connection recovery later fired, the driver would dereference freed memory.

### Solution

**Pool-key differentiation** - After building the connection string, a FNV-1a 64-bit hash of the token content is appended as `APP={AT-<hash>};`. This gives the DM a distinct pool key per token without affecting the ODBC driver's behavior. If the user also sets `APP`, the first occurrence determines the server-side program_name, but the DM sees the full string for pool matching.

**Token cache for pointer stability** - A process-wide token cache (`token_cache_get`) ensures the same raw token content always resolves to the same `ACCESSTOKEN*`. Tokens are allocated with `malloc` (not `emalloc`) so they persist across PHP requests. The `sqlsrv_conn::azure_ad_access_token` field is changed from an owning `sqlsrv_malloc_auto_ptr<ACCESSTOKEN>` to a non-owning `ACCESSTOKEN*` managed by the cache.

**TTL-based expiry** - Since there's no API hook for when physical connections are actually closed in the pool, cached tokens are removed after a configurable TTL derived from the ODBC driver's CPTimeout setting (TTL = max(CPTimeout × 2, 120s)). The TTL is read from ODBCINST.INI on Linux (`SQLGetPrivateProfileString`) and the Windows registry (`RegOpenKeyExA`/`RegQueryValueExA`) during MINIT. Token memory is scrubbed with `explicit_bzero`/`SecureZeroMemory` before being freed.

**Thread safety** - All cache operations are protected by `std::mutex`, safe for ZTS builds. `token_cache_touch()` updates the entry's timestamp before `SQLDisconnect()` returns the physical connection to the DM pool, closing a potential race window where another thread's cleanup could expire the entry.

#### Core driver (source/shared/):

`core_conn.cpp` - Token cache implementation (get/touch/expire/destroy/init_ttl), `secure_zero` helper, pool-key injection in `build_connection_string_and_set_conn_attr`, non-owning token reference in `core_sqlsrv_close` and `access_token_set_func`
`core_sqlsrv.h` - `core_sqlsrv_hash_fnv1a_64` inline function, cache init/cleanup declarations, `azure_ad_access_token` changed from `sqlsrv_malloc_auto_ptr<ACCESSTOKEN>` to raw pointer
`core_init.cpp` - `core_sqlsrv_init_token_cache()` called in MINIT, `core_sqlsrv_cleanup_token_cache()` called in MSHUTDOWN

#### Test infrastructure:

`test/tools/mock_tds_server.py` - Mock TDS 7.4 server (~1,450 lines) supporting LOGIN7, access token extraction, PRELOGIN, session recovery FeatureExtAck, KILL by SPID, TLS wrapping, and SELECT SUSER_SNAME() / @@SPID queries
`test/tools/test_mock_tds_server.py` - Unit tests for the mock server
`test/functional/sqlsrv/mock_tds_helper.inc` - Shared PHP helper for managing the mock server process and generating self-signed certificates
`test/functional/sqlsrv/sqlsrv_mock_access_token.phpt` / `pdo_sqlsrv/pdo_mock_access_token.phpt` - PHPT tests that connect with two different access tokens via the mock server and verify each connection returns the correct identity
`test/functional/sqlsrv/sqlsrv_mock_access_token_worker.php` / `pdo_sqlsrv/pdo_mock_access_token_worker.php` - Worker scripts for the above tests

#### Build:

`build-linux.sh` - Linux build script for both extensions
`.gitignore` - Ignore build artifacts from phpize/make


source/shared/core_conn.cpp
```diff
@@ -79,6 +81,214 @@ bool core_search_odbc_driver_unix(_In_ ODBC_DRIVER driver);
...
+    TokenCacheEntry* next;
+};
+
+static TokenCacheEntry* s_token_cache = nullptr;
+static time_t s_token_cache_ttl = 120; // default floor; updated by token_cache_init_ttl()
+static std::mutex s_token_cache_mutex;
+
+// Minimum TTL floor (seconds).  Even if CPTimeout is very small, we keep
+// tokens alive for at least this long to account for in-flight recoveries.
+static const time_t TOKEN_CACHE_TTL_FLOOR = 120;
```
saurabh500, line 121: Out of curiosity, what was the thought process behind 120?

source/shared/core_sqlsrv.h
```diff
@@ -1052,6 +1054,19 @@ extern HashTable* g_encodings;                    // encodings supported by this
 
 void core_sqlsrv_minit( _Outptr_ sqlsrv_context** henv_cp, _Inout_ sqlsrv_context** henv_ncp, _In_ error_callback err, _In_z_ const char* driver_func );
 void core_sqlsrv_mshutdown( _Inout_ sqlsrv_context& henv_cp, _Inout_ sqlsrv_context& henv_ncp );
+void core_sqlsrv_init_token_cache();
+void core_sqlsrv_cleanup_token_cache();
+
+// FNV-1a 64-bit hash used by both the token cache and APP pool-key generation.
+inline uint64_t core_sqlsrv_hash_fnv1a_64(const char* data, size_t len)
```
saurabh500, line 1061: I am curious about FNV-1a hash. Why was this used instead of collision resistant Sha256?

test/functional/pdo_sqlsrv/pdo_mock_access_token.phpt
```diff
@@ -0,0 +1,67 @@
...
+On Linux/macOS, pooling is enabled via a custom odbcinst.ini with CPTimeout on
+each driver section and the test runs in a subprocess with ODBCSYSINI set.
+On Windows, pooling is controlled by the ConnectionPooling option.
+--SKIPIF--
+<?php
+if (!extension_loaded("pdo_sqlsrv")) {
+    die("skip pdo_sqlsrv extension not loaded");
+}
+require_once(__DIR__ . '/../sqlsrv/mock_tds_helper.inc');
+$mock = start_mock_tds_server();
```
saurabh500, line 20: The legendary server !

test/functional/sqlsrv/mock_tds_helper.inc
```diff
@@ -0,0 +1,398 @@
...
+    return strtoupper(substr(PHP_OS, 0, 3)) === 'WIN';
+}
+
+/**
+ * Find a working Python interpreter (python3 on Linux/macOS, python on Windows).
+ * Returns the command name, or null if not found.
+ */
+function find_python()
+{
+    $candidates = is_windows() ? ['python', 'python3'] : ['python3', 'python'];
```
saurabh500: Nit:
Why reverse the order? Are the chances of finding "python" higher on Windows ? For the sake of simplicity, can we not keep a single array on the two platforms?

source/shared/core_conn.cpp
```diff
@@ -814,6 +1028,27 @@ void build_connection_string_and_set_conn_attr( _Inout_ sqlsrv_conn* conn, _Inou
...
+        // We use APP because it is a well-known keyword that won't cause
+        // warnings.  If the user already set APP, the first occurrence wins
+        // for the server-side program_name, but the DM still sees the full
+        // string (including our duplicate) for pool-key matching.
+        if (access_token_used && conn->azure_ad_access_token) {
+            ACCESSTOKEN* tok = conn->azure_ad_access_token;
+            uint64_t h = core_sqlsrv_hash_fnv1a_64(
+                tok->data, static_cast<size_t>(tok->dataSize));
+            char pool_key[64];
+            snprintf(pool_key, sizeof(pool_key),
```
saurabh500, line 1047: Is it OK to override the APP? if the App is set in the PHP connection string, then this line should override it, isn't it? And this will get sent to the server as well.
Or are we blocking the App from conn str like Python ?

test/functional/pdo_sqlsrv/pdo_mock_access_token.phpt
```diff
@@ -0,0 +1,67 @@
...
+$mock = start_mock_tds_server([
+    '--map-token', "$tokenA=alice",
+    '--map-token', "$tokenB=bob",
+]);
+
+$server = "127.0.0.1,{$mock['port']}";
+
+// On Linux/macOS, create custom odbcinst.ini with pooling enabled.
+// On Windows, this returns null (pooling is controlled via ConnectionPooling option).
+$pooldir = create_pooling_odbcinst();
```
saurabh500, line 39: Should we be testing without the CPTimeout settings as well?

saurabh500: I mean test with default ini, which doesn't have `CPTimeout` ? (ignore if that case is already covered)

saurabh500 (approved):
Left some comments. Nothing blocking though

source/shared/core_conn.cpp
David-Engel: On Windows, if a driver doesn't specify it, the default is 60 seconds. CoPilot picked 120. I didn't argue. On Linux, if a driver doesn't specify CPTimeout, pooling doesn't happen.

source/shared/core_sqlsrv.h
David-Engel: Detailed AI response: 

FNV-1a is the right tool here because neither use case needs cryptographic properties:

Token cache lookup — The hash is a fast-rejection filter before `memcmp`. A false positive just means we do one extra comparison. There's no security implication — an attacker can't exploit a hash collision to get a different token's pointer, because the full content comparison is authoritative.

APP pool key — The hash differentiates pool entries in the DM. A collision would mean two different tokens share a pool slot (the original bug). With 64-bit FNV-1a, the birthday-paradox threshold is ~5 billion distinct tokens before a 50% collision probability. A PHP process will never see anywhere near that many.

SHA-256 would add:

A dependency on a crypto library (OpenSSL/BCrypt) or bundling an implementation
~50× slower throughput per byte (SHA-256 ~500 MB/s vs FNV-1a ~25 GB/s on modern CPUs)
A 32-byte output that would need to be truncated or hex-encoded into the connection string anyway
No actual security benefit — the hash isn't protecting against adversarial input. The token content comes from the application's own authentication flow, not from untrusted external sources.
SHA-256 is for when you need collision resistance against an adversary or need a one-way function. Here we just need a fast, well-distributed hash to tell tokens apart. FNV-1a fits that exactly.

test/functional/sqlsrv/mock_tds_helper.inc
David-Engel: On Windows the standard installer creates python (not python3), while on Linux/macOS python3 is the canonical name and python may point to Python 2 or not exist. But the preference order is just a micro-optimization for skipping one failed exec. It doesn't matter functionally — both candidates are tried regardless.

A single unified array is simpler and works fine. I'll update it.

source/shared/core_conn.cpp
David-Engel: For generic ODBC keywords, the last one wins (spec requirement). For driver-specific keywords, the spec doesn't specify. For APP in MS ODBC SQL, the first APP in the connection string wins (validated). But all entries in the connection string, including duplicates, are considered for the pool key. (We don't block APP.)

test/functional/pdo_sqlsrv/pdo_mock_access_token.phpt
David-Engel: I considered it. But when pooling is off, as long as we only free the Access Token after disconnect, there's no crash risk and there is no risk of returning the wrong connection from the pool.

David-Engel:
~~Don't merge this, yet. The new test is crashing on macOS (borked). The task to analyze results needs to be updated to detect borked tests.~~

saurabh500 (approved):

Issue #1466, opened 2023-07-27
Fatal error: Invalid column number in pdo_sqlsrv_stmt_get_col_data


kicken:
**PHP version**
8.1.10

**PHP SQLSRV or PDO_SQLSRV version** 
PDO_SQLSRV 5.11.0

**Microsoft ODBC Driver version**
2018.182.02.01

**SQL Server version**
Microsoft SQL Server 2019 (RTM) - 15.0.2000.5 (X64)   Sep 24 2019 13:48:23   Copyright (C) 2019 Microsoft Corporation  Developer Edition (64-bit) on Windows 10 Pro 10.0 <X64> (Build 22621: ) (Hypervisor) 

**Client operating system**
Windows 11 22H2

**Problem description**  
Attempting to re-use a prepared statement object that returns multiple result sets fails on subsequent executions.  

In some instances, PHP exists with the fatal error:

    Fatal error: Invalid column number in pdo_sqlsrv_stmt_get_col_data in

In other instances, the PHP executable hard-crashes 

    Faulting application name: php.exe, version: 8.1.10.0, time stamp: 0x630e529d
    Faulting module name: php8.dll, version: 8.1.10.0, time stamp: 0x630e5af2
    Exception code: 0xc0000005
    Fault offset: 0x000000000027be4a

In yet other instances, the returned data is simply incorrect.  The result of the first result set is returned with the column names of the last result set.

**Expected behavior and actual behavior**  
No error or crash.

**Repro code or steps to reproduce**

```php
<?php
var_dump(PHP_VERSION);
$db = new PDO('sqlsrv:server=localhost;Database=tempdb;TrustServerCertificate=1;Encrypt=1;Driver=ODBC Driver 18 for SQL Server');
$db->setAttribute(PDO::ATTR_DEFAULT_FETCH_MODE, PDO::FETCH_ASSOC);
$db->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);

$sql = /** @lang TSQL */
    '
set nocount on;
declare @id int = :id;
-- Result set 1
SELECT
        @id AS r1int1
        , 2 as r1int2
        , 3 as r1int3
        , 4 as r1int4
        , N\'Test\' AS r1str1
        , N\'test\' AS r1str2
        , 5 AS r1int5
        , SYSUTCDATETIME() as r1dt1
where @id in (2,3);

-- Result set 2
select
        @id as r2int1
        , 1 AS r2int6
[...]

PR #1596, opened 2026-04-03
FIX: Freeing metadata entries and clearing the vector

jahnvi480:
Github Issue #1466 
This pull request addresses a critical bug (GH#1466) where re-executing a prepared statement that returns multiple result sets with different column layouts could cause crashes or data corruption in the SQLSRV and PDO_SQLSRV drivers. The changes ensure that stale column metadata is properly cleared between executions, and comprehensive tests are added to verify correct behavior in various scenarios.

**Bug Fixes:**

* Ensured that stale PDO column descriptors are freed before re-executing a prepared statement in `pdo_stmt.cpp`, so that columns are correctly re-described for new result sets and to prevent crashes or data corruption.
* Updated `core_stmt.cpp` to delete results metadata from the previous result set when starting a new result set, preventing stale metadata issues on re-execution.

**Testing Improvements:**

* Added a functional test `pdo_reexecute_multi_resultset.phpt` for PDO_SQLSRV that verifies re-executing prepared statements with multiple result sets does not cause fatal errors or data corruption, including edge cases with and without cursor closing and with different column names.
* Added a similar functional test `sqlsrv_reexecute_multi_resultset.phpt` for SQLSRV, ensuring consistent behavior and validating the fix for both drivers.


David-Engel (commented):
Here are my recommendations, ordered by importance:

---

### 1. Keep the `new_result_set()` change — it's the correct fix

Replacing `reset_php_type()` with `clean_up_results_metadata()` in `new_result_set()` is right. The old code only reset the cached PHP type mapping but preserved stale SQL metadata (column names, types, sizes). That's the root cause of the bug. No change needed here.

---

### 2. Scope the explicit cleanup in `nextRowset`/`next_result` to only the end-of-results case

`core_sqlsrv_next_result()` calls `new_result_set()` when there **are** more results, which now does the cleanup. It does **not** call `new_result_set()` when `past_next_result_end` — so the explicit call is only needed there. Currently it runs unconditionally, causing a redundant double-cleanup.

**In stmt.cpp**, change to:

```cpp
core_sqlsrv_next_result( stmt, true );

if( stmt->past_next_result_end ) {
    // Clean up remaining metadata since new_result_set() was not called
    stmt->clean_up_results_metadata();
    RETURN_NULL();
}

RETURN_TRUE;
```

**In pdo_stmt.cpp**, change to:

```cpp
core_sqlsrv_next_result( static_cast<sqlsrv_stmt*>( stmt->driver_data ) );

if( driver_stmt->past_next_result_end == true ) {
    driver_stmt->clean_up_results_metadata();
    return 0;
}

stmt->column_count = core::SQLNumResultCols( driver_stmt );
// ... rest unchanged
```

This eliminates the redundant path and makes the cleanup responsibility clear: `new_result_set()` owns cleanup when advancing to a valid result set; explicit cleanup happens only at end-of-results.

---

### 3. Simplify the PDO `executed` workaround block

After `php_pdo_stmt_set_column_count(stmt, 0)`, `stmt->columns` is always NULL. So the old workaround guard `if (stmt->columns == NULL)` always evaluates to true. Merge the two into a single coherent block:

```cpp
// On re-execution, free stale PDO column descriptors and reset the
// executed flag so that PDO re-describes columns for the new result
// set (see GH#1466). Setting executed = 0 works around a PDO driver
// manager optimization that otherwise skips the describe_col call.
if ( stmt->columns ) {
    php_pdo_stmt_set_column_count( stmt, 0 );
}
stmt->executed = 0;
```

Remove the old separated workaround block and its lengthy comment. The `executed = 0` is now always needed on re-execution (not just the niche `nextRowset → execute → getColumnMeta` sequence), and the new comment explains both mechanisms together. On first execution, `columns` is NULL so we skip the free, but `executed = 0` is harmless (it's already 0).

---

### Summary

| Recommendation | Why |
|---|---|
| Keep `clean_up_results_metadata()` in `new_result_set()` | Core correctness fix |
| Scope explicit cleanup to `past_next_result_end` only | Eliminates redundant double-cleanup, clarifies ownership |
| Merge `php_pdo_stmt_set_column_count` + `executed = 0` | Removes dead code path, makes intent explicit |

David-Engel (approved):

PR #1598, opened 2026-04-07
Fix stream invalidation when statement goes out of scope (GH#1443)

jahnvi480:
This pull request addresses a critical bug where binary streams returned from a statement could become invalid if the statement object went out of scope, even though the stream was still in use. The changes ensure the statement resource remains alive as long as any associated stream is referenced, preventing premature destruction and stream invalidation. Additionally, a regression test is added to verify the fix.

**Bug Fixes and Resource Management:**

* Modified `sqlsrv_stream` and `sqlsrv_stmt` structures to track and retain a reference to the statement's `zend_resource`, ensuring the statement is not destroyed while a stream is active.
* Updated stream creation logic to increment the statement resource's reference count when a stream is created, and to properly release it when the stream is closed.
* Ensured that the statement's `zend_resource` is stored during both `sqlsrv_prepare` and `sqlsrv_query` so streams can access it. 

**Testing and Documentation:**

* Added a functional test (`sqlsrv_stream_outlives_stmt.phpt`) to confirm that streams remain valid after their originating statement goes out of scope, covering multiple scenarios.
* Updated the changelog to document the bug fix for binary streams becoming invalid when the statement goes out of scope.


test/functional/sqlsrv/sqlsrv_stream_outlives_stmt.phpt
```diff
@@ -0,0 +1,116 @@
+--TEST--
```
David-Engel, line 1: Can we validate that the statement and stream get cleaned up appropriately? (I'm not sure if we can at the PHPT level...)

David-Engel (dismissed):

test/functional/sqlsrv/sqlsrv_stream_outlives_stmt.phpt
jahnvi480: Added tests for this

David-Engel (approved):

